PR #13, opened 2017-05-11
Upgrade gradle infrastructure

schlosna:
Switch to nebula dependency recommendation


schlosna:
Note that this removes the `jmh` gradle plugin as it had some conflicts with nebula, and in place there is now a `benchmarks` gradle task to run all JMH benchmarks.

PR #26, opened 2017-06-01
Make registerSafe more threadsafe.

j-baker:
Before, we'd see a bunch of our Spark tasks attempt to register the same
metrics concurrently and throw and get retried (the first time).

Now, it is threadsafe to call registerSafe from multiple threads


schlosna:
Talked through this with James and his use seems reasonable, though in most cases where there would be multiple threads registering metric under same name, you probably wouldn't want that as you effectively get `last-write-wins` semantics.

Issue #14, opened 2017-05-11
Add `@SlowLog(duration,unit)` annotation


schlosna:
From @markelliot: I'd like to annotate some methods with `@SlowLog(duration,unit)` such that if the method takes longer than that execute we log the method and the parameters to record that it ran slower than expected.

A similar enhancement would be to do this for slowest 1% of method calls.

@ellisjoe @ash211 for SA

ellisjoe:
this would be great

schlosna:
Interested in feedback on what logger this would go to, at what level, and if you'd annotate the instrumented interface and/or implementation?

ellisjoe:
think for services we'll want to stick them on the implementation. the service interfaces currently just define the rest api and this seems like something directly related to a specific implementation anyway.

if we're logging just the slowest 1% we should likely log at `INFO`. if we we're able to log calls that appear to be outliers, think those would be better logged at `WARN` (taking the stance that `WARN` logs are something we should take a look at within a day or so).

could see this being logged to the same place as the rest of metrics or it's own file. don't have a strong opinion there.

PR #33, opened 2017-10-02
Pd/metric grouping

pdunn:
Provide facilities for controlling the granularity of metrics reporting within an instrumented class and across classes that share a MetricsRegistry.

@schlosna , can you take look here?  This differs from tagging in that tagging allows you to differentiate reporters of a shared metric name, while this allows for the bundling of instrumented calls into a shared metric.  The annotation facilities could be leveraged to provide additional handling for Method (static) and Parameter (dynamic) tags.


tritium-lib/src/test/java/com/palantir/tritium/TritiumTest.java
```diff
@@ -73,6 +73,10 @@ public void testInstrument() {
         Slf4jReporter.forRegistry(metricRegistry).withLoggingLevel(Slf4jReporter.LoggingLevel.INFO).build().report();
     }
 
+    public void testInstrumentWithPrefix() {
+
```
schlosna: - [ ] TODO?

tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
```diff
@@ -118,14 +118,20 @@ private Builder(Class<T> interfaceClass, U delegate) {
             this.delegate = checkNotNull(delegate, "delegate");
         }
 
-        public Builder<T, U> withMetrics(MetricRegistry metricRegistry) {
+        public Builder<T, U> withMetrics(MetricRegistry metricRegistry, String globalPrefix) {
```
schlosna: We should add JavaDoc for what the global prefix entails

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/MetricsInvocationEventHandler.java
```diff
@@ -42,16 +47,55 @@
     private final MetricRegistry metricRegistry;
     private final String serviceName;
 
+    //consider creating annotation handlers as separate objects
+    private final Map<AnnotationHelper.MethodSignature, String> metricGroups;
+    private final String globalGroupPrefix;
```
schlosna: nit: mark `@Nullable private final String globalGroupPrefix;`

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/MetricsInvocationEventHandler.java
```diff
@@ -42,16 +47,55 @@
     private final MetricRegistry metricRegistry;
     private final String serviceName;
 
+    //consider creating annotation handlers as separate objects
```
schlosna, line 50: +1, I think we'll want to have mapping of method signature to handler to implement https://github.com/palantir/tritium/issues/14

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/annotations/AnnotationHelper.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    private static List<Class<?>> getParentClasses(Class<?> clazz) {
+        ImmutableList.Builder<Class<?>> builder = ImmutableList.builder();
+        if (clazz.getSuperclass() != null) {
+            builder.add(clazz.getSuperclass());
+        }
+
+        return builder.addAll(Arrays.asList(clazz.getInterfaces())).build();
+    }
```
schlosna: Wondering if we need to traverse the full super-type hierarchy with something like:

```
    @VisibleForTesting
    static Set<Class<?>> getAllParentClasses(Class<?> clazz) {
        ImmutableSet.Builder<Class<?>> builder = ImmutableSet.builder();
        builder.add(clazz.getInterfaces());
        Class<?> superclass = clazz.getSuperclass();
        while (superclass != null) {
            builder.add(superclass.getInterfaces());
            builder.add(superclass);
            superclass = superclass.getSuperclass();
        }
        return builder.build();
    }
```

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/annotations/AnnotationHelper.java
```diff
@@ -0,0 +1,156 @@
...
+
+    public static final class MethodSignature {
+        private final String methodName;
+        private final Class<?>[] parameterTypes;
+
+        private MethodSignature(String methodName, Class<?>... parameterTypes) {
+            this.methodName = checkNotNull(methodName);
+            this.parameterTypes = parameterTypes != null ? parameterTypes : new Class<?>[]{};
```
schlosna: nit: we probably need to clone the parameter types

```
private static final Class<?>[] NO_ARGS = new Class<?>[0];

this.parameterTypes = (parameterTypes == null || parameterTypes.length == 0) ? NO_ARGS : parameterTypes.clone();
```

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/MetricsInvocationEventHandler.java
```diff
@@ -42,16 +47,55 @@
...
     }
 
+    private static Map<AnnotationHelper.MethodSignature, String> createMethodGroupMapping(Class<?> serviceClass) {
+        ImmutableMap.Builder<AnnotationHelper.MethodSignature, String> builder = ImmutableMap.builder();
+
+        MetricGroup classGroup = AnnotationHelper.getSuperTypeAnnotation(serviceClass, MetricGroup.class);
+
+        for (Method method : serviceClass.getMethods()) {
+            AnnotationHelper.MethodSignature sig = AnnotationHelper.MethodSignature.of(method);
+            MetricGroup methodGroup = AnnotationHelper.getMethodAnnotation(MetricGroup.class, serviceClass, sig);
```
schlosna, line 87: see other comment

```
            MetricGroup methodGroup = AnnotationHelper.getAnnotationIfPresent(MetricGroup.class, serviceClass, method);
```

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/annotations/AnnotationHelper.java
```diff
@@ -0,0 +1,156 @@
...
+
+        for (Class<?> iface : getParentClasses(clazz)) {
+            T foundAnnotation = getMethodAnnotation(annotation, iface, methodSignature);
+            if (foundAnnotation != null) {
+                return foundAnnotation;
+            }
+        }
+
+        return null;
+    }
```
schlosna, line 87:
```
    /**
     * Depth first search up the Type hierarchy to find a matching annotation,  Types which do not implement the
     * specified method signature are ignored.
     *
     * @param annotation annotation class
     * @param clazz containing class
     * @param method method
     * @param <T> annotation type
     * @return - First found matching annotation or null
     */
    @Nullable
    public static <T extends Annotation> T getAnnotationIfPresent(Class<T> annotation, Class<?> clazz, Method method) {
        return getAnnotationIfPresent(annotation, clazz, method, method.getParameterTypes());
    }

    /**
     * Depth first search up the Type hierarchy to find a matching annotation. Types which do not implement the
     * specified method signature are ignored.
     *
     * @param annotationClass annotation class
     * @param clazz containing class
     * @param method method
     * @param parameterTypes method parameter types
     * @param <T> annotation type
     * @return - First found matching annotation or null
     */
    @Nullable
    private static <T extends Annotation> T getAnnotationIfPresent(
            Class<T> annotationClass,
            Class<?> clazz,
            Method method,
            Class... parameterTypes) {

        if (method == null) {
            return null;
        }

        T maybeAnnotation = method.getAnnotation(annotationClass);
        if (maybeAnnotation != null) {
            return maybeAnnotation;
        }

        String methodName = method.getName();
        for (Class<?> parentClass : getAllParentClasses(clazz)) {
            for (Method parentMethod : parentClass.getMethods()) {
                if (methodName.equals(parentMethod.getName())) {
                    Class<?>[] parentMethodParamTypes = parentMethod.getParameterTypes();
                    if (Arrays.equals(parameterTypes, parentMethodParamTypes)) {
                        maybeAnnotation = parentMethod.getAnnotation(annotationClass);
                        if (maybeAnnotation != null) {
                            log.info("Found {} on matching method {} on {}",
                                    maybeAnnotation, parentMethod, parentClass);
                            return maybeAnnotation;
                        }
                    }
                }
            }
        }

        return null;
    }
```

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/annotations/AnnotationHelper.java
```diff
@@ -0,0 +1,156 @@
...
+            this.methodName = checkNotNull(methodName);
+            this.parameterTypes = parameterTypes != null ? parameterTypes : new Class<?>[]{};
+        }
+
+        public String getMethodName() {
+            return methodName;
+        }
+
+        public Class<?>[] getParameterTypes() {
+            return parameterTypes;
```
schlosna, line 119: I think we can get away without exposing the getters, and probably move this to a top level class.

```
public final class MethodSignature {

    private final String name;
    private final List<Class<?>> parameterTypes;

    private MethodSignature(String name, List<Class<?>> parameterTypes) {
        this.name = name;
        this.parameterTypes = ImmutableList.copyOf(parameterTypes);
    }

    public static MethodSignature from(Method method) {
        return new MethodSignature(method.getName(), ImmutableList.copyOf(method.getParameterTypes()));
    }

    @Override
    public boolean equals(Object obj) {
        if (this == obj) {
            return true;
        }
        if (!(obj instanceof MethodSignature)) {
            return false;
        }

        MethodSignature that = (MethodSignature) obj;
        return name.equals(that.name) && parameterTypes.equals(that.parameterTypes);
    }

    @Override
    public int hashCode() {
        return 31 * name.hashCode() + parameterTypes.hashCode();
    }

    @Override
    public String toString() {
        return "MethodSignature{"
                + "methodName='" + name + '\''
                + ", parameterTypes=" + parameterTypes
                + '}';
    }

}
```

tritium-metrics/src/test/java/com/palantir/tritium/event/metrics/MetricsInvocationEventHandlerTest.java
```diff
@@ -69,4 +93,54 @@ public void testSystemPropertySupplier_Handler_Enabled() throws Exception {
...
+
+        //GlobalPrefix Tests
+        assertThat(metricRegistry.timer(globalPrefix + ".DEFAULT").getCount()).isEqualTo(2L);
+        assertThat(metricRegistry.timer(globalPrefix + ".ONE").getCount()).isEqualTo(2L);
+    }
+
+    private void callVoidMethod(
+            MetricsInvocationEventHandler handler, Object obj, String methodName, boolean success) throws Exception {
+
+        Method method = obj.getClass().getMethod(methodName);
```
schlosna: nit: do you want to use this in line below?

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/annotations/AnnotationHelper.java
pdunn: @schlosna , are you advocating ditching the MethodSignature as the argument here and going back to either (Method) or (String, Class[])?

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/annotations/AnnotationHelper.java
pdunn: Both times this used it's accessed in recursive methods that climb the type hierarchy.  I'm not sure it's needed and the behavior on on which interface is selected for annotations becomes the parent overwriting the child.

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/annotations/AnnotationHelper.java
schlosna: I had branched off your previous changes and was testing out some things. We probably need some test cases where method name is the same but signatures differ by argument types.

tritium-lib/src/main/java/com/palantir/tritium/Tritium.java
```diff
@@ -44,4 +44,22 @@ private Tritium() {
...
+     * metrics and performance trace logging.
+     *
+     * @param serviceInterface service interface
+     * @param delegate delegate to instrument
+     * @param metricRegistry metric registry
+     * @param globalPrefix shared prefix across service interfaces for Methods annotated with MetricGroup
+     * @return instrumented proxy implementing specified service interface
+     */
+    public static <T, U extends T> T instrument(
+            Class<T> serviceInterface, U delegate, MetricRegistry metricRegistry, String globalPrefix) {
```
schlosna: I'm leaning toward not adding this method here in `Tritium` as it is probably clearer for folks that want control over global prefix to use the `Instrumentation` builder explicitly

```
Instrumentation.builder(serviceInterface, delegate)
                .withMetrics(metricRegistry, globalPrefix)
                .withPerformanceTraceLogging()
                .build();
```

pdunn: cool, can remove that

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/MetricsInvocationEventHandler.java
pdunn: I'm thinking about ways to do this.  Most clean way seems to be:

1) MetricsInvocationEventHandler has a Multimap<MethodSignature, MetricsAnnotationHandler>
2) Each MetricsAnnotationHandler contains a Map<MethodSignature, Annotation>

Sort of begs the question if the MetricsInvocationEventHandler needs awareness of a the Annotation Handlers or whether the annotation handlers can be completely generic invocation handlers.

schlosna: yeah, I think we can leave this as-is for now and extract flesh out the annotation handlers separately

schlosna (approved):

PR #35, opened 2017-10-04
Pd/annotation handler

pdunn:
Extract MetricGroup annotation handling

@schlosna , here's a prototype to clean up that code.  The abstraction isn't perfect, e.g. failure metric name, and I'm obviously leaving a generalizing point for more annotation handlers unaddressed.


schlosna:
Going to close this out for now as its stale

PR #39, opened 2017-10-17
Create tritium-tags module

schlosna:
Strawman example tagging API

TODO:

- [ ] finalize format & parser
- [ ] `@Safe` & `@Unsafe` tags
- [ ] wire through `Instrumentation` and maybe `Tritium`


tritium-tags/src/main/java/com/palantir/tritium/tags/TaggedMetric.java
```diff
@@ -0,0 +1,145 @@
...
+        get = {"get*", "is*"},
+        // Try to avoid leaking Immutable prefixed objects to avoid
+        // api fallout when we switch to another processor.
+        // Side effect: Return the base type from builders.
+        // Side effect: Nest implementation class inside of builder class.
+        visibility = Value.Style.ImplementationVisibility.PRIVATE,
+        builderVisibility = Value.Style.BuilderVisibility.PACKAGE,
+        // Default interface methods don't need to be annotated @Value.Default
+        defaultAsDefault = true)
+public abstract class TaggedMetric {
```
ellisjoe, line 44: should this just be a static util class? similar to `MetricRegistry#name`?

ellisjoe:
think you can avoid wiring this through if we make this a static util and add a listener when we create metric registries that validates names like we talked about

tritium-tags/src/main/java/com/palantir/tritium/tags/TaggedMetric.java
```diff
@@ -0,0 +1,145 @@
...
+     */
+    @Value.Lazy
+    public String canonicalName() {
+        if (nonEmptyTags().isEmpty()) {
+            return name();
+        } else {
+            StringBuilder builder = new StringBuilder();
+            builder.append(name()).append('[');
+
+            for (Iterator<Map.Entry<String, String>> it = nonEmptyTags().entrySet().iterator(); it.hasNext(); ) {
```
j-baker: I think you'll benefit at some point from sorting the keys (e.g. if you make a map of these things, you'll be able to lookup).

j-baker: like presumably the user is going to be like metricRegistry.meter(TaggedMetric.of("name", singletonMap("tag", "value"))

ellisjoe:
looks good to me

tritium-tags/src/main/java/com/palantir/tritium/tags/TaggedMetric.java
schlosna: `TaggedMetric.toCanonicalName(String name, Map<String, String> tags)` static method will now generate the canonical string name for given metric name and tags.

tritium-tags/src/main/java/com/palantir/tritium/tags/TaggedMetric.java
schlosna: applied sorting and some normalization of keys (lower casing, require <= 20 alpha numeric)

PR #40, opened 2017-10-23
Make tritium-tracing compatible with remoting-3

j-baker:
At the moment it is not compatible - generated dynamic proxies will log into remoting 1 trace logs, which will not be listened to due to package name changes (yes this is rubbish, but will push for a fix in remoting 4).


versions.props
```diff
@@ -4,7 +4,7 @@ com.google.code.findbugs:* = 3.0.1
 com.google.code.findbugs:jFormatString = 3.0.0
 com.google.guava:* = 18.0
 com.palantir.baseline:* = 0.16.0
-com.palantir.remoting1:* = 1.0.0
+com.palantir.remoting3:* = 3.3.2
```
schlosna: need 3.3.3

schlosna:
Separate from this, but would really like to pull the tracing stuff out of http-remoting into its own stable versioned thing that tritium & http-remoting both depend upon

j-baker:
Yep! filed https://github.com/palantir/http-remoting/issues/580

versions.props
j-baker: done

PR #51, opened 2017-10-27
add a TaggedMetricRegistry

ellisjoe:


tritium-tags/src/main/java/com/palantir/tritium/tags/MetricName.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.tags;
+
+import java.util.Map;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@Value.Style(
```
markelliot: isn't there a tritium style we can depend on?

markelliot (commented):
so simple!

markelliot:
do we want to have a canonical static registry class?

tritium-tags/src/main/java/com/palantir/tritium/tags/MetricName.java
ellisjoe: nothing else in the project uses Immutables

j-baker:
Can we try to avoid this (the static registry)? It’ll make testing a huge pita unless our api is very careful - i.e. have to do last write wins

markelliot:
the dropwizard-style SharedMetricRegistries thing ends up being pretty gross, would be ideal to find a way to avoid its pitfalls, but not sure we then want everything to have to plumb the registry through every class?

ellisjoe:
I'm fine with one global metric registry. a global registry is required for things implementing apis we don't control (soho, fsclient) as well.

ellisjoe:
thinking we should move this to `tritium-registries` and package `com.palantir.tritium.metrics`. @schlosna sound good?

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricName.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package com.palantir.tritium.metrics;
+
+import java.util.Map;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@Value.Style(
+        jdkOnly = true,
```
schlosna: Since tritium-metrics depends on Guava, we might as well generate the guava immutables now

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricRegistry.java
```diff
@@ -0,0 +1,83 @@
...
+    private static final TaggedMetricRegistry DEFAULT = new TaggedMetricRegistry();
+
+    private final Map<MetricName, Metric> registry = new ConcurrentHashMap<>();
+
+    public TaggedMetricRegistry() {}
+
+    /**
+     * Get the global default {@link TaggedMetricRegistry}.
+     */
+    public TaggedMetricRegistry getDefault() {
```
schlosna: Is the assumption that metric producers will register via this globally accessible singleton registry? I'm not opposed to this as plumbing the old `MetricRegistry` down through many layers is often a challenge. Might want to consider testability and how folks should handle this to prevent poor concurrent unit test interactions.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricRegistry.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public Map<MetricName, Metric> getMetrics() {
+        return Collections.unmodifiableMap(registry);
+    }
+
+    private <T extends Metric> T getOrAdd(MetricName metricName, Class<T> metricClass, Supplier<T> metricSupplier) {
+        Metric metric = registry.computeIfAbsent(metricName, name -> metricSupplier.get());
+        if (!metricClass.isInstance(metric)) {
+            throw new IllegalArgumentException(metricName.name()
+                    + " already used for a different type of metric. tags: " + metricName.tags());
```
schlosna: could include the old & new metric types in message to help debugging

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricRegistry.java
```diff
@@ -0,0 +1,83 @@
...
+    public Meter meter(MetricName metric) {
+        return getOrAdd(metric, Meter.class, Meter::new);
+    }
+
+    public Timer timer(MetricName metric) {
+        return getOrAdd(metric, Timer.class, Timer::new);
+    }
+
+    public Map<MetricName, Metric> getMetrics() {
+        return Collections.unmodifiableMap(registry);
```
schlosna, line 72: Is the intention is that this registry will only contain tagged metrics, and regular `MetricRegistry` contains non-tagged metrics, and we push the non-tagged over to tagged? Do we care about collisions of just the name part of metrics between the two?

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricName.java
markelliot: don't think it's harmful to set jdkOnly

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricRegistry.java
markelliot: so one idea was for classes to grab a metric registry the same way classes grab a logger:

    private static final TaggedMetricRegistry registry = 
        MetricContext.getTaggedRegistry(ClassName.class);

markelliot: might then allow for tests to adjust the `MetricContext` in some smart way.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricRegistry.java
markelliot: we can keep them totally separate and log both?

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricRegistry.java
ellisjoe: think we should only do that if we think sticking the FQN on all metrics is useful

schlosna: I'm assuming that would pre-populate any metrics with a tag like `class: ClassName` (or similar), is that correct? How does that play into API where you provide the `MetricName` that can already include tags, and which one trumps in case of tag name collision?

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricRegistry.java
ellisjoe: +1 to keeping them separate and logging both. I don't think name collisions are an issue here since the new metrics should be uniquely identified by their tags, even if they do have a name clash with a metric in the vanilla registry

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricRegistry.java
markelliot: I thought one of the tags was a class name. The other way to think about it is that this is a way to get the 'right' registry but to still allow it to be statically grabbed.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricRegistry.java
schlosna: We can log both, just want to think through what collisions mean, mostly downstream from this when they feed into systems for analysis. I'd expect that the new tagged MetricName won't collide in general as they should have cleaner names anyway. I think its probably ok and we can warn on the metric extraction side to resolve and fix any potential collisions.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricName.java
```diff
@@ -0,0 +1,37 @@
...
+        get = {"get*", "is*"},
+        overshadowImplementation = true,
+        visibility = Value.Style.ImplementationVisibility.PACKAGE)
+public interface MetricName {
+    String name();
+    Map<String, String> tags();
```
schlosna: I know we talked about this, but probably worth JavaDoc is the assumption that both tag name and value are assumed to be `@Safe` loggable

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricRegistry.java
ellisjoe: done

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricName.java
ellisjoe: done

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricRegistry.java
ellisjoe: the list of tags we've discussed doesn't include class name. I'm also not particularly concerned with a single global registry for now (given we only use a single instance of `MetricRegistry` everywhere right now). If we think class name is a useful tag in general then happy to add the `get(Class)` method

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricRegistry.java
ellisjoe: agreed. think this is a problem today anyway, since two separate services can register metrics with the same name and won't get the `MetricRegistry` validation.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricName.java
```diff
@@ -0,0 +1,49 @@
...
+@Value.Immutable
+@Value.Style(
+        jdkOnly = true,
+        get = {"get*", "is*"},
+        overshadowImplementation = true,
+        visibility = Value.Style.ImplementationVisibility.PACKAGE)
+public interface MetricName {
+
+    /**
+     * General/abstract measure with units (e.g. time of a request).
```
schlosna: nit: we generally don't include units in most metric names, especially timers which typically have units of microseconds

schlosna (approved):
minor JavaDoc nit, but otherwise LGTM

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricName.java
ellisjoe: done. changed the example to an actual metric name as well

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricName.java
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.metrics;
+
+import java.util.Map;
+import org.immutables.value.Value;
+
+@Value.Immutable
```
schlosna: perf nit: Do we want to precompute hashCode via `@Value.Immutable(prehash = true)` to avoid expensive lookups? See https://immutables.github.io/immutable.html#prehashed

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricRegistry.java
schlosna: Actually this method needs to be `static`, correct?

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricName.java
ellisjoe: +1

PR #47, opened 2017-10-25
Add tags to metrics instrumentation and caches

schlosna:
Add tags to cache metric sets
Add tags to MetricsInvocationEventHandler
Optimize Instrumentation.withMetrics empty tags

PR #52, opened 2017-11-01
safe metric name params

ellisjoe:


ellisjoe:
@schlosna with this change I think we'd be alright with a real release of tritium? as @markelliot mentioned in https://github.com/palantir/tritium/pull/47 we probably want to at least leave the existing methods that register things into the original `MetricRegistry`, which means the rest of those changes can be additive. let me know if that doesn't sound right for any reason.

schlosna (approved):

PR #53, opened 2017-11-02
metric registry project

ellisjoe:


tritium-registry/build.gradle
```diff
@@ -0,0 +1,13 @@
+apply from: "${rootDir}/gradle/publish-jars.gradle"
+
+dependencies {
+    processor "org.immutables:value"
+
+    compile 'com.palantir.safe-logging:safe-logging'
+    compile 'io.dropwizard.metrics:metrics-core'
```
markelliot: This is the thing I poked at in remoting.

ellisjoe: as in you want interfaces for all of these things?

markelliot: @uschi2000's concern was that this would end up making impossible diamond deps. I'm not hard over on the split vs. not, except that if we're really going to try to get this to be the 'api' kit then it shouldn't depend on anything but the apis?

ellisjoe: I've split out an interface for the TaggedMetricRegistry so if the remaining concern is depending on metrics-core I'd argue it's fine to depend on this as an api dependency. It's currently used by 60k projects https://github.com/search?q=com.codahale.metrics.MetricRegistry&type=Code&utf8=%E2%9C%93 and things like spark will break if they break their api.

markelliot: not sure I understand what this split buys us, then

ellisjoe: interface split or project split? the project split means remoting doesn't depend on tritium-api or tritium-core. I'm indifferent to the interface split, but don't think it hurts.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
```diff
@@ -14,60 +14,61 @@
...
 import java.util.Collections;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Supplier;
 
-/**
- * Similar to {@link MetricRegistry} but allows tagging of {@link Metric}s.
- */
-public final class TaggedMetricRegistry {
+public final class DefaultTaggedMetricRegistry implements TaggedMetricRegistry {
```
schlosna: Do you want to include the service registry info for this in `META-INF/services/com.palantir.tritium.metrics.registry.TaggedMetricRegistry`?

tritium-registry/build.gradle
schlosna: I'm not sure how much indirection we want to buy. I think this PR gets us enough so that folks that want the minimum can depend on `tritium-registry`, and in apps depend on the full thing and force deps to avoid any diamond issues.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
ellisjoe: since the interface and impl end up in the same jar the service loader should already be able to find the impl, right? can add it for completeness though.

tritium-registry/build.gradle
```diff
@@ -0,0 +1,14 @@
+apply from: "${rootDir}/gradle/publish-jars.gradle"
+
+dependencies {
+    processor 'com.google.auto.service:auto-service'
+    processor 'org.immutables:value'
+
+    compile 'com.palantir.safe-logging:safe-logging'
+    compile 'io.dropwizard.metrics:metrics-core'
```
uschi2000, line 8: this brings in the same dependencies as before, no?

ellisjoe: this only brings in the Metrics (Counter, Timer, etc), not anything else from tritium (tritium-api, tritium-core)
```
compile - Dependencies for source set 'main' (deprecated, use 'implementation ' instead).
+--- com.palantir.safe-logging:safe-logging: -> 0.1.3
\--- io.dropwizard.metrics:metrics-core: -> 3.2.5
     \--- org.slf4j:slf4j-api:1.7.22 -> 1.7.12
```

uschi2000: ok

uschi2000: sorry, didn't right properly

uschi2000:
👍

schlosna (approved):

PR #56, opened 2017-11-06
Fix additional backward compatibility breaks

schlosna:
Addresses backward compatibility breaks introduced by https://github.com/palantir/tritium/pull/50 due to field type change from `com.palantir.tritium.api.functions.LongPredicate` to `java.util.functions.LongPredicate`

Restore types on LoggingInvocationEventHandler constants to `com.palantir.tritium.api.functions.LongPredicate`

Add explicit backward compatibility bridge constructors to `AbstractInvocationEventHandler` and `LoggingInvocationEventHandler`


schlosna:
@uschi2000 for SA

PR #57, opened 2017-11-06
Bridge to back compatibility

schlosna:
Introduce bridge to address backward compatibility issue:

```
java.lang.NoSuchMethodError: com.palantir.tritium.proxy.Instrumentation$Builder.withLogging(Lorg/slf4j/Logger;Lcom/palantir/tritium/event/log/LoggingLevel;Lcom/palantir/tritium/api/functions/LongPredicate;)Lcom/palantir/tritium/proxy/Instrumentation$Builder;
    at com.palantir.atlasdb.util.AtlasDbMetrics.instrument(AtlasDbMetrics.java:76)
```

### DIFF
--- a/tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
+++ b/tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
@@ -30,7 +30,6 @@ import com.palantir.tritium.event.metrics.MetricsInvocationEventHandler;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


ellisjoe (approved):
probably worth `@Deprecate`ing the non `java.util` method?

schlosna:
@ellisjoe yeah, I've hesitated to mark the back-compat bridges `@Deprecate`d because if you use them with one of the common constants, you'd end up actually using the deprecated one and get warnings :( (unless you explicitly cast it to `java.util.functions.LongPredicate` which makes the code look more awful than if you just wrote your own predicate).

ellisjoe:
I see. Yeah no harm leaving it as is then.

PR #68, opened 2017-12-06
Support Removing tagged metrics

### LINKED ISSUE #59: Allow removal of metrics from TaggedMetricRegistry
We need a way to replace metrics in the `TaggedMetricRegistry` similar to https://github.com/palantir/tritium/blob/develop/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java#L188

See https://github.com/palantir/atlasdb/pull/2659#issuecomment-343209721

@hsaraogi for SA

hsaraogi:
Fixes #59. This unblocks adding tag metrics for cassandra-hosts and timelock for AtlasDB.

@tboam for SA.

We should add tests but the class is generally untested :(.

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
@@ -25,6 +25,7 @@ import com.codahale.metrics.Metric;
 import com.codahale.metrics.Timer;
 import com.google.auto.service.AutoService;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Supplier;
@@ -73,6 +74,11 @@ public final class DefaultTaggedMetricRegistry implements TaggedMetricRegistry {
     @Override
     public Map<MetricName, Metric> getMetrics() {
         return Collections.unmodifiableMap(registry);
+    }
+
+    @Override
+    public void removeMetrics(List<MetricName> metricNameList) {
+        metricNameList.forEach(registry::remove);
     }
 
     private <T extends Metric> T getOrAdd(MetricName metricName, Class<T> metricClass, Supplier<T> metricSupplier) {
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
@@ -25,8 +25,8 @@ import com.codahale.metrics.Metric;
 import com.codahale.metrics.Timer;
 import com.google.auto.service.AutoService;
 import java.util.Collections;
-import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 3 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
```diff
@@ -41,4 +42,6 @@
     Counter counter(MetricName metric);
 
     Map<MetricName, Metric> getMetrics();
+
+    void removeMetrics(List<MetricName> metricNameList);
```
schlosna: I think we need to adjust the API to be singular and have a return value to indicate the metric that was removed, if it existed and was successfully removed.

```
    /**
     * Removes the tagged metric with the specified metric name.
     * @param metricName metric name
     * @return the removed metric
     */
    Optional<Metric> remove(MetricName metricName);
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
```diff
@@ -75,6 +76,11 @@ public Timer timer(MetricName metric) {
         return Collections.unmodifiableMap(registry);
     }
 
+    @Override
+    public void removeMetrics(List<MetricName> metricNameList) {
+        metricNameList.forEach(registry::remove);
```
schlosna: see below, but we should use:

```
    @Override
    public Optional<Metric> remove(MetricName metricName) {
        return Optional.ofNullable(registry.remove(metricName));
    }
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
ellisjoe: would prefer this just taking a single `MetricName`

schlosna (changes requested):
We should adjust the API to allow consumers to know whether metric was removed, and also we need to add tests.

schlosna:
@hsaraogi @ellisjoe just pushed some tweaks, please let me know if you're ok and I can merge.

Is there anything else you need after this or should I tag a 0.9.0 release?

ellisjoe:
what's the use case for removing here? seems like cases where you want to removed would lead to a large number of unique series.

schlosna:
@ellisjoe the main use case I've seen is around removing/replacing registered gauges, such as those related to caches (e.g. https://github.com/palantir/tritium/blob/develop/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java#L166 ). Since gauges often hold references back to the thing they're instrumenting (say a cache, executor, or queue), if we dynamically reload or recreate the backing instrumented object, we need to remove the previously registered metric (gauge) and register the new one.

tboam:
The 2 use cases we've come across in Atlas so far are:
1) CassandraClientPool reports metrics tagged with the specific Cassandra node they are talking to.  If Cassandra clients are removed then we want to tidy up the metrics so we no longer report on non-existent hosts.
2) The Timelock leader node wants to report metrics tagged by the client (e.g. catalog, etc).  When the node loses leadership we want to stop reporting the metrics (another node will start reporting them.  In this case we could revert to the Gauge returning `null` if this will prevent values appearing in metrics monitoring.

schlosna (approved):

ellisjoe:
don't think we want to be tagging things dynamically. we've already removed similar tags from witchcraft and remoting because they generate too many unique metrics. can you tag me in the PR where you use `remove` @tboam @hsaraogi ?

tboam:
For Cassandra the metrics PR is here: https://github.com/palantir/atlasdb/pull/2659/files#diff-b88d7ebe9dabc333cf7e4dff942e1830

For Timelock, this was just something I was hacking on today and hit this blocker.  No PR currently.  If there's a way to decline to give a metric (throw Exception from `getValue()`, return `null`) then I can use that.

hsaraogi:
The use-case in the cassandra PR is to allow client connection pools to be closed and re-opened with the tag indicating the right hosts, will clean it up with the tritium bump and add you for review.

PR #76, opened 2018-05-04
Remove findbugs

iamdanfox:
The findbugs plugin was removed from baseline in March 2017 (https://github.com/palantir/gradle-baseline/pull/131), so this repo contains many (unvalidated) findbugs annotations. 

These would be mostly harmless, except that they cause a new `com.palantir.baseline-class-uniqueness` plugin to fail: the findbugs jars are notoriously unhygienic.  `jsr305`, `annotations` and `findbugs-annotations` all define some classes with the same name.

Rather than just ignoring these violations, I figured we might as well fix them at the source :)


schlosna:
For future reference, worth tracking and investigating:

- https://github.com/google/error-prone/issues/44
- https://github.com/google/error-prone/issues/280
- https://github.com/uber/NullAway
- https://checkerframework.org/

PR #77, opened 2018-05-16
Support providing a Supplier for TaggedMetricRegistry

justinuang:
We want to use a SlidingTimeWindowArrayReservoir in our code base:
https://metrics.dropwizard.io/4.0.0/manual/core.html#sliding-time-window-reservoirs

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
@@ -24,6 +24,7 @@ import com.codahale.metrics.Metric;
 import com.codahale.metrics.Timer;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Supplier;
 
 /**
  * Similar to {@link com.codahale.metrics.MetricRegistry} but allows tagging of {@link Metric}s.
@@ -38,6 +39,8 @@ public interface TaggedMetricRegistry {
      */
     Timer timer(MetricName metricName);
 
+    Timer timer(MetricName metricName, Supplier<Timer> timerSupplier);
+
     /**
      * Returns existing or new meter metric for the specified metric name.
      *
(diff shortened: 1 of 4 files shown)


justinuang:
@iamdanfox can I get a review?

iamdanfox:
Defering to @schlosna on this one - I don't actually have write on this repo ;)

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
```diff
@@ -38,6 +39,8 @@
      */
     Timer timer(MetricName metricName);
 
+    Timer timer(MetricName metricName, Supplier<Timer> timerSupplier);
```
schlosna, line 42: I don't know of any other implementations of `TaggedMetricRegistry` than `DefaultTaggedMetricRegistry` but we might want to implement these new methods via Java 8 `default`s here in interface for back compat

schlosna (commented):
Can we add some tests using alternate suppliers (e.g. time window)?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
schlosna: I found one interface that extends `TaggedMetricRegistry` and one implementation of that extension, but I think we're ok to make this interface change without the `default` and just update that one usage.

Would still like to get test coverage over these new methods for hygiene, then happy to merge and tag a 0.10.0.

justinuang:
I added some tests using mocks so we can assert the exact behavior we wanted (supplier being called only once). I wasn't too keen on typing the same test over and over so i pulled out a common test method. Let me know if you have any comments.

schlosna (approved):

justinuang:
Awesome, can we get a release?

schlosna:
@justinuang tagged 0.10.0 and should be released pending https://circleci.com/gh/palantir/tritium/258 build

PR #81, opened 2018-06-18
TaggedMetricRegistry composability

j-baker:
Frequently libraries produce metrics, but the context of usage is not known (e.g. Atlas metrics are coupled to an Atlas client, Spark metrics are tied to a job). In this case, if a caller has multiple metric registries, they can add the child metrics to the parent with a tag, so that they can be distinguished.

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
@@ -24,7 +24,7 @@ import com.codahale.metrics.Meter;
 import com.codahale.metrics.Metric;
 import com.codahale.metrics.Timer;
 import com.google.auto.service.AutoService;
-import java.util.Collections;
+import com.google.common.collect.ImmutableMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
@@ -36,6 +36,7 @@ public final class DefaultTaggedMetricRegistry implements TaggedMetricRegistry {
     private static final TaggedMetricRegistry DEFAULT = new DefaultTaggedMetricRegistry();
 
     private final Map<MetricName, Metric> registry = new ConcurrentHashMap<>();
+    private final Map<Map<String, String>, TaggedMetricSet> taggedRegistries = new ConcurrentHashMap<>();
 
     public DefaultTaggedMetricRegistry() {}
 
@@ -93,12 +94,32 @@ public final class DefaultTaggedMetricRegistry implements TaggedMetricRegistry {
 
     @Override
     public Map<MetricName, Metric> getMetrics() {
-        return Collections.unmodifiableMap(registry);
+        ImmutableMap.Builder<MetricName, Metric> result = ImmutableMap.<MetricName, Metric>builder()
+                .putAll(registry);
+        taggedRegistries.forEach((tags, metrics) -> metrics.getMetrics()
+                .forEach((metricName, metric) -> result.put(
+                        MetricName.builder()
+                                .from(metricName)
+                                .putAllSafeTags(tags)
+                                .build(),
+                        metric)));
+
+        return result.build();
     }
 
     @Override
     public Optional<Metric> remove(MetricName metricName) {
         return Optional.ofNullable(registry.remove(metricName));
+    }
+
+    @Override
+    public void addMetrics(Map<String, String> tags, TaggedMetricSet other) {
+        taggedRegistries.put(tags, other);
+    }
+
+    @Override
+    public void removeMetrics(Map<String, String> tags) {
+        taggedRegistries.remove(tags);
     }
 
     private <T extends Metric> T getOrAdd(MetricName metricName, Class<T> metricClass, Supplier<T> metricSupplier) {
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
@@ -25,6 +25,7 @@ import com.codahale.metrics.Metric;
 import com.codahale.metrics.Timer;
 import com.google.auto.service.AutoService;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
@@ -36,7 +37,7 @@ public final class DefaultTaggedMetricRegistry implements TaggedMetricRegistry {
     private static final TaggedMetricRegistry DEFAULT = new DefaultTaggedMetricRegistry();
 
     private final Map<MetricName, Metric> registry = new ConcurrentHashMap<>();
-    private final Map<Map<String, String>, TaggedMetricSet> taggedRegistries = new ConcurrentHashMap<>();
+    private final Map<Map.Entry<String, String>, TaggedMetricSet> taggedRegistries = new ConcurrentHashMap<>();
 
     public DefaultTaggedMetricRegistry() {}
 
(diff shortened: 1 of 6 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DropwizardTaggedMetricSet.java
```diff
@@ -0,0 +1,37 @@
...
+public final class DropwizardTaggedMetricSet implements TaggedMetricSet {
+    private final MetricRegistry metricRegistry;
+
+    public DropwizardTaggedMetricSet(MetricRegistry metricRegistry) {
+        this.metricRegistry = metricRegistry;
+    }
+
+    @Override
+    public Map<MetricName, Metric> getMetrics() {
+        return metricRegistry.getMetrics().entrySet().stream()
```
j-baker: I could avoid the copy here if we think it's important by implementing ForwardingMap.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
```diff
@@ -36,6 +36,7 @@
     private static final TaggedMetricRegistry DEFAULT = new DefaultTaggedMetricRegistry();
 
     private final Map<MetricName, Metric> registry = new ConcurrentHashMap<>();
+    private final Map<Map<String, String>, TaggedMetricSet> taggedRegistries = new ConcurrentHashMap<>();
```
j-baker: probably not _quite right_, since will want to shove probably two MetricSets in for each AtlasDB, also probably shouldn't be passing in an empty map here. Maybe just a Set of Pair type structure is the right thing here.

schlosna: Yeah, this feels odd and I worry about someone putting a mutable `Map<String, String>` in as a key and causing all kinds of chaos.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DropwizardTaggedMetricSet.java
```diff
@@ -0,0 +1,37 @@
...
+                .collect(toMap(entry -> MetricName.builder().safeName(entry.getKey()).build(), Map.Entry::getValue));
```
schlosna: are we ok to assume all metric names are safe here?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
```diff
@@ -81,13 +81,6 @@
 
     Counter counter(MetricName metricName, Supplier<Counter> counterSupplier);
 
-    /**
-     * Returns a map of registered metrics.
-     *
-     * @return map of registered metrics
-     */
-    Map<MetricName, Metric> getMetrics();
```
schlosna, line 89: I haven't looked at how painful the API break would be, do you have a sense?

schlosna (changes requested):
Hey James, interested to see how you're planning on using this, maybe sketch out some tests showing intent? I'm wondering if we want to have a single canonical metric registry that other partitions can feed into and auto adjust MetricName to include that partition's tags.

For example if we have service Foo with a single registry, and Atlas clients Bar and Baz, we'd inject a registry that auto adds tag `client=Bar` and `client=Baz` respectively, Atlas adds a counter `qux` and the underlying Foo registry has two counters `qux;client=Bar` and `qux;client=Baz`.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
j-baker: that's not actually an API break; it's just moved into the super interface

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DropwizardTaggedMetricSet.java
j-baker: yes - they'll get logged internally in the same way as they currently do. The reason this was called safe name was to make clear that they have to be safe - but that's a baseline requirement.

j-baker:
hey dave - the motivation here is basically: we have multiple atlas clients and want to be able to control how they're reported.

I don't think libraries should be using static metrics. This is for a few reasons, but I've grown conviction in this workflow over time. It's painful for producers because they have to understand the metrics with regards to a lifecycle. For example, with AtlasDB they have to go through the workflows of removing metrics upon failed initialisation tasks, so as to not leave now-unused gauges around, and the complexity of this is nontrivial.

For consumers, it's painful because now the metrics that are being recorded are defined by the producer. For example, AtlasDB made the bad assumption that no one has more than one AtlasDB hanging around, and this was very very painful to change (2000 lines +-), because the metrics needed to be plumbed through (even though it was only like 100 lines of additions overall).

It's also painful for aggregators, since it makes the incremental cost of adding metrics almost zero, which means that tracking which metrics exist and are useful is essentially impossible. For example, AtlasDB by default adds around 1500 metrics. This makes it very difficult to have a coherent metric reporting story (which metrics are even valuable? which should I use?) because they're so numerous.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
```diff
@@ -96,4 +89,20 @@
...
+     *
+     * @param tags the tags which will be added to all the metrics in the metric set
+     * @param metrics the metrics which should be added
+     */
+    void addMetrics(String safeTagName, String safeTagValue, TaggedMetricSet metrics);
+
+    /**
+     * Removes a TaggedMetricsSet added via addMetrics from this metrics set.
+     */
+    void removeMetrics(String safeTagName, String safeTagValue);
```
schlosna: API question: should we return either a `boolean` letting consumer know if it was removed, or the removed `TaggedMetricSet`?

schlosna (commented):
Hey James, makes sense and agreed AtlasDB should have been injecting the appropriately scoped metrics registry from the beginning.

Had one minor API question before merging.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
j-baker: I now return an optional taggedmetricset to be consistent

j-baker: with remove()

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
```diff
@@ -93,14 +97,38 @@ public Timer timer(MetricName metricName, Supplier<Timer> timerSupplier) {
 
     @Override
     public Map<MetricName, Metric> getMetrics() {
-        return Collections.unmodifiableMap(registry);
+        return Stream.concat(registry.entrySet().stream(),
+                taggedRegistries.entrySet().stream().flatMap(entry -> addTag(
+                        entry.getKey().getKey(),
+                        entry.getKey().getValue(),
+                        entry.getValue().getMetrics().entrySet().stream())))
+                .collect(toMap(Map.Entry::getKey, Map.Entry::getValue, (a, b) -> a));
```
schlosna: you got checkstyled

```
ParameterNameCheck - com.palantir.tritium.metrics.registry.DefaultTaggedMetricRegistry - checkstyleMain
DefaultTaggedMetricRegistry.java:105: Parameter name 'b' must match pattern '^[a-z][a-zA-Z0-9]+$'.

ParameterNameCheck - com.palantir.tritium.metrics.registry.DefaultTaggedMetricRegistry - checkstyleMain
DefaultTaggedMetricRegistry.java:105: Parameter name 'a' must match pattern '^[a-z][a-zA-Z0-9]+$'.
```

schlosna (approved):

schlosna:
Approved pending checkstyle fix

schlosna:
@j-baker go ahead and update the description when you're happy to squash & merge

jiahuijiang:
@j-baker Why do we only allow one tag for each metricSet here not a map of tags and their values?

PR #93, opened 2018-07-24
Add metric for maximum size of cache

fsamuel-bs:
We use a Caffeine to cache all rows of a small view of our DB, since we have an endpoint which requires us to fetch such view regularly.

The size of the cache is also configurable in the configuration of the service. We're planning on adding alerts if the cache is close to being full (in order to evaluate if we can and should increase the cache). In order to add that alert, we need the metric of the maximum size that the cache was configured to use.

I could instrument this locally, on my service, but also think this might be useful in other services which have configurable caches thus I'm adding this here.

### DIFF
--- a/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheMetricSet.java
+++ b/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheMetricSet.java
@@ -95,6 +95,15 @@ final class CaffeineCacheMetricSet implements MetricSet {
                     }
                 });
 
+        cache.policy().eviction().ifPresent(eviction -> cacheMetrics.put(
+                cacheMetricName("maximum", "size"),
+                new CachedGauge<Long>(clock, 500, TimeUnit.MILLISECONDS) {
+                    @Override
+                    protected Long loadValue() {
+                        return eviction.getMaximum();
+                    }
+                }));
+
         cacheMetrics.put(cacheMetricName("request", "count"),
                 derivedGauge(CacheStats::requestCount));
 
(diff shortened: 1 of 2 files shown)


tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheMetricSet.java
```diff
@@ -95,6 +95,15 @@ protected Long loadValue() {
...
                 });
 
+        cache.policy().eviction().ifPresent(eviction -> cacheMetrics.put(
+                cacheMetricName("maximum", "size"),
+                new CachedGauge<Long>(clock, 500, TimeUnit.MILLISECONDS) {
+                    @Override
+                    protected Long loadValue() {
+                        return eviction.getMaximum();
+                    }
+                }));
```
schlosna: do we want to report weighted size as well?

```
        cache.policy().eviction().ifPresent(eviction -> {
            eviction.weightedSize().ifPresent(weightedSize -> cacheMetrics.put(
                    cacheMetricName("weighted", "size"),
                    (Gauge<Long>) () -> weightedSize));
            cacheMetrics.put(
                    cacheMetricName("maximum", "size"),
                    new CachedGauge<Long>(clock, 500, TimeUnit.MILLISECONDS) {
                        @Override
                        protected Long loadValue() {
                            return eviction.getMaximum();
                        }
                    });
        });
```

schlosna (commented):
I pushed tweak to fixup unit tests

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheMetricSet.java
fsamuel-bs: I don't need it now, but imagined it could be useful in the future. Done.

schlosna (approved):

PR #94, opened 2018-07-24
Make publish depend on bintrayUpload

fsamuel-bs:


fsamuel-bs:
We used circle 1, and on the circle config we specified directly to `bintrayUpload`. Circle 2 runs `publish`, thus we need to set the dependency explicitly. Same as https://github.com/palantir/http-remoting-api/blob/develop/gradle/publish-jar.gradle#L27.

schlosna (approved):

PR #106, opened 2018-09-27
Upgrade Guava to 23.6.1-jre

schlosna:
[dev-break] Upgrade from minimum compatible Guava 18.0 to 23.6.1-jre

Upgrade gradle-baseline to 0.32.3


schlosna:
Closes #104

PR #134, opened 2018-10-24
Tracing handler cleans up trace state after creating a root span

carterkozak:
Otherwise multiple root spans will be created.


carterkozak:
recreated :-)

schlosna (approved):

PR #150, opened 2018-12-05
Atomic remove.

jkozlowski:


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
```diff
@@ -123,6 +123,12 @@ public void addMetrics(String safeTagName, String safeTagValue, TaggedMetricSet
         return Optional.ofNullable(taggedRegistries.remove(Maps.immutableEntry(safeTagName, safeTagValue)));
     }
 
+    @Override
+    public boolean removeMetrics(
+            String safeTagName, String safeTagValue, TaggedMetricSet metrics) {
+        return taggedRegistries.remove(Maps.immutableEntry(safeTagName, safeTagValue), metrics);
```
schlosna, line 129: Do we need to implement `equals`/`hashCode` on `TaggedMetricSet` impls?

jkozlowski: I don't think so, it's the same contract as Map#remove. The point is you provide the same instance to remove as to add, so it's up to you.

jkozlowski: Basically this is fixing the fact that there isn't really a safe way to remove and then add if you can't guarantee the ordering; like if you have a cache of some objects that produce the metrics, and upon deleting from the cache you unregister the taggedmetrics, because usually cache removal notification is asynchronous, you can't guarantee that the object hasn't been recreated already.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
```diff
@@ -108,4 +108,12 @@
...
      */
     Optional<TaggedMetricSet> removeMetrics(String safeTagName, String safeTagValue);
+
+    /**
+     * Removes a TaggedMetricsSet added via addMetrics from this metrics set, if currently registered to this metric
+     * set.
+     *
+     * @return true if value was removed
+     */
+    boolean removeMetrics(String safeTagName, String safeTagValue, TaggedMetricSet metrics);
```
schlosna, line 121: can we document the assumption around equality comparison check for `TaggedMetricSet` a la https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#remove-java.lang.Object-java.lang.Object-

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
schlosna: makes sense, let's just document the contract in the interface and should be good to go

schlosna (approved):

PR #151, opened 2018-12-06
Add TaggedMetricsServiceInvocationEventHandler

mglazer:

PR #153, opened 2018-12-10
Fix NPE with tagged metrics instrumentation

schlosna:
Fixes NPE introduced in #151 and incorporates test coverage

### DIFF
--- a/tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
+++ b/tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
@@ -19,6 +19,7 @@ package com.palantir.tritium.proxy;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.codahale.metrics.MetricRegistry;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.palantir.tritium.api.event.InstrumentationFilter;
 import com.palantir.tritium.event.InstrumentationFilters;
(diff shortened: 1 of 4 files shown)


tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
```diff
@@ -143,18 +144,18 @@ private Builder(Class<T> interfaceClass, U delegate) {
...
-         * @param globalPrefix - Metrics name prefix to be used
+         * @param prefix - Metrics name prefix to be used
          * @return - InstrumentationBuilder
          */
-        public Builder<T, U> withTaggedMetrics(TaggedMetricRegistry metricRegistry, String globalPrefix) {
+        public Builder<T, U> withTaggedMetrics(TaggedMetricRegistry metricRegistry, String prefix) {
             checkNotNull(metricRegistry, "metricRegistry");
-            this.handlers.add(new TaggedMetricsServiceInvocationEventHandler(
-                    metricRegistry, globalPrefix));
+            String serviceName = (Strings.isNullOrEmpty(prefix)) ? interfaceClass.getName() : prefix;
```
mglazer, line 152: I've been favoring `.getSimpleName()` since it tends to be more representative of the service.

tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
```diff
@@ -143,18 +144,18 @@ private Builder(Class<T> interfaceClass, U delegate) {
...
+            this.handlers.add(new TaggedMetricsServiceInvocationEventHandler(metricRegistry, serviceName));
             return this;
         }
 
         public Builder<T, U> withTaggedMetrics(TaggedMetricRegistry metricRegistry) {
-            return withTaggedMetrics(metricRegistry, null);
+            return withTaggedMetrics(metricRegistry, "");
```
mglazer, line 158: Oops.

mglazer (approved):

tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
schlosna: This was just a nit fix as blank would previously also have thrown as well in https://github.com/palantir/tritium/blob/develop/tritium-core/src/main/java/com/palantir/tritium/event/InstrumentationProperties.java#L42 as you need a non-null && non-blank service name.

tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
schlosna: I went with full name fallback default because its a bit more descriptive as the `service-name` will be simple name: `com.palantir.tritium.test.TestInterface:{service-name=TestInterface, endpoint=test}`

carterkozak:
Thanks!

PR #177, opened 2019-01-31
Update io.dropwizard.metrics:metrics-core to 4.0.5

nartmal:
We forced our dw metrics version from 3.2.6 -> 4.0.5 so we could take dropwizard 1.3.x since there were some java breaks (namely the one biting me is the move from ``com.codahale.metrics.*`` to ``com.codahale.metrics.jvm.*``). 

It looks like it isn't an issue in this repo (no usages of metrics-jvm), but it is for downstream consumers. My preference is to start here and work my way downstream and get everyone to metrics-core 4.0.5 and metrics-jvm 4.0.5


schlosna:
@nartmal thanks, I will take a look in a bit. I will probably cut a last release from Dropwizard metrics 3.x and then we can merge this on a new minor version as the DW metrics 4.x has a couple other breaks (e.g. jmx moves to a separate module out of metrics-core, and some other package changes)

schlosna:
@cakofony do you mind taking a look to see if bumping the metrics-core dependency will cause any downstream pain for your projects?

See also https://metrics.dropwizard.io/4.0.0/about/release-notes.html

carterkozak:
We're already using it in the large internal project, wc has held off because of some package names, but we can provide a substitute shim library to unblock code built assuming 3.x if it becomes a problem.
I don't know of any real technical blockers.
@robert3005 may have specific examples.

robert3005:
Spark and Hadoop break, can you merge bump there first?

schlosna:
@nartmal wanted to check to see if you had a chance to test out the dropwizard metrics bumps from 3.2.x to 4.0.x in Spark and Hadoop? As Rob mentioned, we probably have to land those before Tritium can bump.

- [Spark is currently on `3.2.6`](https://github.com/palantir/spark/blob/master/pom.xml#L147)
- [Hadoop is currently on `3.2.4`](https://github.com/palantir/hadoop/blob/branch-2.9.2/hadoop-project/pom.xml#L1028)

nartmal:
sorry, I should have time next week

nartmal:
speaking with @iamdanfox and @robert3005 , taking 4.0 isn't worth the effort it seems like.

5.0-rc is out, so we might just be better off waiting for that instead. LMK if you feel differently.

schlosna:
Upgraded to metrics-core 4.1.1 minimum in https://github.com/palantir/tritium/pull/1294

PR #181, opened 2019-02-06
Invoke onSuccess and onFailure handlers properly for async methods

ellisjoe:


tritium-lib/src/test/java/com/palantir/tritium/proxy/InvocationEventProxyTest.java
```diff
@@ -138,6 +138,11 @@ public void onFailure(@Nullable InvocationContext context, @Nonnull Throwable ca
         assertThat(throwable).isSameAs(expected);
     }
 
+    @Test
+    public void testInstrumentCompletableFuture() {
+
+    }
```
carterkozak: Did you intend to commit this?

ellisjoe: haha nope. removed :)

tritium-lib/src/main/java/com/palantir/tritium/proxy/InvocationEventProxy.java
```diff
@@ -116,14 +121,52 @@ public final Object instrumentInvocation(Object instance, Method method, Object[
...
             throw handleOnFailure(context, e.getCause());
         } catch (Throwable t) {
             throw handleOnFailure(context, t);
         }
     }
 
+    private Object handleResult(InvocationContext context, Object result) {
+        if (result instanceof ListenableFuture) {
+            return handleFuture(context, (ListenableFuture) result);
+        } else if (result instanceof CompletableFuture) {
```
carterkozak: Should use `CompletionStage` over `CompletableFuture`

carterkozak (changes requested):
I think this will break tracing-java and any other handler which sets and removes thread-local state. We need some way to opt into asynchronous callbacks to avoid breaking existing implementations.

There's a separate related conversation around how we want to support tracing-java for asynchronous operations.

ellisjoe:
@cakofony cool. Let's find some time to work through the current issues and what a better solution would look like

tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
```diff
@@ -51,6 +51,16 @@ private CompositeInvocationEventHandler(List<InvocationEventHandler<InvocationCo
...
     }
 
+    @Override
+    public boolean asyncSupport() {
+        for (InvocationEventHandler<?> handler : handlers) {
+            if (!handler.asyncSupport()) {
+                return false;
+            }
+        }
+        return true;
```
carterkozak, line 61: Tritium bundles all handlers together using the composite handler, I think this will disable async support if the tracing handler is enabled.

tritium-lib/src/main/java/com/palantir/tritium/proxy/InvocationEventProxy.java
```diff
@@ -87,13 +88,19 @@ public String toString() {
     private boolean isEnabled(Object instance, Method method, Object[] args) {
         try {
             return eventHandler.isEnabled()
-                    && filter.shouldInstrument(instance, method, args);
+                    && filter.shouldInstrument(instance, method, args)
+                    && (eventHandler.asyncSupport() || !isAsync(method));
```
carterkozak, line 92: I think we want to remove this line. Tracing doesn't support async operations, but the tracing thread state is passed to executors, which will provide valuable information.

tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
carterkozak: I not sure CompositeInvocationEventHandler is the right abstraction for InvocationEventProxy any longer given this change.

schlosna:
@ellisjoe @cakofony interested in your thoughts on the following questions:

- What async return types do we want to support?
    - `CompletionStage` / `CompletableFuture`
    - `ListenableFuture`
    - `Future`?
    - `InputStream`?
    - [`ClosableIterator`](https://github.com/palantir/atlasdb/blob/develop/atlasdb-commons/src/main/java/com/palantir/common/base/ClosableIterator.java)?
- What do we want the semantics to be if some registered handlers support async, while others do not?
- Do we want to capture anything to indicate completion (success/failure) for the instrumented method itself before/after the async completion?

carterkozak:
I can imagine requests for https://projectreactor.io and http://reactivex.io types. Perhaps we should provide special handling for retrofit in conjure-java-runtime without making changes to tritium.

> What do we want the semantics to be if some registered handlers support async, while others do not?

All handlers are expected be invoked as though they're the only handler registered, regardless of configuration. We shouldn't change this.

ellisjoe:
Hey so I'm gonna have to punt on working through the rest of the changes here. At the end of the day it doesn't actually fix my issue which is that I need tracing to support async invocations so I'm gonna close this for now and can come back later or someone else can feel free to pick it up.

PR #191, opened 2019-02-22
fix: Add the type parameter to gauge in TaggedMetricRegistry

rubenfiszel:


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
```diff
@@ -68,7 +68,7 @@
      * @return gauge metric
      */
     // This differs from MetricRegistry and takes the Gauge directly rather than a Supplier<Gauge>
-    Gauge gauge(MetricName metricName, Gauge gauge);
+    <T> Gauge<T> gauge(MetricName metricName, Gauge<T> gauge);
```
iamdanfox, line 71: what effect does this have on ABI compatibility? If people compiled against the old one, what will happen if the new jar just appears on their classpath??

schlosna: In terms of binary compatibility, there is no change as the genetics are erased when compiled, so the existing class files only contain the raw `Gauge` type.

There’s a potential source incompatibility if one had specified an incorrect type parameter.

schlosna (approved):

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
schlosna: If there is a source incompatibility, it is likely a bug in the existing code

PR #193, opened 2019-02-26
 Use gradle-consistent-versions

schlosna:
Utilize https://github.com/palantir/gradle-consistent-versions to manage locked dependency version consistency.


gradle/publish-bom.gradle
```diff
@@ -1,7 +1,7 @@
 publishing {
     publications {
         nebulaBom(MavenPublication) {
-            nebulaDependencyManagement.fromConfigurations { configurations.compile }
+            configurations.compile
```
iamdanfox: beware that we haven't ironed out kinks with gradle-consistent-versions & BOM publishing yet, so if you do a `publishToMavenLocal` you may find the bom is empty.

We had one PR in flight here: https://github.com/palantir/gradle-consistent-versions/pull/23, but there's also a different approach internally in the foundry-spark repo too..

schlosna: good to know, was curious to test this out. FWIW I don't know that anyone is actually consuming the tritium BOM, and ideally they wouldn't need to.

schlosna:
@iamdanfox curious if you think gradle-consistent-versions 1.4.0 is good to go now. I noticed the BOM feature hasn't landed yet, but think its probably fine for tritium for now.

iamdanfox:
Yep, we're definitely benefitting from it even without the BOM stuff.  Likely worth doing a `publishToMavenLocal` just to sanity check everything before merging, but otherwise looks good to me. (As more of your consumers pick up GCV, BOMs become less important)

FWIW, if you name your thingy `publish-jar.gradle` instead of just `publish.gradle` then excavator will come round and keep it up to date:

https://github.com/palantir/tracing-java/blob/develop/gradle/publish-jar.gradle

important snippet to fill in version numbers in the POM;
```gradle
publishing {
    publications.withType(MavenPublication).configureEach {
        versionMapping {
            allVariants {
                fromResolutionResult()
            }
        }
    }
}
```

PR #209, opened 2019-03-21
Exclude method arguments from invocation context toString()

schlosna:
Method arguments may have expensive `toString()` implementation, so exclude them from `DefaultInvocationContext.toString()`


tritium-core/src/main/java/com/palantir/tritium/event/DefaultInvocationContext.java
```diff
@@ -73,8 +72,9 @@ public final Method getMethod() {
     @Override
     @SuppressWarnings("DesignForExtension")
     public String toString() {
-        return "DefaultInvocationContext [startTimeNanos=" + startTimeNanos + ", instance=" + instance + ", method="
```
carterkozak, line 76: I'm curious where this was called

PR #210, opened 2019-03-21
Avoid logging method arguments

schlosna:
Follow on to #209 to avoid potentially expensive arguments toString().

### DIFF
--- a/tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
+++ b/tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
@@ -22,7 +22,6 @@ import com.google.common.collect.ImmutableList;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
 import java.lang.reflect.Method;
-import java.util.Arrays;
 import java.util.List;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
```diff
@@ -120,12 +119,10 @@ public String toString() {
...
             Object instance, Method method, Object[] args, Exception exception) {
-        logger.warn("Exception handling preInvocation({}): "
-                        + "invocation of {}.{} with arguments {} on {} threw: {}",
+        logger.warn("Exception handling preInvocation({}): invocation of {}.{} on {} threw: {}",
                 UnsafeArg.of("handler", handler),
                 SafeArg.of("class", method.getDeclaringClass().getCanonicalName()),
                 SafeArg.of("method", method.getName()),
-                UnsafeArg.of("args", Arrays.toString(args)),
                 UnsafeArg.of("instance", instance),
                 UnsafeArg.of("exception", exception),
```
carterkozak: unrelated to this change, but this line can be removed, and we can log `"Exception handling preInvocation({}): invocation of {}.{} on {} threw"`, the stack trace already provides more information.

carterkozak (approved):
Change looks good, noted one oddity but it's not related to this change.

tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
```diff
@@ -150,10 +148,9 @@ private static void handleFailure(InvocationEventHandler<?> handler,
         try {
             handler.onFailure(context, cause);
         } catch (RuntimeException exception) {
-            logger.warn("Exception handling onFailure({}, {}): {}",
+            logger.warn("Exception handling onFailure({}, {})",
```
carterkozak, line 151: Thanks!

PR #211, opened 2019-03-22
Enable safe logging preconditions

schlosna:
Enables `PreferSafeLoggingPreconditions:ERROR` check and converts from Guava to log-safe preconditions checks

### DIFF
--- a/tritium-slf4j/build.gradle
+++ b/tritium-slf4j/build.gradle
@@ -6,6 +6,7 @@ dependencies {
     api project(':tritium-core')
 
     implementation 'com.google.guava:guava'
+    implementation 'com.palantir.safe-logging:preconditions'
     implementation 'com.palantir.safe-logging:safe-logging'
     implementation 'org.slf4j:slf4j-api'
 
--- a/tritium-slf4j/src/main/java/com/palantir/tritium/event/log/LoggingInvocationEventHandler.java
+++ b/tritium-slf4j/src/main/java/com/palantir/tritium/event/log/LoggingInvocationEventHandler.java
@@ -16,9 +16,8 @@
 
 package com.palantir.tritium.event.log;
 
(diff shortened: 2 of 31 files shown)


tritium-slf4j/build.gradle
```diff
@@ -6,6 +6,7 @@ dependencies {
     api project(':tritium-core')
 
     implementation 'com.google.guava:guava'
+    implementation 'com.palantir.safe-logging:preconditions'
     implementation 'com.palantir.safe-logging:safe-logging'
```
carterkozak, line 10: nit: can remove this line

carterkozak: nvm, probably safest not to since we reference classes from both directly

carterkozak (approved):
lgtm. We need to be careful about using preconditions with SafeArg/UnsafeArg parameters on hot paths where we're better off using a conditional and throwing the exception ourselves. I don't see anything matching that pattern, and the error-prone check will only fail for uses that can be migrated without changing the exception message.

tritium-slf4j/src/main/java/com/palantir/tritium/event/log/LoggingInvocationEventHandler.java
```diff
@@ -192,9 +191,6 @@ static String generateMessagePattern(int argCount) {
             message.append("{}");
         }
         message.append(") took {}ms");
-        Preconditions.checkState(estimatedSize == message.length(),
-                "Incorrect estimated size for %s arguments, expected %s, was %s",
-                argCount, estimatedSize, message.length());
```
schlosna, line 197: this was the one potential hot path and this check was more of a perf optimization sanity check that is not needed

schlosna (commented):
yeah, I checked all the places explicitly creating `Arg` and they aren't on hot paths.

PR #212, opened 2019-03-23
More error prone checks, cleanup, and improved test coverage

schlosna:
* Enable `UnusedVariable:ERROR`
* Enable `JavaTimeDefaultTimeZone:ERROR`
* Add JavaTracingTracerTest
* Enable unchecked type linting
* Include `@Nonnull` annotations on handler implementations
* Clarify API and test method visibility
* Remove extraneous generics from InvocationEventProxy
* Improve `tritium-metrics` test coverage
* `DefaultTaggedMetricRegistry` uses safe-logging preconditions
* Update default included benchmarks

### DIFF
--- a/tritium-lib/src/main/java/com/palantir/tritium/proxy/InvocationEventProxy.java
+++ b/tritium-lib/src/main/java/com/palantir/tritium/proxy/InvocationEventProxy.java
@@ -37,8 +37,7 @@ import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-abstract class InvocationEventProxy<C extends InvocationContext>
-        extends AbstractInvocationHandler implements InvocationHandler {
+abstract class InvocationEventProxy extends AbstractInvocationHandler implements InvocationHandler {
 
     private static final Logger logger = LoggerFactory.getLogger(InvocationEventProxy.class);
 
@@ -192,11 +191,11 @@ abstract class InvocationEventProxy<C extends InvocationContext>
 
     static void logInvocationWarning(String event, Object instance, Method method, Throwable cause) {
         if (logger.isWarnEnabled()) {
-            logger.warn("{} exception handling {} invocation of {}.{} on {}",
+            logger.warn("{} exception handling {} invocation of {} {} on {}",
                     safeSimpleClassName("throwable", cause),
                     SafeArg.of("event", event),
-                    SafeArg.of("class", method.getDeclaringClass()),
-                    SafeArg.of("method", method.getName()),
+                    SafeArg.of("class", method.getDeclaringClass().getName()),
+                    SafeArg.of("method", method),
                     UnsafeArg.of("instance", instance),
                     cause);
         }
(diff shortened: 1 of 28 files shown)


tritium-lib/src/main/java/com/palantir/tritium/proxy/InvocationEventProxy.java
```diff
@@ -185,11 +191,11 @@ private static void logInvocationWarningOnFailure(@Nullable InvocationContext co
...
     static void logInvocationWarning(String event, Object instance, Method method, Throwable cause) {
         if (logger.isWarnEnabled()) {
-            logger.warn("{} exception handling {} invocation of {}.{} on {}",
+            logger.warn("{} exception handling {} invocation of {} {} on {}",
                     safeSimpleClassName("throwable", cause),
                     SafeArg.of("event", event),
-                    SafeArg.of("class", method.getDeclaringClass()),
-                    SafeArg.of("method", method.getName()),
+                    SafeArg.of("class", method.getDeclaringClass().getName()),
+                    SafeArg.of("method", method),
```
carterkozak, line 198: +1 passing the method object here, since method.toString provides more information. I think this will include the method declaring class type.

Should the value of `class` use `instance.getClass() /* maybe .getName() */` instead of the method declaring class to provide more information describing the implementation?

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/TaggedMetricsServiceInvocationEventHandler.java
```diff
@@ -67,7 +67,10 @@ private static BooleanSupplier getEnabledSupplier(final String serviceName) {
     }
 
     @Override
-    public final InvocationContext preInvocation(Object instance, Method method, Object[] args) {
+    public final InvocationContext preInvocation(
```
carterkozak, line 70: Should we annotate the return of `preInvocation` as `@Nullable`?

carterkozak (approved):
Change looks good to me, always supportive of moar linting

tritium-lib/src/main/java/com/palantir/tritium/proxy/InvocationEventProxy.java
schlosna: will tweak this in a followup commit that does a bunch of random cleanup and some JIT optimizations

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/TaggedMetricsServiceInvocationEventHandler.java
schlosna: So the `com.palantir.tritium.event.InvocationEventHandler#preInvocation` API actually requires this to return a non-null, though implementations can potentially throw, in which case we protect against the throwing case, and skip any handlers whose `preInvocation` and treat them as disabled. A followup PR will make this more explicit and take advantage of that fact.

PR #231, opened 2019-04-10
Add instrumented executors

pkoenig10:
Adds `MetricRegistries` methods for instrumenting `ExecutorService` and `ScheduledExecutorService` with tagged metrics.

### DIFF
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
@@ -30,6 +30,7 @@ import com.google.common.cache.Cache;
 import com.google.common.collect.ImmutableSet;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
@@ -37,6 +38,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.Supplier;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
@@ -204,6 +207,29 @@ public final class MetricRegistries {
         CacheMetricSet.create(cache, name, clock)
                 .getMetrics()
                 .forEach((key, value) -> registerWithReplacement(registry, key, value));
+    }
+
+    public static ScheduledExecutorService instrument(
+            TaggedMetricRegistry registry,
+            ScheduledExecutorService delegate,
+            String name) {
+        return new TaggedMetricsScheduledExecutorService(
+                checkNotNull(delegate, "delegate"),
+                checkNotNull(registry, "registry"),
+                checkNotNull(name, "name"));
+    }
+
+    public static ExecutorService instrument(
+            TaggedMetricRegistry registry,
+            ExecutorService delegate,
+            String name) {
+        if (delegate instanceof ScheduledExecutorService) {
+            return instrument(registry, (ScheduledExecutorService) delegate, name);
+        }
+        return new TaggedMetricsExecutorService(
+                checkNotNull(delegate, "delegate"),
+                checkNotNull(registry, "registry"),
+                checkNotNull(name, "name"));
     }
 
     /**
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
@@ -97,7 +97,6 @@ public final class MetricRegistries {
         return DateTimeFormatter.ISO_ZONED_DATE_TIME.format(ZonedDateTime.now(ZoneOffset.UTC));
     }
 
-    @SuppressWarnings("unchecked")
     static <T extends Metric> T getOrAdd(MetricRegistry metrics, String name, MetricBuilder<T> builder) {
         Metric existingMetric = tryGetExistingMetric(metrics, name);
         if (existingMetric == null) {
@@ -111,7 +110,6 @@ public final class MetricRegistries {
         return checkNotNull(metrics, "metrics").getMetrics().get(checkNotNull(name));
     }
 
-    @SuppressWarnings("unchecked")
     private static <T extends Metric> T addMetric(
             MetricRegistry metrics,
             String name,
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
@@ -207,6 +207,17 @@ public final class MetricRegistries {
                 .forEach((key, value) -> registerWithReplacement(registry, key, value));
     }
 
+    /**
+     * Returns an instrumented {@link ScheduledExecutorService} that monitors the number of tasks submitted, running,
+     * completed and also keeps a {@link com.codahale.metrics.Timer} for the task duration. Similar to
+     * {@link com.codahale.metrics.InstrumentedScheduledExecutorService}, but produces tagged metrics to the
+     * specified {@link TaggedMetricRegistry}.
+     *
+     * @param registry tagged metric registry
+     * @param delegate executor service to instrument
+     * @param name executor service name
+     * @return instrumented executor service
+     */
     public static ScheduledExecutorService instrument(
             TaggedMetricRegistry registry,
             ScheduledExecutorService delegate,
@@ -217,6 +228,17 @@ public final class MetricRegistries {
                 checkNotNull(name, "name"));
     }
 
+    /**
+     * Returns an instrumented {@link ExecutorService} that monitors the number of tasks submitted, running,
+     * completed and also keeps a {@link com.codahale.metrics.Timer} for the task duration. Similar to
+     * {@link com.codahale.metrics.InstrumentedExecutorService}, but produces tagged metrics to the specified
+     * {@link TaggedMetricRegistry}.
+     *
+     * @param registry tagged metric registry
+     * @param delegate executor service to instrument
+     * @param name executor service name
+     * @return instrumented executor service
+     */
     public static ExecutorService instrument(
             TaggedMetricRegistry registry,
             ExecutorService delegate,
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsExecutorService.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsExecutorService.java
@@ -14,9 +14,7 @@
  * limitations under the License.
  */
 
-package com.palantir.tritium.metrics.executor;
-
-import static com.palantir.logsafe.Preconditions.checkNotNull;
+package com.palantir.tritium.metrics;
 
 import com.codahale.metrics.Counter;
 import com.codahale.metrics.Meter;
@@ -34,7 +32,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
-public final class TaggedMetricsExecutorService implements ExecutorService {
+final class TaggedMetricsExecutorService implements ExecutorService {
 
     private final ExecutorService delegate;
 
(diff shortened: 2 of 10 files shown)


schlosna (approved):

tritium-executor/src/main/java/com/palantir/tritium/metrics/executor/TaggedMetricsExecutorService.java
```diff
@@ -0,0 +1,199 @@
...
+
+    private final ExecutorService delegate;
+
+    private final Meter submitted;
+    private final Counter running;
+    private final Meter completed;
+    private final Timer duration;
+    private final Timer queuedDuration;
+
+    public TaggedMetricsExecutorService(
```
carterkozak: Thoughts on making the executor implementation private and using factory methods instead?

tritium-executor/src/main/java/com/palantir/tritium/metrics/executor/TaggedMetricsExecutorService.java
```diff
@@ -0,0 +1,199 @@
...
+    public boolean isTerminated() {
+        return delegate.isTerminated();
+    }
+
+    @Override
+    public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
+        return delegate.awaitTermination(timeout, unit);
+    }
+
+    private class TaggedMetricsRunnable implements Runnable {
```
carterkozak: ~~static+final~~
disregard

tritium-metrics/src/main/java/com/palantir/tritium/metrics/executor/TaggedMetricsExecutorService.java
```diff
@@ -0,0 +1,199 @@
...
+            } finally {
+                running.dec();
+                completed.mark();
+            }
+        }
+    }
+
+    private static MetricName createMetricName(String metricName, Map<String, String> safeTags) {
+        return MetricName.builder()
+                .safeName(MetricRegistry.name("com/palantir/tritium/metrics/executor", metricName))
```
schlosna: probably don't want the long slashed name

```suggestion
                .safeName(MetricRegistry.name("executor", metricName))
```

tritium-metrics/src/main/java/com/palantir/tritium/metrics/executor/TaggedMetricsScheduledExecutorService.java
```diff
@@ -0,0 +1,255 @@
...
+            } finally {
+                running.dec();
+                completed.mark();
+            }
+        }
+    }
+
+    private static MetricName createMetricName(String metricName, Map<String, String> safeTags) {
+        return MetricName.builder()
+                .safeName(MetricRegistry.name("com/palantir/tritium/metrics/executor", metricName))
```
schlosna: probably don't want the long slashed name
```suggestion
                .safeName(MetricRegistry.name("scheduled-executor", metricName))
```

tritium-metrics/src/main/java/com/palantir/tritium/metrics/executor/TaggedMetricsExecutorService.java
```diff
@@ -0,0 +1,199 @@
...
+    private final Counter running;
+    private final Meter completed;
+    private final Timer duration;
+    private final Timer queuedDuration;
+
+    public TaggedMetricsExecutorService(
+            ExecutorService delegate,
+            TaggedMetricRegistry registry,
+            Map<String, String> safeTags) {
+        this.delegate = delegate;
```
schlosna: probably want to check that safeTags is non-null & not-empty so that they have reasonable metric names
```suggestion
        this.delegate = checkNotNull(delegate, "delegate");
        checkArgument(!safeTags.isEmpty(), "safeTags cannot be empty");
```

tritium-metrics/src/main/java/com/palantir/tritium/metrics/executor/TaggedMetricsScheduledExecutorService.java
```diff
@@ -0,0 +1,255 @@
...
+    private final Meter scheduledOnce;
+    private final Meter scheduledRepetitively;
+    private final Counter scheduledOverrun;
+    private final Histogram scheduledPercentOfPeriod;
+
+    public TaggedMetricsScheduledExecutorService(
+            ScheduledExecutorService delegate,
+            TaggedMetricRegistry registry,
+            Map<String, String> safeTags) {
+        this.delegate = delegate;
```
schlosna: probably want to check that safeTags is non-null & not-empty so that they have reasonable metric names
```suggestion
        this.delegate = checkNotNull(delegate, "delegate");
        checkArgument(!safeTags.isEmpty(), "safeTags cannot be empty");
```

tritium-metrics/src/test/java/com/palantir/tritium/metrics/executor/TaggedMetricsExecutorServiceTest.java
```diff
@@ -0,0 +1,97 @@
...
+        assertThat(registry.meter(SUBMITTED).getCount()).isEqualTo(1);
+        assertThat(registry.counter(RUNNING).getCount()).isEqualTo(0);
+        assertThat(registry.meter(COMPLETED).getCount()).isEqualTo(1);
+        assertThat(registry.timer(DURATION).getCount()).isEqualTo(1);
+        assertThat(registry.timer(QUEUED_DURATION).getCount()).isEqualTo(1);
+    }
+
+    private static MetricName metricName(String metricName) {
+        return MetricName.builder()
+                .safeName(MetricRegistry.name("com/palantir/tritium/metrics/executor", metricName))
```
schlosna:
```suggestion
                .safeName(MetricRegistry.name("executor", metricName))
```

tritium-metrics/src/test/java/com/palantir/tritium/metrics/executor/TaggedMetricsScheduledExecutorServiceTest.java
```diff
@@ -0,0 +1,152 @@
...
+        finishSemaphore.release();
+        startSemaphore.acquire();
+
+        assertThat(registry.counter(SCHEDULED_OVERRAN).getCount()).isEqualTo(1);
+        assertThat(registry.histogram(SCHEDULED_PERCENT_OF_PERIOD).getCount()).isEqualTo(2);
+    }
+
+    private static MetricName metricName(String metricName) {
+        return MetricName.builder()
+                .safeName(MetricRegistry.name("com/palantir/tritium/metrics/executor", metricName))
```
schlosna: should make the name distinct for scheduled versus regular executor

```suggestion
                .safeName(MetricRegistry.name("scheduled-executor", metricName))
```

tritium-metrics/src/main/java/com/palantir/tritium/metrics/executor/TaggedMetricsExecutorService.java
```diff
@@ -0,0 +1,199 @@
...
+
+    private final ExecutorService delegate;
+
+    private final Meter submitted;
+    private final Counter running;
+    private final Meter completed;
+    private final Timer duration;
+    private final Timer queuedDuration;
+
+    public TaggedMetricsExecutorService(
```
schlosna: To @carterkozak's comment about factory method, I think we need to validate that the `safeTags` are not empty so that folks don't end up with difficult to understand or mistakenly overwritten metrics, and we could have an additional factory method to make the common case easy:


```suggestion
    public static ExecutorService wrap(
                ExecutorService delegate,
                TaggedMetricRegistry registry,
                String name) {
        checkArgument(!name.isEmpty(), "name cannot be empty");
        return wrap(delegate, registry, Collections.singletonMap("name", name));
    }

    public static ExecutorService wrap(
                ExecutorService delegate,
                TaggedMetricRegistry registry,
                Map<String, String> safeTags) {
        return new TaggedMetricsExecutorService(delegate, registry, safeTags);
    }

    private TaggedMetricsExecutorService(
```

tritium-metrics/src/main/java/com/palantir/tritium/metrics/executor/TaggedMetricsScheduledExecutorService.java
```diff
@@ -0,0 +1,255 @@
...
+    private final Counter running;
+    private final Meter completed;
+    private final Timer duration;
+
+    private final Meter scheduledOnce;
+    private final Meter scheduledRepetitively;
+    private final Counter scheduledOverrun;
+    private final Histogram scheduledPercentOfPeriod;
+
+    public TaggedMetricsScheduledExecutorService(
```
schlosna: similar to other constructor/factory method

```suggestion
    public static ScheduledExecutorService wrap(
                ExecutorService delegate,
                TaggedMetricRegistry registry,
                String name) {
        checkArgument(!name.isEmpty(), "name cannot be empty");
        return wrap(delegate, registry, Collections.singletonMap("name", name));
    }

    public static ScheduledExecutorService wrap(
                ExecutorService delegate,
                TaggedMetricRegistry registry,
                Map<String, String> safeTags) {
        return new TaggedMetricsScheduledExecutorService(delegate, registry, safeTags);
    }

    private TaggedMetricsScheduledExecutorService(
```

schlosna (changes requested):
looks like force push overwrote some things in a weird way, have some suggestions below

tritium-metrics/src/main/java/com/palantir/tritium/metrics/executor/TaggedMetricsExecutorService.java
carterkozak: I really like the model in `tritium-caffeine` with a single utility class of public functinoality, and the instrumentation implementations live in separate package-private classes.

Since ScheduledExecutorService extends ExecutorService, we may want to provide an overloaded factory method, so the compiler picks the most accurate one.

carterkozak:
I'm curious when we need to provide custom safe tags, I would hope that the executor name is sufficient for most uses. Should we leave out custom tags for now, and add them later if we find a scenario that requires them?

pkoenig10:
Witchcraft wants to create executors with the following tags:
- `name`: it didn't seem right to include the concept of executor names here since it's not necessary
- `type`: this wrapper doesn't know whether an executor is fixed or cached so users would have to pass that as an arguments. At that point it just seemed better to err on the side of flexibility.

carterkozak:
I think name makes sense here, using the cache metrics as an example, we provide a name to each cache despite not being necessary for the functionality of the cache.

type: That's fair. but we have enough information to differentiate between scheduled and non-scheduled, can we tag with those for now instead?

schlosna:
I don’t have super strong opinions on whether one would want to tag on type to be able to get specific aggregates. I’m open to allowing folks to specify their additional tags if they want, but we probably want to enforce some consistency to at least include a `name` tag.

Might be a separate add on later depending on intended usage, but might want to consider some additional metrics if wrapping the raw concrete thread pool types.

* For a `ThreadPoolExecutor` fixed pool, typically want to also capture gauge for the queue depth, and for cached pool, would want gauge for the total thread count and possibly ratio of total thread count to core pool size.

* For `ForkJoinPool` would want similar queued task count, pool size, active thread count.

tritium-metrics/src/test/java/com/palantir/tritium/metrics/executor/TaggedMetricsScheduledExecutorServiceTest.java
pkoenig10: I don't think we want this. The motivation to use common metric names with tags is to have a single metric to measure all executors. Splitting up the metric names again feels like a step backwards.

pkoenig10:
Decided to go with just a `name` tag. Arbitrary tags seems like a footgun. The `type` tag does not seems that useful since you can determine the type from the executor name.

carterkozak:
Thoughts on these changes? #233

schlosna (commented):
Quick scan on mobile looks good, want to poke on a real screen before merging

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
```diff
@@ -204,6 +207,29 @@ static void registerCache(MetricRegistry registry, Cache<?, ?> cache, String nam
                 .forEach((key, value) -> registerWithReplacement(registry, key, value));
     }
 
+    public static ScheduledExecutorService instrument(
```
carterkozak, line 221: We should javadoc these. It would be helpful to document all the expected metric names + types as a reference for dashboard authors.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsExecutorService.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+final class TaggedMetricsExecutorService implements ExecutorService {
```
carterkozak, line 35: If we had a [WrappingExecutorService](https://github.com/palantir/tracing-java/blob/develop/tracing/src/main/java/com/palantir/tracing/WrappingExecutorService.java) available, we could simplify these to call `submitted.mark()` in the callable ctor, and avoid some complexity in each method.
This would result in n invocations of mark(1) rather than mark(n), but I can't imagine that makes a great deal of difference.

Unfortunately we don't have access to that sort of executor, and I'm not sure it's worth defining another two copies in tritium just for these executor implementations.

blab: Ignore this comment, recording for posterity

carterkozak:
I’m happy with this. Docs would be helpful :-)

schlosna (approved):

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
schlosna: Added JavaDoc to the public methods

PR #233, opened 2019-04-11
factory methods in MetricRegristries.java

carterkozak:
ExecutorService implementations are completely internal

### DIFF
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
@@ -30,6 +30,7 @@ import com.google.common.cache.Cache;
 import com.google.common.collect.ImmutableSet;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
@@ -37,6 +38,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.Supplier;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
```diff
@@ -206,6 +209,29 @@ static void registerCache(MetricRegistry registry, Cache<?, ?> cache, String nam
                 .forEach((key, value) -> registerWithReplacement(registry, key, value));
     }
 
+    public static ScheduledExecutorService instrument(
+            TaggedMetricRegistry registry,
+            ScheduledExecutorService delegate,
+            String name) {
```
carterkozak, line 215: Updated argument order to match others in this class: registry, delegate, name

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
```diff
@@ -206,6 +209,29 @@ static void registerCache(MetricRegistry registry, Cache<?, ?> cache, String nam
...
+        return new TaggedMetricsScheduledExecutorService(
+                checkNotNull(delegate, "delegate"),
+                checkNotNull(registry, "registry"),
+                checkNotNull(name, "name"));
```
carterkozak, line 219: Previous implementation checked registry was not null twice, by using the results from checkNotNull we can leverage the compiler for validation.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
```diff
@@ -206,6 +209,29 @@ static void registerCache(MetricRegistry registry, Cache<?, ?> cache, String nam
                 .forEach((key, value) -> registerWithReplacement(registry, key, value));
     }
 
+    public static ScheduledExecutorService instrument(
```
carterkozak, line 212: These return ExecutorService/ScheduledExecutorService so that we can add additional implementations (may make ThreadPoolExecutor specific metrics cleaner to add in the future).

The overload allows us to route invocations to the most correct method based on the executor type at compile time.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
```diff
@@ -206,6 +209,29 @@ static void registerCache(MetricRegistry registry, Cache<?, ?> cache, String nam
...
+                checkNotNull(registry, "registry"),
+                checkNotNull(name, "name"));
+    }
+
+    public static ExecutorService instrument(
+            TaggedMetricRegistry registry,
+            ExecutorService delegate,
+            String name) {
+        if (delegate instanceof ScheduledExecutorService) {
+            return instrument(registry, (ScheduledExecutorService) delegate, name);
```
carterkozak, line 227: This may be overkill, but it shouldn't ever cause us to lose information.

carterkozak:
factory method name:
I chose `instrument` because it's functionally similar to `Tritium.instrument`, but I'm happy to use other names. `register` doesn't quite work because we're not scraping values from an existing instance, the returned instrumented service must be used to take measurements.

factory method location (MetricRegistries):
MetricRegistries already provides similar functionality, I think this is the most discoverable place to use, though a new utility class would also be reasonable.

pkoenig10 (approved):
This all LGTM

schlosna (approved):

PR #245, opened 2019-04-23
Avoid potential deadlock copying system properties

schlosna:
Since system properties are backed by a `java.util.Hashtable`, they can be a point of contention as all access is synchronized. We therefore take an approach of cloning the entire `Hashtable` (which does its own locking), then copying only the entries we are interested in keeping.

See https://bugs.openjdk.java.net/browse/JDK-6977738 and https://bugs.openjdk.java.net/browse/JDK-8029891

### DIFF
--- a/tritium-core/src/main/java/com/palantir/tritium/event/InstrumentationProperties.java
+++ b/tritium-core/src/main/java/com/palantir/tritium/event/InstrumentationProperties.java
@@ -24,7 +24,6 @@ import com.google.common.collect.ImmutableMap;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.tritium.api.functions.BooleanSupplier;
 import java.util.Map;
-import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


tritium-core/src/main/java/com/palantir/tritium/event/InstrumentationProperties.java
```diff
@@ -80,28 +79,23 @@ public static void reload() {
...
          */
-        Properties properties = System.getProperties();
-        //noinspection SynchronizationOnLocalVariableOrMethodParameter
-        synchronized (properties) {
-            return ImmutableMap.copyOf(properties);
-        }
+        @SuppressWarnings("unchecked")
+        Map<Object, Object> clonedSystemProperties = (Map<Object, Object>) System.getProperties().clone();
+        Map<String, String> map = clonedSystemProperties.entrySet().stream()
+                .filter(entry -> String.valueOf(entry.getKey()).startsWith(INSTRUMENT_PREFIX))
```
carterkozak: Should we filter out anything that isn't a string key or value instead, then cast?

tritium-core/src/test/java/com/palantir/tritium/event/InstrumentationPropertiesTest.java
```diff
@@ -18,13 +18,31 @@
...
+import javax.annotation.Nullable;
+import org.awaitility.Duration;
+import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
 public class InstrumentationPropertiesTest {
 
+    private ListeningExecutorService executorService =
+            MoreExecutors.listeningDecorator(Executors.newCachedThreadPool());
```
carterkozak: This needs to be created in `before()` if we're shutting it down in `after()`

tritium-core/src/test/java/com/palantir/tritium/event/InstrumentationPropertiesTest.java
```diff
@@ -82,4 +105,80 @@ public void invalid() {
...
+                            throwable.printStackTrace(System.err);
+                            assertThat(throwable).isNull();
+                        }
+                    }, MoreExecutors.directExecutor());
+                    return task;
+                })
+                .collect(Collectors.toList());
+
+        await().atMost(Duration.FIVE_SECONDS).untilAsserted(() -> {
+            assertThat(Futures.allAsList(futures).get()).hasSize(futures.size());
```
carterkozak: I think we want to assert that these have all succeeded, `allAsList` will include failed futures.

tritium-core/src/test/java/com/palantir/tritium/event/InstrumentationPropertiesTest.java
```diff
@@ -82,4 +105,80 @@ public void invalid() {
...
+                    latch.await();
+                    return "isSpecificEnabled: " + InstrumentationProperties.isSpecificEnabled("test");
+                },
+                () -> {
+                    latch.countDown();
+                    latch.await();
+                    return "isGloballyEnabled: " + InstrumentationProperties.isGloballyEnabled();
+                });
+
+        List<ListenableFuture<Object>> futures = tasks.stream()
```
carterkozak: nit: `executor.invokeAll(tasks)`

tritium-core/src/main/java/com/palantir/tritium/event/InstrumentationProperties.java
schlosna: I'll add it, though since [`java.util.Properties` is a `Map<Object, Object>`](https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html) that masquerades as a `Map<String, String>` and while there are ways to inject non-string keys/values, this is a relatively safe assumption here

tritium-core/src/test/java/com/palantir/tritium/event/InstrumentationPropertiesTest.java
schlosna: [`ListeningExecutorService#invokeAll` unfortunately returns a `List<Future<T>>`](https://google.github.io/guava/releases/23.6-jre/api/docs/com/google/common/util/concurrent/ListeningExecutorService.html#invokeAll-java.util.Collection-) so requires an explicit cast to raw `List`:

```
        @SuppressWarnings("unchecked") // guaranteed by ListenableExecutorService
        List<ListenableFuture<Object>> futures = (List) executorService.invokeAll(tasks);
```

I switched over to that and adjusted the checks a bit to ensure all tasks complete successfully

carterkozak:
👍 lgtm, thanks!

carterkozak (approved):

PR #248, opened 2019-04-24
Allow access to wrapped SSLEngine delegate instances

carterkozak:
This is necessary for servers which provide Conscrypt-based ALPN
using Conscryp on java 8:
https://github.com/google/conscrypt/blob/d7d3ec873a50c733c9c4397d6718c04045b377b3/common/src/main/java/org/conscrypt/Conscrypt.java#L638-L649

### DIFF
--- a/tritium-metrics/src/test/java/com/palantir/tritium/metrics/InstrumentedSslContextTest.java
+++ b/tritium-metrics/src/test/java/com/palantir/tritium/metrics/InstrumentedSslContextTest.java
@@ -36,6 +36,7 @@ import java.util.Collections;
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLSocketFactory;
 import javax.net.ssl.TrustManager;
 import javax.net.ssl.TrustManagerFactory;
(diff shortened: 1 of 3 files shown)


tritium-metrics/src/test/java/com/palantir/tritium/metrics/InstrumentedSslContextTest.java
```diff
@@ -176,6 +177,23 @@ public void testServerInstrumentation() throws Exception {
...
+
+    @Test
+    public void testSslEngineUnwrapInstrumentedTwice() throws IOException, GeneralSecurityException {
+        SSLContext context = newServerContext();
+        TaggedMetricRegistry metrics = new DefaultTaggedMetricRegistry();
+        SSLContext wrapped = MetricRegistries.instrument(
+                metrics, MetricRegistries.instrument(metrics, context, "first"), "second");
+        SSLEngine engine = wrapped.createSSLEngine();
+        assertThat(engine).isInstanceOf(InstrumentedSslEngine.class);
+        assertThat(MetricRegistries.unwrap(engine)).isNotInstanceOf(InstrumentedSslEngine.class);
```
schlosna: Probably worth adding `.isNotNull()` as well just to be sure (though that should be impossible)

```suggestion
        assertThat(MetricRegistries.unwrap(engine)).isNotNull().isNotInstanceOf(InstrumentedSslEngine.class);
```

carterkozak: Good call. Updated `MetricRegistries.unwrap(SSLEngine)` with a null check as well.

schlosna (approved):

PR #249, opened 2019-04-25
[break] Only one 'default' TaggedMetricRegistry

iamdanfox:
In the old world, libraries called `DefaultTaggedMetricRegistry.getDefault()` and everything was kinda OK.

However, we recently added the `SlidingWindowTaggedMetricRegistry`, which also has a `getDefault()` method.  I think this method is dangerous because there's now a chance some libraries will write to the first and other libraries write to the second. In WC we just emit metrics from _one_ registry, so this could lead to some metrics not being emitted!

- [atlasdb](https://github.com/palantir/atlasdb/blob/3b3031b86c972baacccc03d08ed9c40f8ec678a4/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java#L241)
- [conjure-java-runtime](https://github.com/palantir/conjure-java-runtime/blob/develop/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java#L94)

Obviously it's preferable for libraries to just ask users to provide a specific TaggedMetricRegistry, as this is much easier to test and swap in/out implementations! In the meantime, my proposal is to just give people one default to rule them all.


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/SharedTaggedMetricRegistries.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.metrics.registry;
+
+import com.codahale.metrics.SharedMetricRegistries;
+import java.util.concurrent.TimeUnit;
+
+/** Analogous to codahale's {@link SharedMetricRegistries}. */
```
iamdanfox, line 22: note, the codahale one has significantly more features - but I'm not sure we really need them or want them? https://github.com/dropwizard/metrics/blob/v3.2.6/metrics-core/src/main/java/com/codahale/metrics/SharedMetricRegistries.java

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/SlidingWindowTaggedMetricRegistry.java
```diff
@@ -21,18 +21,18 @@
...
 
     /**
      * Get the global default {@link TaggedMetricRegistry}.
+     * @deprecated inline this
      */
     @SuppressWarnings("unused") // public API
+    @Deprecated
     public static TaggedMetricRegistry getDefault() {
-        return SlidingWindowTaggedMetricRegistry.DEFAULT;
+        return SharedTaggedMetricRegistries.getDefault();
```
schlosna: I'm somewhat inclined to just delete this method completely and cut a `0.12.0` breaking release as I don't believe anyone actually uses this.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
```diff
@@ -22,18 +22,18 @@
...
 
-    private static final TaggedMetricRegistry DEFAULT = new DefaultTaggedMetricRegistry();
-
     public DefaultTaggedMetricRegistry() {
         super(ExponentiallyDecayingReservoir::new);
     }
 
     /**
      * Get the global default {@link TaggedMetricRegistry}.
+     * @deprecated inline this method
```
schlosna: nit: link the replacement
```suggestion
     * @deprecated use SharedTaggedMetricRegistries#getDefault
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/SharedTaggedMetricRegistries.java
schlosna: I'm generally supportive of the simple singleton as it is often difficult to inject/plumb through to all the places we actually want to instrument. I don't think we need the complexity of the separately named registries as we implicitly assume the single shared registry for our services.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/SlidingWindowTaggedMetricRegistry.java
```diff
@@ -21,18 +21,8 @@
...
-
     public SlidingWindowTaggedMetricRegistry(int window, TimeUnit windowUnit) {
         super(() -> new SlidingTimeWindowArrayReservoir(window, windowUnit));
     }
 
-    /**
-     * Get the global default {@link TaggedMetricRegistry}.
-     */
-    @SuppressWarnings("unused") // public API
-    public static TaggedMetricRegistry getDefault() {
```
iamdanfox, line 34: Deleting this is a compile break, but the method was only introduced in tritium 0.11.10 (released two days ago), so if we assume people haven't started compiling against this in the last two days then this should be fine...

schlosna: I would also consider the switch for default tagged registry from exponential decay to sliding window makes this a `break` as well, so will make this `0.12.0`

schlosna (approved):

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/SharedTaggedMetricRegistries.java
```diff
@@ -0,0 +1,33 @@
...
+public final class SharedTaggedMetricRegistries {
+
+    private static final TaggedMetricRegistry DEFAULT = new SlidingWindowTaggedMetricRegistry(1, TimeUnit.MINUTES);
```
iamdanfox: Before we release 0.12.0, let's double check whether 35 seconds was actually a better number for some reason...

kasper0406: 35 seconds is due to internal projects reporting metrics every 30 seconds. The though was to make the window as small as possible and still being confident we don't drop events - hence the 5 seconds margin.

iamdanfox: Ok so you're saying the danger is if we dump the metric registry on a slower frequency than its window then we lose information, but larger windows make each dump slightly more confusing (because it still contains data from a previous sample). 

Maybe we should just set the default to 35 seconds then and write some javadoc to this effect?

kasper0406: Yes, if the window is smaller than our report frequency, we will lose information. For larger windows, I guess it won't actually be confusing, as the metrics will simply be aggregated over that window. It just seems desirable to have as granular measurements as possible, to be able to figure out when something changed.

felixdesouza:
re atlas, the in memory transaction manager is for tests and should definitely not be used in prodc, so it's not a real blocker.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/SharedTaggedMetricRegistries.java
felixdesouza: Adding some colour here since I introduced the 35 seconds aha. When we were doing our local benchmarks for timelock we were looking at extreme values i.e. max and p999 for a particularly small window i.e. 6-7 seconds, and having the window be at 60 seconds or even higher just threw our results off entirely for the entire window.

In theory it's probably fine having it be 30 seconds, I was just being conservative. Plus outside of repeated benchmarks, it's likely to not actually matter.

As an aside, do we necessarily want to encode WC metrics reporting frequency inside tritium? Granted it's probably difficult due to cjr but just curious.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
```diff
@@ -22,18 +22,18 @@
 @AutoService(TaggedMetricRegistry.class)
 public final class DefaultTaggedMetricRegistry extends AbstractTaggedMetricRegistry {
 
-    private static final TaggedMetricRegistry DEFAULT = new DefaultTaggedMetricRegistry();
-
     public DefaultTaggedMetricRegistry() {
         super(ExponentiallyDecayingReservoir::new);
```
felixdesouza, line 26: we should probably change this too, feels weird to have a no-args constructor and a no-args "default" static factory that return two different behaviours

tritium-registry/src/test/java/com/palantir/tritium/metrics/registry/SharedTaggedMetricRegistriesTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.junit.Test;
+
+public class SharedTaggedMetricRegistriesTest {
+    @Test
+    public void all_default_methods_return_the_same_thing() {
+        TaggedMetricRegistry deprecated = DefaultTaggedMetricRegistry.getDefault();
+        TaggedMetricRegistry defaultRegistry = SharedTaggedMetricRegistries.getDefault();
+
+        assertThat(deprecated, is(defaultRegistry));
```
felixdesouza, line 31: surprised you're using hamcrest

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
iamdanfox: I actually don't want to change this, because the hierarchy is:

```
TaggedMetricRegistry
├── DefaultTaggedMetricRegistry
└── SlidingWindowTaggedMetricRegistry
```

In `DefaultMetricRegistry.getDefault()`, the 'default' means different things in both cases - the first one could have been 'ExponentiallyDecaying' or 'Impl'. I think the `getDefault()` method could really have been more of a `getSingleton()`.

Basically if we change this constructor then there is no way to get an ExponentiallyDecayingReservoir at all.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/SharedTaggedMetricRegistries.java
```diff
@@ -0,0 +1,33 @@
...
+
+    /** Singleton, for use when it is infeasible to plumb through a user supplied instance. */
+    public static TaggedMetricRegistry getDefault() {
```
iamdanfox: @felixdesouza would you prefer this to be called `getSingleton()`?

felixdesouza: yeah, makes a lot more sense tbh

iamdanfox:
@schlosna ran the JMH benchmarks locally:

```
Benchmark                                                                  Mode  Cnt     Score      Error   Units
ProxyBenchmark.withMetricsHdr                                              avgt    5   363.385 ±   33.315   ns/op
ProxyBenchmark.withMetricsHdr:·gc.alloc.rate.norm                          avgt    5   816.000 ±    0.001    B/op
ProxyBenchmark.withTaggedMetricsDecaying                                   avgt    5   735.931 ±   66.119   ns/op
ProxyBenchmark.withTaggedMetricsDecaying:·gc.alloc.rate.norm               avgt    5   728.010 ±    0.005    B/op
ProxyBenchmark.withTaggedMetricsHdr                                        avgt    5   599.791 ±   28.426   ns/op
ProxyBenchmark.withTaggedMetricsHdr:·gc.alloc.rate.norm                    avgt    5   728.000 ±    0.001    B/op
ProxyBenchmark.withTaggedMetricsSliding                                    avgt    5   762.509 ±  418.917   ns/op
ProxyBenchmark.withTaggedMetricsSliding:·gc.alloc.rate.norm                avgt    5   744.172 ±    0.011    B/op
```

We were using the exponentially decaying reservoir before, so I think the 735ns/op -> 762ns/op is actually quite reasonable here! Reassuring that allocations are actually pretty much the same too!

gmaretic:
When can we expect this to merge and release?

PR #252, opened 2019-04-29
Change executor name tag

pkoenig10:
This tag conflicts with another tag used internally to identify hosts. It's probably better to use something more specific here anyways.


tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsScheduledExecutorService.java
```diff
@@ -248,7 +248,7 @@ public T call() throws Exception {
     private static MetricName createMetricName(String metricName, String name) {
         return MetricName.builder()
                 .safeName(MetricRegistry.name("executor", metricName))
-                .putSafeTags("name", name)
+                .putSafeTags("executor-name", name)
```
carterkozak: How about "executor"?

I think I used `"name"` for the SSLContext metrics as well, would you mind updating that to `context` in this PR as well?

carterkozak:
Thanks! 👍

carterkozak (approved):

schlosna (approved):

schlosna:
@pkoenig10 you can add the `merge when ready` when you’re ready to merge or let me know

PR #254, opened 2019-05-06
Enforce additional error-prone checks

schlosna:


build.gradle
```diff
@@ -70,24 +70,98 @@ allprojects {
...
-                '-Xep:PreferSafeLoggableExceptions:ERROR',
-                '-Xep:PreferSafeLoggingPreconditions:ERROR',
-                '-Xep:ReturnMissingNullable:ERROR',
-                '-Xep:Slf4jLogsafeArgs:ERROR',
-                '-Xep:SwitchStatementDefaultCase:ERROR',
-                '-Xep:UnnecessaryParentheses:ERROR',
-                '-Xep:UnusedVariable:ERROR',
             ]
+
+            def warningsToPromote = [
```
carterkozak: Should we use the list from errorprone?
https://github.com/google/error-prone/blob/3d225c1eec6d8dada11730c9e9199b710ea26495/core/src/main/java/com/google/errorprone/scanner/BuiltInCheckerSuppliers.java#L593

build.gradle
```diff
@@ -70,24 +70,98 @@ allprojects {
...
+
+            if (generateErrorPronePatch) {
+                def patchChecks = [
+                    "PreferSafeLoggableExceptions",
+                    "PreferCollectionTransform",
+                    "PreferListsPartition",
+                ]
+
+                logger.info("Generating error-prone.patch for checks {} apply with:\n"
+                        + "    find . -maxdepth 3 -type f -name 'error-prone.patch' \\ \n"
```
carterkozak, line 109: I think `-maxdepth 2` is sufficient here, though it doesn't make much difference. Some day I'll get around to writing a proper gradle task for these.

carterkozak (approved):

build.gradle
schlosna: Possibly, will give it a shot adding all non-default enabled. Last time I looked at the non-default ones there were some pathologically slow ones.

build.gradle
schlosna: Yeah, maxdepth 2 works for tritium.

Adding a configurable set of checks and option to generate patch to gradle-Baseline would be useful. Might give that a shot if I can find some time.

uschi2000:
Why not discuss this kind of stuff in the context of Baseline? Then everybody would benefit.

schlosna:
Hey @uschi2000 , I will bring this over to baseline. This PR was just some local changes I had from testing out and applying the checks added in https://github.com/palantir/gradle-baseline/pull/622 and https://github.com/palantir/gradle-baseline/pull/623 against Tritium.

uschi2000:
Cheers!

build.gradle
schlosna: done, promoted all enabled warnings & disabled checks to treat as errors

PR #258, opened 2019-05-09
Lower logging level in MetricRegistries

tom-s-powell:
The current `INFO` level seems too high when notifying about existing registered metrics being removed and `DEBUG` feels more appropriate.


carterkozak:
I'm not sure I agree with this, I want to know if metrics I have registered are being removed from the application. Would you mind providing a bit more information describing the usage pattern where this becomes noisy?

tom-s-powell:
I tagged you on a relevant PR @carterkozak

carterkozak:
@tom-s-powell Any reason not to close this out?

PR #260, opened 2019-05-10
[fix] MetricRegistries.createWithSlidingTimeWindowReservoirs() counts are monotonic after window

iamdanfox:
## Before this PR

We started using the SlidingTimeWindowArrayReservoir in WC release candidates, expecting to see slightly different p99s, but someone reported that counts were different!

Screenshot 2019-05-10 at 13 16 09

As you can see, the `.count` of a histogram was resetting to zero after the end of each time window, which is inconsistent with the tagged metric registry.

## After this PR

Histogram counts behave consistently whether they came from a tagged or untagged registry.

## Possible downsides
- I assume the ReservoirHistogram was intended as a performance optimization, so deleting it might cause a minor slowdown somewhere.  I consider this worth it because I'd prefer consistency!!

### DIFF
--- a/tritium-metrics/src/test/java/com/palantir/tritium/metrics/MetricRegistriesTest.java
+++ b/tritium-metrics/src/test/java/com/palantir/tritium/metrics/MetricRegistriesTest.java
@@ -31,14 +31,11 @@ import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Metric;
 import com.codahale.metrics.MetricFilter;
 import com.codahale.metrics.MetricRegistry;
-import com.codahale.metrics.SlidingTimeWindowArrayReservoir;
 import com.codahale.metrics.Snapshot;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheStats;
 import com.google.common.cache.LoadingCache;
-import com.palantir.tritium.metrics.registry.MetricName;
-import com.palantir.tritium.metrics.registry.SlidingWindowTaggedMetricRegistry;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.time.ZoneOffset;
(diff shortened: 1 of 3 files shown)


tritium-metrics/src/test/java/com/palantir/tritium/metrics/MetricRegistriesTest.java
```diff
@@ -394,6 +396,40 @@ public void problematicTimestamps() {
...
+        histogram.update(20);
+        histogram.update(20);
+        histogram.update(20);
+        histogram.update(20);
+
+        assertThat(histogram.getCount()).isEqualTo(4);
+
+        Thread.sleep(1);
+
+        assertThat(histogram.getCount()).isEqualTo(4); // wat
```
iamdanfox: This fails because this counter returns 0 but the tagged counter returns 4 🤯

iamdanfox: Hmmm seems like we knew about this originally @kasper0406 - was this intentional? https://github.com/palantir/tritium/pull/238/files#diff-52a48109605a97396b162e386757ef08R154

iamdanfox:
@schlosna so it seem like this disparity comes because on the untagged side, creating a histogram eventually makes one of these guys:

```java
    private static class ReservoirHistogram extends Histogram {
        private final Reservoir reservoir;

        ReservoirHistogram(Reservoir reservoir) {
            super(reservoir);
            this.reservoir = reservoir;
        }

        @Override
        public long getCount() {
            return reservoir.size(); // 🌶🌶🌶 I don't think this is a valid assumption... size != count
        }
    }
```

But on the tagged side, it just calls `new Histogram(createReservoir())`.

tritium-metrics/src/test/java/com/palantir/tritium/metrics/MetricRegistriesTest.java
kasper0406: Not intentional in the sense of any of my use cases. As noted below, this seems to be an implication of the sliding window, where the count is counting only the entries constituting the reported metric.
I do see that this could give unintended behavior if used with fx the `rate` function in Prometheus, as the count will now be sliding. Didn’t consider that. Seems better if it reported two counts - one being the current count, and one the total. Don’t know if that’s  possible though.

tritium-metrics/src/test/java/com/palantir/tritium/metrics/MetricRegistriesTest.java
```diff
@@ -394,6 +396,42 @@ public void problematicTimestamps() {
...
+        histogram.update(20);
+        histogram.update(20);
+        histogram.update(20);
+
+        assertThat(histogram.getCount()).isEqualTo(4);
+
+        Thread.sleep(1);
+
+        assertThat(histogram.getCount()).isEqualTo(4);
+    }
```
iamdanfox: happy to delete these Thread.sleep tests, just wanted to demonstrate that things were not consistent before and now they are.

schlosna: we might want a visible for test version of `SlidingWindowTaggedMetricRegistry` constructor that takes a `Clock` to simplify this kind of test

tritium-metrics/src/test/java/com/palantir/tritium/metrics/MetricRegistriesTest.java
```diff
@@ -394,6 +396,42 @@ public void problematicTimestamps() {
         assertParsesTimestamp("2019-03-29T23:38:51.092Z");
     }
 
+    @Test
+    public void codahale_registry_histogram_count_should_monotonically_increase_after_window()
+            throws InterruptedException {
+        MetricRegistry registry = new MetricRegistryWithReservoirs(
+                () -> new SlidingTimeWindowArrayReservoir(1, TimeUnit.MILLISECONDS));
```
schlosna: can we verify this with `HdrHistogram` reservoir as well?

schlosna (approved):
good catch, I honestly don't remember why we had size -> count.

tritium-metrics/src/test/java/com/palantir/tritium/metrics/MetricRegistriesTest.java
iamdanfox: Tbh I was going to just delete these two tests as I think the ones above did actually capture behaviour, they just captured the undesired behaviour!

Not sure what Thread.sleep we'd want on the HDR one - I thought it doens't rely on a clock?

PR #265, opened 2019-05-16
Revert "Excavator: Upgrade dependencies (#264)"

schlosna:
This reverts commit 96eb8712af519c2ff88caebfa0ede50562608604.

Jackson 2.9.9 has not been fully released yet per https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.9.9


schlosna:
when bumping downstream consumers, noticed that Jackson 2.9.9 versions of following modules were not yet available:

```
com.fasterxml.jackson.datatype:jackson-datatype-guava
com.fasterxml.jackson.jaxrs:jackson-jaxrs-base
com.fasterxml.jackson.jaxrs:jackson-jaxrs-cbor-provider
com.fasterxml.jackson.jaxrs:jackson-jaxrs-json-provider
com.fasterxml.jackson.jaxrs:jackson-jaxrs-smile-provider
```

versions.props
```diff
@@ -1,5 +1,5 @@
 ch.qos.logback:logback-classic = 1.1.2
-com.fasterxml.jackson.*:* = 2.9.9
```
iamdanfox, line 2: @schlosna we've actually stopped auto-upgrading some jars in tracing-java, to avoid imposing upgrades on people when they're not really necessary.

https://github.com/palantir/tracing-java/blob/develop/versions.props

I think we'd probably want applications (and probably witchcraft) to always be upgrading everything, but I think there might be an argument to have libraries bump stuff a bit less aggressively to avoid imposing a bleeding edge transitives on consumers who might not want them.

schlosna: Hey @iamdanfox, that makes sense. Excavator had bumped jackson for tritium this morning. 

I think we actually do want to take the Jackson 2.9.9 bump ASAP given the [Afterburner issues we've fixed upstream](https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.9.9#afterburner) in this release.

schlosna: Speaking of the bots, https://github.com/palantir/tritium/pull/266 interestingly bumped just the `versions.lock` for `com.fasterxml.jackson.datatype:jackson-datatype-guava` to `2.9.9`

schlosna:
Closing out as Jackson 2.9.9 has now been released per https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.9.9

PR #268, opened 2019-05-17
Add tagged cache metrics

schlosna:
For Guava caches:
`MetricRegistries#registerCache(TaggedMetricRegistry, Cache<?,?>, SafeArg<java.lang.String>)`

For Caffeine caches:
`CaffeineCacheStats#registerCache(TaggedMetricRegistry, Cache<?,?>, SafeArg<java.lang.String>)`

For each registered cache, produces tag name `cache` with tag value of specified `SafeArg`'s value, and all of the following metrics (assuming the relevant cache has `recordStats()` enabled)
```
"cache.estimated.size"
"cache.maximum.size"
"cache.weighted.size"
"cache.request.count"
"cache.hit.count"
"cache.hit.ratio"
"cache.miss.count"
"cache.miss.ratio"
"cache.eviction.count"
"cache.load.success.count"
"cache.load.failure.count"
"cache.load.average.millis"
```


tritium-metrics/src/main/java/com/palantir/tritium/metrics/CacheTaggedMetrics.java
```diff
@@ -0,0 +1,51 @@
...
+    private final String cacheName;
+
+    private CacheTaggedMetrics(Cache<?, ?> cache, SafeArg<String> arg) {
+        String name = checkNotNull(arg.getValue(), "cacheName").trim();
+        checkArgument(!name.isEmpty(), "Cache name cannot be blank or empty");
+        this.cache = checkNotNull(cache, "cache");
+        this.cacheName = name;
+    }
+
+    static CacheTaggedMetrics create(Cache<?, ?> cache, SafeArg<String> cacheName) {
```
carterkozak: Not a huge fan of `SafeArg` as a wrapper around a value since it requires a key that's never used. Perhaps we should use the existing annotation `@Safe String cacheName`?

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
```diff
@@ -37,19 +41,39 @@ private CaffeineCacheStats() {
...
+                .forEach((metricName, gauge) -> {
+                    registry.remove(metricName);
+                    registry.gauge(metricName, gauge);
+                });
     }
 
+    static <K extends Comparable<K>> Map<K, Gauge<?>> createCacheGauges(
+            Cache<?, ?> cache,
+            Function<String, K> metricNamer) {
+        return CacheMetrics.createMetrics(CaffeineStats.create(cache, 1, TimeUnit.SECONDS), metricNamer);
```
carterkozak: Was the change from 5 seconds to 1 second intentional?

carterkozak: Looks like some caches are moving from .5s -> 1s, and others from 5s -> 1s. Using 1 second across the board is easier for me to follow.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/CacheMetrics.java
```diff
@@ -0,0 +1,84 @@
...
+
+package com.palantir.tritium.metrics;
+
+import com.codahale.metrics.Gauge;
+import com.google.common.collect.ImmutableSortedMap;
+import com.palantir.tritium.metrics.registry.MetricName;
+import java.util.Map;
+import java.util.function.Function;
+
+public final class CacheMetrics {
```
carterkozak: Ideally this wouldn't be public, but I suppose it must be in order to share with caffeine? If that's the case, can we add some javadoc describing that this isn't meant to be consumed, and perhaps rename to something including `Internal` (not a strong preference, but I'd look for a class called CacheMetrics if I wanted to add metrics to a cache)

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
schlosna: yeah, with the switch to Guava memoized supplier with expiration, there's not a good way to inject a `clock` for testing, so the the 5 second caching was excessive and slows down tests, and allows for cases where we want 1 second frequency stats.

carterkozak (approved):
Looks great, I'm looking forward to using these!

tritium-metrics/src/main/java/com/palantir/tritium/metrics/CacheMetrics.java
schlosna: Renamed to `InternalCacheMetrics` and marked as `@Beta`

schlosna:
Going to give other folks a chance to review Monday before merging

tritium-metrics/src/main/java/com/palantir/tritium/metrics/CacheMetricSet.java
```diff
@@ -19,113 +19,108 @@
...
+            return stats.get();
+        }
+
+        @Override
+        public long estimatedSize() {
+            return cache.size();
+        }
+
+        @Override
+        public long weightedSize() {
```
mglazer: how can maximum and weighted size be -1 for a guava cache? It supports a maximum size. Is it just because the size is registered when it's built but not supplied on the interface?

mglazer (approved):

tritium-metrics/src/main/java/com/palantir/tritium/metrics/CacheMetricSet.java
schlosna: Yeah, right now there is not a good way to get max size/weight or weighted size. Could try a reflection based field hack.

schlosna: We could also make these couple optional and just not register them if they’re not available

schlosna: @carterkozak and others -- thoughts on removing `weightedSize` & `maximumSize` from all registered cache metrics? Currently we only produce them for Caffeine caches as Guava does not easily expose an equivalent, and not sure they're super useful anyway (see https://github.com/google/guava/issues/1035 ). 

If we really wanted, we could use something like the following fragile package access code (or reflection equivalent):
```
package com.google.common.cache;

import java.util.OptionalLong;

public final class GuavaCacheSpy {
    private GuavaCacheSpy() {}

    public static OptionalLong weightedSize(Cache<?, ?> cache) {
        if (cache instanceof LocalCache.LocalManualCache) {
            LocalCache localCache = ((LocalCache.LocalManualCache) cache).localCache;
            if (localCache != null) {
                long totalWeight = 0;
                for (LocalCache.Segment segment : localCache.segments) {
                    if (segment != null) {
                        totalWeight += segment.totalWeight;
                    }
                }
                return OptionalLong.of(totalWeight);
            }
        }
        return OptionalLong.empty();
    }

    public static OptionalLong maxWeight(Cache<?, ?> cache) {
        if (cache instanceof LocalCache.LocalManualCache) {
            LocalCache localCache = ((LocalCache.LocalManualCache) cache).localCache;
            if (localCache != null) {
                return OptionalLong.of(localCache.maxWeight);
            }
        }
        return OptionalLong.empty();
    }
}
```

carterkozak: I have used neither `weightedSize` nor `maximumSize`, though I can't speak for everyone.

My preference is to make them optional, and report values when they're available, that way we avoid bundling potential breaks with this feature.
I'd like to avoid the package hack and reflective access unless there's a very strong need, and even then I'd prefer migrating usage from guava to caffeine.

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineStats.java
```diff
@@ -0,0 +1,105 @@
...
+                .orElse(0L);
+    }
+
+    @Override
+    public long maximumSize() {
+        return cache.policy()
+                .eviction()
+                .map(Policy.Eviction::getMaximum)
+                .orElse(-1L);
+    }
```
schlosna, line 70: for context, these were originally added to untagged metrics via https://github.com/palantir/tritium/pull/93

tritium-metrics/src/main/java/com/palantir/tritium/metrics/CacheMetricSet.java
schlosna: Going to let these `return -1` for Guava caches for now, clarified JavaDoc on the `InternalCacheMetrics.Stats` interface, and move folks to over to Caffeine caches where they're still using Guava. See https://github.com/palantir/tritium/pull/93 for historical context where maximum & weighted size were added to untagged metrics.

carterkozak (approved):

PR #272, opened 2019-05-24
[break] Fix MetricName comparator for tags, MetricName#safeTags() returns SortedMap

schlosna:
Ensure `MetricName#safeTags` are compared in a total ordered fashion.


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/MetricName.java
```diff
@@ -65,7 +44,7 @@
```
schlosna: Actually should make this:

```
    @Value.NaturalOrder
    NavigableMap<String, String> safeTags();
```

carterkozak: That would do the trick, if we're okay with exposing `NavigatableMap` instead of `Map`

carterkozak: Why does MetricName need to be comparable? Are we better off removing that piece?

schlosna: I mainly added it for convenience for the `toString` form when analyzing `getMetrics` output

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/MetricNames.java
```diff
@@ -0,0 +1,57 @@
...
+
+final class MetricNames {
+
+    private MetricNames() {}
+
+    private static final Comparator<Map.Entry<String, String>> entryComparator =
+            Map.Entry.<String, String>comparingByKey()
+                    .thenComparing(Map.Entry.comparingByValue());
+
+    private static final Comparator<Map<String, String>> sizeComparator = Comparator.comparingInt(Map::size);
```
carterkozak: Nice touch

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/MetricName.java
```diff
@@ -61,11 +40,12 @@
      * <p>
      * All tags and keys must be {@link Safe} to log.
      */
-    Map<String, String> safeTags();
+    @Value.NaturalOrder
+    NavigableMap<String, String> safeTags();
```
carterkozak: Should we loosen this to SortedMap?

carterkozak (approved):

carterkozak (approved):

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/MetricName.java
schlosna: I'd prefer to just leave this as a `NavigableMap` as the underlying impl generated by immutables is a `TreeMap` either way with `jdkOnly = true` (though we could remove that since we already rely on Guava `implementation` dependency, created optional https://github.com/palantir/tritium/pull/273 for that).

schlosna (commented):
Introducing a slight break since 0.12.5 by removing `Comparable` from `MetricName`, but I don't think that was actually necessary and was just something I had added for my own convenience when implementing the tagged cache metrics.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/MetricName.java
carterkozak: I'd prefer the guava approach. We heavily rely on guava immutable collections, and they will avoid collection re-allocations when we attempt to re-wrap them.

I'm not sure any of the functionality provided by `NavigableMap` will be helpful for `safeTags`, where `SortedMap` provides clear advantages. Using the less specific type allows more freedom to change the implementation in the future. That said, I'm not actually aware of a `SortedMap` implementation which doesn't also implement `NavigableMap`, so I defer to your judgement.

schlosna: ok, you convinced me, went with `SortedMap` for now

carterkozak:
Thanks!

iamdanfox:
As mentioned by @JacekLach, automatically upgrading to this version of tritium has caused NoSuchMethodErrors at runtime! It seems like changing Map -> SortedMap was API compatible but not ABI compatible!

We had some code in sls-logging, which did this:

```java
        Map<String, String> tags = metricName.safeTags();

        if (metric instanceof Counter) {
            return Optional.of(MetricLogs.v1counter(name, (Counter) metric, tags));
        } else if (metric instanceof Meter) {
            ...
```

And it seems that first line triggers the exception:

```
java.lang.NoSuchMethodError: com.palantir.tritium.metrics.registry.MetricName.safeTags()Ljava/util/Map;
	at com.palantir.sls.logging.metric.ScheduledTaggedMetricLogReporter.metricLog(ScheduledTaggedMetricLogReporter.java:121)
	at com.palantir.sls.logging.metric.ScheduledTaggedMetricLogReporter.reportMetrics(ScheduledTaggedMetricLogReporter.java:102)
	at com.palantir.sls.logging.metric.ScheduledTaggedMetricLogReporter.lambda$report$2(ScheduledTaggedMetricLogReporter.java:91)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at com.palantir.tracing.DeferredTracer.withTrace(DeferredTracer.java:115)
	at com.palantir.tracing.Tracers$TracingAwareCallable.call(Tracers.java:343)
	at com.palantir.tracing.WrappingExecutorService.lambda$wrapTask$0(WrappingExecutorService.java:67)
	at com.codahale.metrics.InstrumentedExecutorService$InstrumentedRunnable.run(InstrumentedExecutorService.java:176)
	at com.palantir.tritium.metrics.TaggedMetricsExecutorService$TaggedMetricsRunnable.run(TaggedMetricsExecutorService.java:161)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.lang.Thread.run(Thread.java:834)
```

We worked around this by just releasing a new version of sls-logging.

schlosna:
@iamdanfox sorry about that ABI break. I'm open to reverting that change if necessary and tagging a 0.13.1 (I'm half inclined to just remove immutables generation for `MetricName` as well so that I can fully control the generated implementation).

We should probably consider an actual 1.0.0 so that we can effectively do real API/ABI breaks and mark them as proper semver major version bumps.

carterkozak:
I would support removing immutables since I've found a few places that subclass MetricName, I would prefer if we provided  a concrete final implementation instead. interface -> class would be another ABI break though.

schlosna:
I would probably leave `MetricName` as an interface, but provide an ABI compatible `Immutable` builder and simple static factory for just name & name+tags.

iamdanfox:
Given how easy it was to workaround this, not sure it's worth reverting just yet, just wanted to flag because this is such a low level library that even tiny breaks can end up having quite far reaching consequences!

JacekLach:
It was an easy workaround, but we only caught it in prod (when we noticed we were missing a subset of metrics for ~12h)! So there's a bit more to the tradeoff than just how difficult/easy the fix was.

PR #291, opened 2019-06-24
Upgrade tracing-java to take advantage of low-garbage fastStartSpan

carterkozak:
For context:
http://github.com/palantir/tracing-java/pull/180


tritium-tracing/src/main/java/com/palantir/tritium/tracing/JavaTracingTracer.java
```diff
@@ -21,7 +21,7 @@
 
     @Override
     public void startSpan(String operationName) {
-        com.palantir.tracing.Tracer.startSpan(operationName);
+        com.palantir.tracing.Tracer.fastStartSpan(operationName);
```
schlosna, line 24: We'll need to ensure that downstream consumers all pickup tracing-java 2.16.0+ at the same time as tritium bump. I'm on the fence about providing a temporary fallback:

```
    private volatile NoSuchMethodError failure = null;

    @Override
    public void startSpan(String operationName) {
        try {
            if (failure == null) {
                com.palantir.tracing.Tracer.fastStartSpan(operationName);
            } else {
                fallbackToSlowStartSpan(operationName, failure);
            }
        } catch (NoSuchMethodError e) {
            fallbackToSlowStartSpan(operationName, e);
        }
    }

    private void fallbackToSlowStartSpan(String operationName, NoSuchMethodError e) {
        // temporary bridge in case downstream consumer has
        com.palantir.tracing.Tracer.startSpan(operationName);
        if (failure == null) {
            failure = e;
            log.error("Falling back to slower Tracer.startSpan because tracing-java is not version 2.16.0+", e);
        }
    }
```

tritium-tracing/src/main/java/com/palantir/tritium/tracing/TracingInvocationEventHandler.java
```diff
@@ -63,7 +63,7 @@ public TracingInvocationEventHandler(String component) {
     public InvocationContext preInvocation(@Nonnull Object instance, @Nonnull Method method, @Nonnull Object[] args) {
         InvocationContext context = DefaultInvocationContext.of(instance, method, args);
         String operationName = getOperationName(method);
-        Tracer.startSpan(operationName);
+        Tracer.fastStartSpan(operationName);
```
schlosna, line 66: similar to above

tritium-tracing/src/main/java/com/palantir/tritium/tracing/JavaTracingTracer.java
carterkozak: Aside from one large internal project, I expect most consumers are using https://github.com/palantir/gradle-consistent-versions to ensure proper dependencies.

versions.lock
```diff
@@ -16,15 +16,15 @@ com.google.guava:listenablefuture:9999.0-empty-to-avoid-conflict-with-guava (1 c
...
+com.palantir.tracing:tracing:2.16.0 (1 constraints: 3b05413b)
+com.palantir.tracing:tracing-api:2.16.0 (1 constraints: d10c1116)
 io.dropwizard.metrics:metrics-core:3.2.5 (1 constraints: 4f0c80f3)
 org.checkerframework:checker-qual:2.6.0 (3 constraints: 201fc302)
 org.codehaus.mojo:animal-sniffer-annotations:1.17 (1 constraints: ed09d8aa)
 org.hdrhistogram:HdrHistogram:2.1.11 (1 constraints: 3705303b)
 org.immutables:value:2.7.5 (1 constraints: 10050f36)
 org.mpierce.metrics.reservoir:hdrhistogram-metrics-reservoir:1.1.2 (1 constraints: 0605f535)
-org.slf4j:slf4j-api:1.7.22 (5 constraints: 68389d4d)
+org.slf4j:slf4j-api:1.7.25 (5 constraints: 6f38d54e)
```
schlosna: Should bump this to 1.7.26 per https://nvd.nist.gov/vuln/detail/CVE-2018-8088

> February 16th, 2019 - Release of SLF4J 1.7.26
Due to popular demand in relation to CVE-2018-8088, EventData, EventException and EventLogger classes were removed from the org.slf4j.ext package located in the slf4j-ext module.
> 
> The aforementioned vulnerabilty has a low severty but was incorrectly marked critical in the National Vulnerability Database.

schlosna: handle this separately in #294 .

tritium-tracing/src/main/java/com/palantir/tritium/tracing/JavaTracingTracer.java
schlosna: There is at least one other major project not using GCV that we'll need to coordinate excavation.

wenhoujx: why don't they use it? can i add GCV to them ?

carterkozak: The larger of the two products is already on the upgraded tracing-java version, and the smaller does not depend directly on tracing, so it will pull the highest version from its dependencies (or fail resolution until a human does so).

I'm more worried about testing correctness of the fallback approach in tritium than upgrading products which attempt to run unsupported configurations. It's common for other libraries to bump dependency versions in order to take advantage of new features.

iamdanfox: @schlosna GCV should mean you no longer need to worry about consumers of forcing stuff to too low versions.  I think it'll work seamlessly without the fallback.

iamdanfox (approved):

tritium-tracing/src/main/java/com/palantir/tritium/tracing/JavaTracingTracer.java
schlosna: @iamdanfox yeah GCV "should" do the work, just want to make sure someone can ensure it all flows through and can shepherd things through so we don't see failures as I probably won't have cycles to do that for many of those projects

schlosna (approved):

PR #295, opened 2019-06-25
Replace usages of `assertThat(*.toString())` with `assertThat(*).asString()

carterkozak:
This provides cleaner null checks without writing the code ourselves.


schlosna (approved):
👍 nice, thanks for cleaning this up!

PR #308, opened 2019-07-07
[feature] Add new `tritium-metrics-jvm` module with standard JVM instrumentation

carterkozak:
==COMMIT_MSG==
[feature] Add new `tritium-metrics-jvm` module with standard JVM instrumentation
==COMMIT_MSG==


settings.gradle
```diff
@@ -7,6 +7,7 @@ include 'tritium-jmh'
 include 'tritium-lib'
 include 'tritium-proxy'
 include 'tritium-metrics'
+include 'tritium-metrics-jvm'
```
carterkozak, line 10: This module depends on dropwizard-metrics `metrics-jvm`, however that is relatively small and provides no additional dependencies.
Thoughts on the separate module vs adding the dependency on metrics-jvm to `tritium-metrics`?

dansanduleac: I don't mind either way, don't see a problem with separating things more like this, since in practice we'll only take a direct dependency on these from a couple of internal projects.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
```diff
@@ -441,4 +448,35 @@ public static SSLEngine unwrap(SSLEngine engine) {
...
+            } else if (metric instanceof Meter) {
+                registry.meter(metricName, () -> (Meter) metric);
+            } else if (metric instanceof Timer) {
+                registry.timer(metricName, () -> (Timer) metric);
+            } else {
+                throw new SafeIllegalArgumentException("Unknown Metric Type",
+                        SafeArg.of("type", metric.getClass()));
+            }
+        });
+    }
```
carterkozak, line 486: New function on the `MetricRegistries` utility class.

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -0,0 +1,65 @@
...
+                .put("jvm.attribute", new JvmAttributeGaugeSet())
+                .put("jvm.buffers", new BufferPoolMetricSet(ManagementFactory.getPlatformMBeanServer()))
+                .put("jvm.classloader", new ClassLoadingGaugeSet())
+                .put("jvm.memory", () -> Maps.filterKeys(
+                        // Memory pool metrics are already provided by MetricRegistries.registerMemoryPools
+                        new MemoryUsageGaugeSet().getMetrics(), name -> !name.startsWith("pools")))
+                .put("jvm.threads", new ThreadStatesGaugeSet())
+                .build();
+        dropwizardMetricSets.forEach((prefix, metric) -> MetricRegistries.registerAll(registry, prefix, metric));
+    }
```
carterkozak, line 58: This is the only public api in `tritium-metrics-jvm`

settings.gradle
ferozco: Agree with @dansanduleac. The separation seems fine

ferozco (approved):

schlosna (approved):

PR #309, opened 2019-07-07
Add metric for JVM process cpu load

carterkozak:
==COMMIT_MSG==
Added `process.cpu.utilization` gauge for process cpu load.
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/OperatingSystemMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/OperatingSystemMetrics.java
@@ -22,12 +22,19 @@ import java.lang.management.ManagementFactory;
 import java.lang.management.OperatingSystemMXBean;
 
 final class OperatingSystemMetrics {
+    private static final MetricName OS_LOAD_NORM_1 = MetricName.builder().safeName("os.load.norm.1").build();
+    private static final MetricName OS_LOAD_1 = MetricName.builder().safeName("os.load.1").build();
+    private static final MetricName OS_PROCESS_CPU_UTILIZATION
+            = MetricName.builder().safeName("process.cpu.utilization").build();
 
     static void register(TaggedMetricRegistry registry) {
         OperatingSystemMXBean osMxBean = ManagementFactory.getOperatingSystemMXBean();
-        registry.gauge(MetricName.builder().safeName("os.load.norm.1").build(),
-                () -> osMxBean.getSystemLoadAverage() / osMxBean.getAvailableProcessors());
-        registry.gauge(MetricName.builder().safeName("os.load.1").build(), osMxBean::getSystemLoadAverage);
+        registry.gauge(OS_LOAD_NORM_1, () -> osMxBean.getSystemLoadAverage() / osMxBean.getAvailableProcessors());
+        registry.gauge(OS_LOAD_1, osMxBean::getSystemLoadAverage);
+        if (osMxBean instanceof com.sun.management.OperatingSystemMXBean) {
+            com.sun.management.OperatingSystemMXBean sunBean = (com.sun.management.OperatingSystemMXBean) osMxBean;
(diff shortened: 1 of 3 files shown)


tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/OperatingSystemMetrics.java
```diff
@@ -28,6 +28,11 @@ static void register(TaggedMetricRegistry registry) {
         registry.gauge(MetricName.builder().safeName("os.load.norm.1").build(),
                 () -> osMxBean.getSystemLoadAverage() / osMxBean.getAvailableProcessors());
         registry.gauge(MetricName.builder().safeName("os.load.1").build(), osMxBean::getSystemLoadAverage);
+        if (osMxBean instanceof com.sun.management.OperatingSystemMXBean) {
+            com.sun.management.OperatingSystemMXBean sunBean = (com.sun.management.OperatingSystemMXBean) osMxBean;
+            registry.gauge(MetricName.builder().safeName("os.process.cpu.load").build(), sunBean::getProcessCpuLoad);
+            registry.gauge(MetricName.builder().safeName("os.system.cpu.load").build(), sunBean::getSystemCpuLoad);
```
carterkozak: perhaps `utilization` instead of `load`?

mglazer: So the problem with this metric, is it's going to duplicate exactly what other mechanisms on the same server are already going to capture. Does it make sense for us to *also* produce this from every Java application (but not provide similar metrics from any Go/Rust application)?

carterkozak: I wouldn't mind removing the system metric in favor of only reporting the load from the java process. A added them both so that we can calculate percentage of overall load, but that will only be marginally more useful than just os.process.cpu

tritium-metrics-jvm/src/test/java/com/palantir/tritium/metrics/jvm/JvmMetricsTest.java
```diff
@@ -82,7 +82,8 @@
             "jvm.threads.terminated.count",
             "jvm.threads.blocked.count",
             "os.load.1",
-            "os.load.norm.1"
+            "os.load.norm.1",
+            "os.process.cpu.utilization"
```
dansanduleac: What if we called this `jvm.cpu.utilization`, since it applies only to this JVM?
The os prefix is a bit confusing as in the other 2 cases, it refers to OS-wide metrics.

carterkozak: Possibly, but `process.cpu.utilization` might be better, we could compare this between java/rust/go services, not just those running in a jvm. Thoughts?

CRogers:
Probably useful to collect, but we should be wary about what cpu utilization means from linux (as it counts stalled tasks as well those executing it normally isn't saying the CPU is the bottleneck, which is what most people expect). More at http://www.brendangregg.com/blog/2017-05-09/cpu-utilization-is-wrong.html

CRogers (approved):

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/OperatingSystemMetrics.java
```diff
@@ -22,12 +22,19 @@
 import java.lang.management.OperatingSystemMXBean;
 
 final class OperatingSystemMetrics {
+    private static final MetricName OS_LOAD_NORM_1 = MetricName.builder().safeName("os.load.norm.1").build();
+    private static final MetricName OS_LOAD_1 = MetricName.builder().safeName("os.load.1").build();
+    private static final MetricName OS_PROCESS_CPU_UTILIZATION
+            = MetricName.builder().safeName("process.cpu.utilization").build();
```
raiju: @carterkozak Should this be `os.process.cpu.utilization`?

PR #337, opened 2019-08-08
Check unused dependencies

schlosna:
## Before this PR

We were not checking that modules declared minimum possible dependencies.

## After this PR
==COMMIT_MSG==
Check that all modules declare minimum dependencies
==COMMIT_MSG==

## Possible downsides?

Currently blocked by https://github.com/palantir/gradle-baseline/issues/747

### DIFF
--- a/tritium-caffeine/build.gradle
+++ b/tritium-caffeine/build.gradle
@@ -5,18 +5,14 @@ targetCompatibility = 1.8
 
 dependencies {
 
-    api project(':tritium-api')
-    api project(':tritium-core')
     api project(':tritium-metrics')
     api project(':tritium-registry')
     api 'com.github.ben-manes.caffeine:caffeine'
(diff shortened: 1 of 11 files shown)


tritium-caffeine/build.gradle
```diff
@@ -5,18 +5,14 @@ targetCompatibility = 1.8
 
 dependencies {
 
-    api project(':tritium-api')
-    api project(':tritium-core')
```
carterkozak, line 9: nice

carterkozak (approved):
lgtm, thanks

tritium-caffeine/build.gradle
schlosna: these are still transitively exposed as `api` via `tritium-metrics`, so consumers' view is same, but simplifies things here

PR #338, opened 2019-08-09
Upgrade to null-away 0.7.5

schlosna:
## Before this PR

Using outdated nullaway gradle checks

## After this PR
==COMMIT_MSG==
Upgrade to null-away 0.7.5
==COMMIT_MSG==


carterkozak (approved):

versions.props
```diff
@@ -13,7 +13,7 @@ com.palantir.remoting3:* = 3.46.0
 com.palantir.safe-logging:* = 1.10.1
 com.palantir.tracing:* = 3.1.0
 com.squareup.okhttp3:okhttp = 4.0.0
-com.uber.nullaway:nullaway = 0.6.4
+com.uber.nullaway:nullaway = 0.7.5
```
iamdanfox, line 16: Have also added this to excavator-versions-props so you should get this automatically next time :)

schlosna: Thanks!

PR #354, opened 2019-08-26
Treat warnings as compile errors

schlosna:
## Before this PR

Warnings were not treated as compile time errors

## After this PR
==COMMIT_MSG==
Treat warnings as compile errors
==COMMIT_MSG==

## Possible downsides?

Previously this was blocked by Immutables annotations causing spurious warnings for generated JSON classes in https://github.com/immutables/immutables/issues/291 but these are no longer blocking.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -75,7 +75,7 @@ allprojects {
 
         tasks.withType(JavaCompile).configureEach {
             options.compilerArgs += [
-                // '-Werror', // disable due to https://github.com/immutables/immutables/issues/291 for now
+                '-Werror',
                 '-Xlint:deprecation',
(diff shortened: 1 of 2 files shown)


carterkozak (approved):
I don't think a release note is necessary since this doesn't impact tritium consumers, but it doesn't hurt anything.

carterkozak:
@schlosna actually, I don't think this quite works because we have `errorProneApply=true` set in `gradle.properties`, which disables some validation to allow fixes to be applied without failing the build.

carterkozak:
PR incoming shortly

edit: PR is up: https://github.com/palantir/tritium/pull/355

PR #355, opened 2019-08-26
Unset `errorProneApply=true` in gradle.properties to enable validation

carterkozak:
The `errorProneApply` flag disables several warnings in order to
successfully migrate repos away from deprecated methods when
automated ifxes are present.

==COMMIT_MSG==
Unset `errorProneApply=true` in gradle.properties to enable validation
==COMMIT_MSG==

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,6 +66,10 @@ allprojects {
                 because "Spark still uses 3.X, which can't co-exist with 4.X"
             }
         }
+
+        // Work around https://github.com/immutables/immutables/issues/291
+        compileOnly 'org.immutables:value::annotations'
+        testCompileOnly 'org.immutables:value::annotations'
     }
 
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -66,6 +66,10 @@ allprojects {
                 because "Spark still uses 3.X, which can't co-exist with 4.X"
             }
         }
+
+        // Work around https://github.com/immutables/immutables/issues/291
+        compileOnly 'org.immutables:value::annotations'
+        testCompileOnly 'org.immutables:value::annotations'
```
carterkozak, line 72: Adds immutables annotations as a compile-only dep to every project.

schlosna (approved):
Interesting, I did not realize that `errorProneApply=true` would disable some validations. Thanks for fixing the immutables issue.

PR #356, opened 2019-08-26
Use Blackhole instead of AsyncSlf4jSpanObserver in benchmarks

carterkozak:
## Before this PR
Tracing spans were consumed using an slf4j subscriber we don't use in production at a logging level that's disabled, allowing the jit to potentially optimize away code that runs when tracing is configured.

## After this PR
==COMMIT_MSG==
Prevent tracing DCE in ProxyBenchmark
==COMMIT_MSG==


carterkozak:
closing, this is included in #357

PR #357, opened 2019-08-26
Optimize TracingInvocationEventHandler for nested use

carterkozak:
Prevent trace name calculation for unsampled spans.
Longer term we should proivide a better startSpan API
which either takes a string supplier, or uses slf4j style
interpolation.

Before:
```
Benchmark                                     Mode  Cnt    Score    Error  Units
ProxyBenchmark.instrumentedWithTracingNested  avgt    5  900.542 ± 29.627  ns/op
```

After:
```
Benchmark                                     Mode  Cnt    Score   Error  Units
ProxyBenchmark.instrumentedWithTracingNested  avgt    5  214.656 ± 7.242  ns/op
```

## After this PR
==COMMIT_MSG==
Optimize TracingInvocationEventHandler for nested use
==COMMIT_MSG==

### DIFF
--- a/tritium-tracing/src/main/java/com/palantir/tritium/tracing/TracingInvocationEventHandler.java
+++ b/tritium-tracing/src/main/java/com/palantir/tritium/tracing/TracingInvocationEventHandler.java
@@ -90,6 +90,15 @@ public final class TracingInvocationEventHandler extends AbstractInvocationEvent
     }
 
     static BooleanSupplier getEnabledSupplier(String component) {
-        return InstrumentationProperties.getSystemPropertySupplier(component);
+        BooleanSupplier systemPropertyEnabled = InstrumentationProperties.getSystemPropertySupplier(component);
+        return () -> systemPropertyEnabled.getAsBoolean() && !inUnsampledTrace();
+    }
+
+    /**
+     * In an unsampled trace, there's no reason to record additional spans. Note that this will create new root spans
+     * if not present in order to create a new traceId regardless of sampling state to preserve behavior.
+     */
+    static boolean inUnsampledTrace() {
+        return Tracer.hasTraceId() && !Tracer.isTraceObservable();
     }
 }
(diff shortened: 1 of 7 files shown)


tritium-tracing/src/main/java/com/palantir/tritium/tracing/TracingInvocationEventHandler.java
```diff
@@ -90,6 +90,15 @@ private void complete(@Nullable InvocationContext context) {
...
+        BooleanSupplier systemPropertyEnabled = InstrumentationProperties.getSystemPropertySupplier(component);
+        return () -> systemPropertyEnabled.getAsBoolean() && !inUnsampledTrace();
+    }
+
+    /**
+     * In an unsampled trace, there's no reason to record additional spans. Note that this will create new root spans
+     * if not present in order to create a new traceId regardless of sampling state to preserve behavior.
+     */
+    static boolean inUnsampledTrace() {
+        return Tracer.hasTraceId() && !Tracer.isTraceObservable();
```
schlosna, line 102: probably overkill optimization, but could push this upstream to `Tracer` to avoid the double thread local lookup:

```
    public static boolean isUnsampledTrace() {
        Trace trace = currentTrace.get();
        return trace != null && !trace.isObservable();
    }
```

schlosna (approved):

tritium-tracing/src/main/java/com/palantir/tritium/tracing/TracingInvocationEventHandler.java
carterkozak: Longer term I'd prefer to add an API to take care of the expensive piece, so we don't need to think about this, e.g.

```java
// lambda allocation, but the API is easy to use
Tracer.fastStartSpan(() -> "foo" + bar);
```
or
```java
// garbage free, but requires many overloads.
Tracer.fastStartSpan("foo {}", bar);
```

PR #358, opened 2019-08-26
Optimize TaggedMetricsServiceInvocationEventHandler

carterkozak:
Create the global failure meter eagerly, cache timers by method
to avoid recreating metric name objects.

Before:
```
Benchmark                                     Mode  Cnt    Score     Error  Units
ProxyBenchmark.instrumentedWithTaggedMetrics  avgt    5  669.303 ± 108.389  ns/op
```

After:
```
Benchmark                                     Mode  Cnt    Score    Error  Units
ProxyBenchmark.instrumentedWithTaggedMetrics  avgt    5  233.019 ± 49.507  ns/op
```

## After this PR
==COMMIT_MSG==
Optimize TaggedMetricsServiceInvocationEventHandler
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/TaggedMetricsServiceInvocationEventHandler.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/TaggedMetricsServiceInvocationEventHandler.java
@@ -18,6 +18,8 @@ package com.palantir.tritium.event.metrics;
 
 import static com.palantir.logsafe.Preconditions.checkNotNull;
 
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.Timer;
 import com.palantir.tritium.event.AbstractInvocationEventHandler;
 import com.palantir.tritium.event.DefaultInvocationContext;
 import com.palantir.tritium.event.InstrumentationProperties;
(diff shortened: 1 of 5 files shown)


tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/TaggedMetricsServiceInvocationEventHandler.java
```diff
@@ -103,6 +115,6 @@ public final void onFailure(@Nullable InvocationContext context, @Nonnull Throwa
     }
 
     private void markGlobalFailure() {
-        taggedMetricRegistry.meter(FAILURES_METRIC).mark();
+        globalFailureMeter.mark();
```
schlosna: should we just inline this now?

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/TaggedMetricsServiceInvocationEventHandler.java
```diff
@@ -77,16 +91,14 @@ public final void onSuccess(@Nullable InvocationContext context, @Nullable Objec
...
         }
     }
 
+    private Timer getSuccessTimer(Method method) {
+        return timerCache.computeIfAbsent(method, onSuccessTimerMappingFunction);
+    }
+
     @Override
     public final void onFailure(@Nullable InvocationContext context, @Nonnull Throwable cause) {
         markGlobalFailure();
```
schlosna:
```suggestion
        globalFailureMeter.mark();
```

schlosna (approved):
overall LGTM 👍 , minor nit on possible cleanup

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/TaggedMetricsServiceInvocationEventHandler.java
carterkozak: Good call

schlosna (approved):

carterkozak:
This is stacked on top of the tracing optimization, I'll hold off merging until that has been reviewed.

PR #378, opened 2019-09-03
Tritium provides an opt-in flag to use ByteBuddy instead of dynamic proxies

carterkozak:
==COMMIT_MSG==
Tritium provides an opt-in flag to use ByteBuddy instead of dynamic proxies
==COMMIT_MSG==

If the `instrument.dynamic-proxy` property is set `false`, byte-buddy is used.

before:
```
java.lang.RuntimeException: Stack Trace
        at com.palantir.DemoMain.run(DemoMain.java:32)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:567)
        at com.palantir.tritium.proxy.InvocationEventProxy.invoke(InvocationEventProxy.java:102)
        at com.sun.proxy.$Proxy0.run(Unknown Source)
        at com.palantir.DemoMain.main(DemoMain.java:27)
```

after:
```
java.lang.RuntimeException: Stack Trace
        at com.palantir.DemoMain.run(DemoMain.java:32)
        at com.palantir.tritium.proxy.InstrumentedRunnable$0.run(Unknown Source)
        at com.palantir.DemoMain.main(DemoMain.java:27)
```

Notable differences:

* The bytebuddy implementation cannot instrument inaccessible interfaces, so it will log a warning and return the uninstrumented delegate object.
* Added an implementation dependency on byte-buddy, increasing the footprint of tritium-lib.

Benchmark results (haven't changed much, but they weren't the goal):
```
Benchmark                                                 (mode)  Mode  Cnt    Score    Error  Units
ProxyBenchmark.instrumentedDefaultTagged              BYTE_BUDDY  avgt    5  401.312 ± 17.375  ns/op
ProxyBenchmark.instrumentedDefaultTagged           DYNAMIC_PROXY  avgt    5  408.340 ± 10.732  ns/op
ProxyBenchmark.instrumentedDefaultUntagged            BYTE_BUDDY  avgt    5  518.907 ± 39.311  ns/op
ProxyBenchmark.instrumentedDefaultUntagged         DYNAMIC_PROXY  avgt    5  508.477 ±  7.854  ns/op
ProxyBenchmark.instrumentedWithEnabledNopHandler      BYTE_BUDDY  avgt    5   15.254 ±  0.457  ns/op
ProxyBenchmark.instrumentedWithEnabledNopHandler   DYNAMIC_PROXY  avgt    5   19.069 ±  0.555  ns/op
ProxyBenchmark.instrumentedWithEverything             BYTE_BUDDY  avgt    5  418.855 ± 57.766  ns/op
ProxyBenchmark.instrumentedWithEverything          DYNAMIC_PROXY  avgt    5  475.463 ± 46.280  ns/op
ProxyBenchmark.instrumentedWithMetrics                BYTE_BUDDY  avgt    5  285.950 ± 22.050  ns/op
ProxyBenchmark.instrumentedWithMetrics             DYNAMIC_PROXY  avgt    5  295.070 ± 15.405  ns/op
ProxyBenchmark.instrumentedWithPerformanceLogging     BYTE_BUDDY  avgt    5    7.195 ±  0.074  ns/op
ProxyBenchmark.instrumentedWithPerformanceLogging  DYNAMIC_PROXY  avgt    5   13.690 ±  0.344  ns/op
ProxyBenchmark.instrumentedWithTaggedMetrics          BYTE_BUDDY  avgt    5  214.352 ± 31.053  ns/op
ProxyBenchmark.instrumentedWithTaggedMetrics       DYNAMIC_PROXY  avgt    5  222.926 ± 62.158  ns/op
ProxyBenchmark.instrumentedWithTracing                BYTE_BUDDY  avgt    5  212.589 ±  6.832  ns/op
ProxyBenchmark.instrumentedWithTracing             DYNAMIC_PROXY  avgt    5  219.651 ±  6.300  ns/op
ProxyBenchmark.instrumentedWithTracingNested          BYTE_BUDDY  avgt    5  181.806 ±  3.045  ns/op
ProxyBenchmark.instrumentedWithTracingNested       DYNAMIC_PROXY  avgt    5  195.890 ±  3.228  ns/op
ProxyBenchmark.raw                                    BYTE_BUDDY  avgt    5    2.227 ±  0.095  ns/op
ProxyBenchmark.raw                                 DYNAMIC_PROXY  avgt    5    2.204 ±  0.058  ns/op
```


tritium-lib/src/test/java/com/palantir/tritium/proxy/InstrumentationTest.java
```diff
@@ -62,20 +63,24 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+@ResourceLock("SYSTEM_PROPERTIES")
 @ExtendWith(MockitoExtension.class)
-@SuppressWarnings("NullAway") // mock injection
-final class InstrumentationTest {
+@SuppressWarnings({"NullAway", "WeakerAccess"}) // mock injection
+@ResourceLock("SYSTEM_PROPERTIES")
+public abstract class InstrumentationTest {
```
carterkozak, line 78: moved to public otherwise we don't have access to instrument inner interfaces

tritium-lib/src/test/java/com/palantir/tritium/proxy/InstrumentationTest.java
```diff
@@ -575,11 +587,13 @@ void testStackDepth_notFunctionOfHandlers() {
         assertThat(singleHandlerInstrumentedStackTrace).hasSameSizeAs(multipleHandlerInstrumentedStack);
     }
 
-    public interface StackTraceSupplier extends Supplier<StackTraceElement[]> {}
+    public interface StackTraceSupplier {
+        StackTraceElement[] get();
+    }
```
carterkozak, line 594: Modified to avoid lambda class 'hidden' frames. They both work, the `extends Supplier` version has an extra stack frame.

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentation.java
```diff
@@ -0,0 +1,234 @@
...
+                    UnsafeArg.of("delegate", delegate),
+                    SafeArg.of("delegateType", delegate.getClass()));
+            return delegate;
+        }
+
+        // Use the interface classloader to avoid creating additional proxies for other loaders delegating
+        // to the same type.
+        ClassLoader classLoader = getClassLoader(interfaceClass);
+        ImmutableList<Class<?>> interfaces = Arrays.stream(Proxies.interfaces(interfaceClass, delegate.getClass()))
+                .filter(iface -> isAccessibleFrom(classLoader, iface))
```
schlosna: Are we going to surprise folks with not fulfilling Liskov substitution principle if we filter out implemented interfaces here?

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentation.java
```diff
@@ -0,0 +1,234 @@
...
+            return (T) newInstrumentationClass(classLoader, interfaces)
+                    .getConstructor(interfaceClass, InvocationEventHandler.class, InstrumentationFilter.class)
+                    .newInstance(delegate, CompositeInvocationEventHandler.of(handlers), instrumentationFilter);
+        } catch (ReflectiveOperationException e) {
+            throw new SafeRuntimeException("Failed to instrumented delegate", e);
+        }
+    }
+
+    private static Class<?> newInstrumentationClass(ClassLoader classLoader, List<Class<?>> interfaces) {
+        Class<?> interfaceClass = interfaces.get(0);
```
schlosna: This assumes primary interface is first, do we test this?

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentation.java
```diff
@@ -0,0 +1,234 @@
...
+                    .load(classLoader, ClassLoadingStrategy.Default.WRAPPER)
+                    .getLoaded();
+        });
+    }
+
+    private static ClassLoader getClassLoader(Class<?> clazz) {
+        ClassLoader loader = clazz.getClassLoader();
+        // Some internal classes return a null classloader, in these cases we provide the instrumentation loader.
+        if (loader == null) {
+            return ByteBuddyInstrumentation.class.getClassLoader();
```
schlosna: Should we prefer thread context class-loader before this class’?

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentation.java
```diff
@@ -0,0 +1,234 @@
...
+        return loader;
+    }
+
+    private static boolean isAccessible(Class<?> clazz) {
+        // Base case for enclosing class accessibility check
+        if (clazz == null) {
+            return true;
+        }
+        return (clazz.isInterface() || Modifier.isPublic(clazz.getModifiers()))
+                && isAccessible(clazz.getEnclosingClass());
```
schlosna: What happens for interfaces defined in a non-interface class?

tritium-lib/src/test/java/com/palantir/tritium/proxy/InstrumentationTest.java
```diff
@@ -551,7 +563,7 @@ void testStackDepth() {
         // The value isn't particularly important, this test exists to force us to acknowledge changes in
         // stack trace length due to Tritium instrumentation. It's not uncommon to have >10 Tritium proxies
         // in a single trace, so increases in frames can make debugging more difficult.
-        assertThat(instrumentedStack).hasSize(rawStack.length + 6);
+        assertThat(instrumentedStack).hasSize(rawStack.length + (useByteBuddy() ? 1 : 6));
```
schlosna: Nit: is it worth extracting to 
```suggestion
    int instrumentedStackSize = useByteBuddy() ? 1 : 6;
     assertThat(instrumentedStack).hasSize(rawStack.length + instrumentedStackSize);
```

schlosna (commented):
Overall LGTM 👍

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentation.java
carterkozak: Not unless they do unsafe casting, our type-safe API only guarantees that we implement the passed interface, not additional interfaces implemented by the delegate. It is possible that this will lead to a behavior change, though unlikely.

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentation.java
carterkozak: I went back and forth on passing `ifaceClass, additionalInterfaces` vs a single list of interfaces, this approach lead to fewer lines, but your raise a good point that it comes at expense to clarity, I'll refactor.

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentation.java
carterkozak: This recursively checks that enclosing classes are public, otherwise we cannot implement the interface.

bug: I also need to check if interfaces are public, InstrumentationTest happens to pass because it's in the same package as `ByteBuddyInstrumentation`.

carterkozak: Adding a slew of tests for these scenarios.

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentation.java
schlosna: I don't think you necessarily have to refactor, I'd just suggest docs + test explaining assumption

carterkozak: I read a blog post a few years ago that I can't seem to find anymore, but the gist was that we should push failures/validation to the "left" in the chain of compiler -> unit tests -> integration/ete -> runtime. I think the refactor I've pushed obviates the need for a unit test.

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentation.java
carterkozak: I think we should check the context classloader, but we should prefer this one to avoid caching the same class against several child ccls.

I should add validation that the fallback classloader can actually load the provided class. It may be worthwhile checking `ByteBuddyInstrumentation.class.getClassLoader()` before the serviceInterface classloader to allow for potentially more additional interfaces to be instrumented.

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentation.java
schlosna: I'm fully supportive of having the compiler robots identify issues earlier in the pipeline 😄

versions.lock
```diff
@@ -33,7 +34,6 @@ ch.qos.logback:logback-core:1.1.2 (1 constraints: 050dfb1c)
 com.squareup.okhttp3:okhttp:4.0.0 (1 constraints: 0605ff35)
 com.squareup.okio:okio:2.2.2 (1 constraints: 500c29fd)
 io.undertow:undertow-core:2.0.20.Final (1 constraints: 4e07ff60)
-net.bytebuddy:byte-buddy:1.9.10 (1 constraints: 450b52de)
 net.bytebuddy:byte-buddy-agent:1.9.10 (1 constraints: 450b52de)
```
schlosna, line 37: I don't think this will cause issues, but curious if we should bump this ([possibly upstream](https://github.com/mockito/mockito/blame/25119e12946d5e79a003b4e778436a647a2faef4/gradle/dependencies.gradle#L7))?

```
$ gw why --hash 450b52de
net.bytebuddy:byte-buddy-agent:1.9.10
        org.mockito:mockito-core -> 1.9.10
```

schlosna (approved):

versions.lock
carterkozak: https://github.com/mockito/mockito/pull/1772

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentation.java
```diff
@@ -0,0 +1,275 @@
...
+                                    .to(ByteBuddyInstrumentationAdvice.class)
+                                    .wrap(allowDirectAccess
+                                            ? MethodCall.invokeSelf().onField("delegate").withAllArguments()
+                                            : MethodCall.invokeSelf()
+                                                    // Byte buddy doesn't seem to allow casting from fields, but
+                                                    // we can cast the result of a trivial call (in this case
+                                                    // Objects.requireNonNull) into the desired type.
+                                                    .onMethodCall(passThroughMethod().withField("delegate"))
+                                                    .withAllArguments()
+                                                    .withAssigner(Assigner.DEFAULT, Assigner.Typing.DYNAMIC)));
```
carterkozak, line 153: Could simplify this a bit more and remove `boolean allowDirectAccess = iface.isAssignableFrom(interfaceClass);`, using the `Objects.requireNonNull` wrapper around all accesses of the delegate. I could go either way.

PR #379, opened 2019-09-03
CompositeInvocationEventHandler completes delegates when disabled in flight

carterkozak:
InvocationEventHandler implementations are often responsible for
setting up and tearing down thread state. When a handler is
enabled at the beginning of a call, and disabled at the end, it
should still be executed to clean up what it started. This matches
expectations when only a single handler is used.

## Before this PR
preInvoke may be called without onSuccess/onFailure ever being executed.

## After this PR
==COMMIT_MSG==
CompositeInvocationEventHandler correctly completes delegate handlers when they are disabled in flight
==COMMIT_MSG==

### DIFF
--- a/tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
+++ b/tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
@@ -22,6 +22,7 @@ import com.google.common.collect.ImmutableList;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
 import java.lang.reflect.Method;
+import java.util.Arrays;
 import java.util.List;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -32,10 +33,11 @@ public final class CompositeInvocationEventHandler extends AbstractInvocationEve
 
     private static final Logger logger = LoggerFactory.getLogger(CompositeInvocationEventHandler.class);
 
-    private final List<InvocationEventHandler<InvocationContext>> handlers;
+    private final InvocationEventHandler<InvocationContext>[] handlers;
 
+    @SuppressWarnings("unchecked")
     private CompositeInvocationEventHandler(List<InvocationEventHandler<InvocationContext>> handlers) {
-        this.handlers = ImmutableList.copyOf(handlers);
+        this.handlers = ImmutableList.copyOf(handlers).toArray(new InvocationEventHandler[0]);
     }
 
     public static InvocationEventHandler<InvocationContext> of(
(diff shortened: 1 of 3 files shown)


carterkozak:
First commit  adds a test that fails on develop with a minimal fix, second commit optimizes the fix to avoid pushing methods beyond the inline limit.

tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
```diff
@@ -32,10 +33,11 @@
 
     private static final Logger logger = LoggerFactory.getLogger(CompositeInvocationEventHandler.class);
 
-    private final List<InvocationEventHandler<InvocationContext>> handlers;
+    private final InvocationEventHandler<InvocationContext>[] handlers;
 
+    @SuppressWarnings("unchecked")
     private CompositeInvocationEventHandler(List<InvocationEventHandler<InvocationContext>> handlers) {
-        this.handlers = ImmutableList.copyOf(handlers);
+        this.handlers = ImmutableList.copyOf(handlers).toArray(new InvocationEventHandler[0]);
```
schlosna: should we remove the `ImmutableList` copy?

```suggestion
        this.handlers = handlers.toArray(new InvocationEventHandler[0]);
```

tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
```diff
@@ -134,11 +135,11 @@ private static void preInvocationFailed(
     }
 
     private static void handleSuccess(
-            @Nullable InvocationEventHandler<?> handler,
+            InvocationEventHandler<?> handler,
             @Nullable InvocationContext context,
             @Nullable Object result) {
         try {
-            if (handler != null) {
+            if (context != null) {
```
schlosna: wondering if we should pull the follow up out of the `try`?

```suggestion
        if (context == null) {
            return;
        }
```

carterkozak: I was thinking the same thing, but didn't want to put too much into one PR. Might as well though.

tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
carterkozak: Good call, it's an expensive null element check, I'll write a loop.

tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
```diff
@@ -32,38 +32,41 @@
...
+        }
     }
 
     public static InvocationEventHandler<InvocationContext> of(
             List<InvocationEventHandler<InvocationContext>> handlers) {
         if (handlers.isEmpty()) {
             return NoOpInvocationEventHandler.INSTANCE;
         } else if (handlers.size() == 1) {
-            return checkNotNull(handlers.iterator().next(), "at index 0");
+            return checkNotNull(handlers.iterator().next(), "Null handlers are not allowed");
```
schlosna: IntelliJ was yelling at me about this the other day to avoid the iterator allocation:

```suggestion
            return checkNotNull(handlers.get(0), "Null handlers are not allowed");
```

schlosna (approved):

Issue #278, opened 2019-05-28
FR: alternative Gauge API that doesn't require a lambda (e.g. SettableGauge)


iamdanfox:
Gauges are different from all other metric types in that sampling them is a 'pull' operation that involves running a user-provided lambda, which could (and in practise often does) end up causing potentially very slow network requests.

This is bad because it means dumping an entire metrics registry can take a wildly unpredictable amount of time (e.g. 16 minutes).

## Proposal

Invent some kind of `SettableGauge<T>` class, expose it in both regular & tagged metric registries.

- We might want to also store the time at which a value was set, so that when the SettableGauge is sampled and we want to emit a metric.1 log line to disk, we can put an accurate timestamp next to the value we received.

Possibly worth making up specialized versions of this class for `long`, `integer` and `double`.

uschi2000:
Is this something we should discuss upstream with the dropwizard folks?

schlosna:
see also https://github.com/dropwizard/metrics/issues/1126

iamdanfox:
Closing out because we changed our reporting methodology so that each gauge reporting is scheduled independent of each other gauge, meaning that a single slow gauge can't block others.

PR #389, opened 2019-09-06
Use Byte Buddy for Instrumentation by default

carterkozak:
Previous behavior can be used by setting the
`instrument.dynamic-proxy` property to `true`.

## Before this PR
Dynamic proxies were used by default.

## After this PR
==COMMIT_MSG==
Use Byte Buddy for Instrumentation by default, reducing excess stack frames from instrumentation.
==COMMIT_MSG==

## Possible downsides?
There's always a risk in change, particularly for environments with complex class loader hierarchies and security managers. We've rolled this out to our largest project already without any issues.

### DIFF
--- a/changelog/@unreleased/pr-389.v2.yml
+++ b/changelog/@unreleased/pr-389.v2.yml
@@ -1,5 +1,5 @@
-type: improvement
-improvement:
+type: feature
+feature:
   description: Use Byte Buddy for Instrumentation by default, reducing excess stack
     frames from instrumentation.
   links:
--- a/changelog/@unreleased/pr-389.v2.yml
+++ b/changelog/@unreleased/pr-389.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Use Byte Buddy for Instrumentation by default, reducing excess stack
+    frames from instrumentation.
+  links:
+  - https://github.com/palantir/tritium/pull/389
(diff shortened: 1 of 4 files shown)


iamdanfox:
Is it worth getting Java12/13 tests running on our OSS repos too?

carterkozak:
I fully support testing everything we possibly can. I've manually verified that the relevant tests to this change pass on java 12.

changelog/@unreleased/pr-389.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
```
schlosna: probably want to tag this as a `feature` to bump major version

carterkozak: I don't think we get a major version bump on features until we release the first major version. Want to call this 1.0.0?

schlosna: I'm inclined to just release as `0.16.0` for now, I think we might want to tackle a couple more (possibly #8 & #278 ) before bumping to `1.0.0`

iamdanfox (approved):
LGTM, defer to @schlosna to pull the trigger though!

schlosna (approved):

schlosna:
Going to defer to @carterkozak when he's comfortable marking to "merge when ready" given ApacheCon this week

carterkozak:
I have comms, let's ship it.

PR #399, opened 2019-09-10
Fix #393: Loosen InvocationEventHanndler context bound to object 

### LINKED ISSUE #393: Consider loosening InvocationEventHandler generic to Object
## What happened?
Several of our handlers require a small slice of what is available on InvocationHandler:
* `TracingInvocationEventHandler` only uses `java.reflect.Method.getName()` for operation names
* `MetricsInvocationEventHandler` uses `getStartTimeNanos()` and `getMethod()`
* `CompositeInvocationEventHandler` uses no values from `InvocationContnext`, only a array of contexts.

## What did you want to happen?

We should consider loosening InvocationEventHandler generic to Object from InvocationContext. `CompositeInvocationEventHandler` may implement `InvocationEventHandler<Object[]>` and avoid copying argument arrays, additional System.nanoTime calls, etc.

carterkozak:
WIP: looking for feedback

General idea is that bounding to InvocationContext doesn't buy us much. We cannot add to it because that would break custom implementations, and contexts aren't read outside of the handler which created them. Most handlers don't use all the fields provided by InvocationContext, and some (composite, tracing) don't use any.

## Before this PR
Every handler created a full copy of the argument array, and accessed the precise system clock.

## After this PR
==COMMIT_MSG==
InvocationEventHandler context objects do not need to implement `InvocationContext`
==COMMIT_MSG==

## Possible downsides?
This is an API break, though type erasure should allow us to implement it in a way that doesn't break ABI for standard usage.

### DIFF
--- a/tritium-api/src/main/java/com/palantir/tritium/event/InvocationEventHandler.java
+++ b/tritium-api/src/main/java/com/palantir/tritium/event/InvocationEventHandler.java
@@ -26,7 +26,7 @@ import javax.annotation.Nullable;
  * @see java.lang.reflect.InvocationHandler
  * @param <C> invocation context
  */
-public interface InvocationEventHandler<C extends InvocationContext> {
+public interface InvocationEventHandler<C> {
 
     /**
      * Returns true if this event handler instance is enabled, otherwise false.
@@ -57,7 +57,7 @@ public interface InvocationEventHandler<C extends InvocationContext> {
      * @param context the current invocation context.
      * @param result the return value from invocation, or null if {@link Void}.
      */
-    void onSuccess(@Nullable InvocationContext context, @Nullable Object result);
+    void onSuccess(@Nullable C context, @Nullable Object result);
 
     /**
      * Invoked when an invocation fails.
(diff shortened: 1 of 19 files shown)


schlosna:
If we break API, should we do so to fully encapsulate context? I have not dug through all the extended handlers, do you have a sense of the impacts there?

schlosna:
The revapi build checks seem to be working 😄

carterkozak:
> If we break API, should we do so to fully encapsulate context?

I'm not sure I follow, do you mean keeping the context object completely hidden from components which access an InvocationEventHandler? I'm not sure how that would work cleanly -- the approach I've pushed is a bit dirty around CompositeInvocationEventHandler, but in turn it provides type safety and cleanliness to custom InvocationEventHandler implementations so casting results is never required.

> I have not dug through all the extended handlers, do you have a sense of the impacts there?

Custom handlers should continue to work without modification, just like the performance logging handler, which was not modified. I haven't explicitly tested ABI compatibility yet, but it should be fine because only generic parameters have changed, and code outside of tritium does not invoke InvocationEventHandler methods.

> The revapi build checks seem to be working 😄

Very nifty :-)

carterkozak:
No rush on reviewing this. It was in the back of my mind for a while and I thought I'd figure out what it would look like since we've started discussing a 1.0.0.

PR #418, opened 2019-09-13
Mark unused public method arguments

schlosna:
## Before this PR
`StrictUnusedVariable` would flag errors in tritium

## After this PR
==COMMIT_MSG==
Tritium compiles cleanly with `StrictUnusedVariable` enabled
==COMMIT_MSG==

### DIFF
--- a/tritium-lib/src/test/java/com/palantir/tritium/proxy/InvocationEventProxyTest.java
+++ b/tritium-lib/src/test/java/com/palantir/tritium/proxy/InvocationEventProxyTest.java
@@ -94,9 +94,9 @@ public class InvocationEventProxyTest {
         InvocationEventHandler<InvocationContext> testHandler = new SimpleHandler() {
             @Override
             public InvocationContext preInvocation(
-                    @Nonnull Object instance,
-                    @Nonnull Method method,
-                    @Nonnull Object[] args) {
+                    @Nonnull Object unusedInstance,
+                    @Nonnull Method unusedMethod,
+                    @Nonnull Object[] unusedArgs) {
                 throw new IllegalStateException("expected");
             }
         };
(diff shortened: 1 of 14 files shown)


tritium-lib/src/test/java/com/palantir/tritium/proxy/InvocationEventProxyTest.java
```diff
@@ -132,12 +132,12 @@ public void onSuccess(@Nullable InvocationContext context, @Nullable Object resu
     public void testInstrumentOnFailureThrows() throws Throwable {
         InvocationEventHandler<InvocationContext> testHandler = new SimpleHandler() {
             @Override
-            public void onSuccess(@Nullable InvocationContext context, @Nullable Object result) {
+            public void onSuccess(@Nullable InvocationContext unusedContext, @Nullable Object unusedResult) {
```
carterkozak, line 135: Does SuppressWarnings(“unused”) work? It feels odd to change parameter names from the interface.

schlosna: yeah, `@SuppressWarnings("unused")` or `@SuppressWarnings("StrictUnusedVariable")` will suppress the warning (error), though in general I would want to avoid that at the method level to avoid accidentally suppressing legit problems.

iamdanfox: i _think_ you can also put that annotation on a parameter, not a method.

schlosna: yep, `@SuppressWarnings("unused")` on the param itself also works. Not sure how I feel about that everywhere, it will definitely re-wrap a bunch of args, especially for what are often no-op callbacks, for example:


```
        @Override
        public void onSuccess(
                @SuppressWarnings("unused") @Nullable InvocationContext context,
                @SuppressWarnings("unused") @Nullable Object result) {}
```

vs.

```
        @Override
        public void onSuccess(@Nullable InvocationContext unusedContext, @Nullable Object unusedResult) {}
```

PR #430, opened 2019-09-18
Nested metric registries don't cause a sortedmap creation

j-baker:
This might seem like an unnecessary optimization, but profiling
internally implies this uses plenty of CPU and allocates gigabytes and
gigabytes and gigabytes (about 1.8GB of allocations to report metrics).

BLUF: With nested tags, our tooling spends all of its time recreating
nested sorted maps and rehashing, which is very expensive. That said,
breaking binary backcompat here is super expensive, as we know from past
experience.

This PR solves this issue. We no longer use the immutables generated
class, though we do use the builder. Instead, we use our own
implementation, which hashes consistently and uses two map
implementations to ensure that in the cases we see, equality, hash, and
extending with another tag are cheap.


carterkozak:
Would you mind adding a representative JMH benchmark?

j-baker:
That seems like a great idea.

CRogers:
Chatted to @j-baker on slack, would be good to see a profile on this (possibly in prod) before we merge this as it's used by every product.

carterkozak:
Could we do something with a TaggedMetricRegistry.forEach(BiConsumer<MetricName, Metric>) instead?

j-baker:
I ended up doing both (and came to the forEach conclusion as well). Here are benchmark results before and after:

```
Before                                                                    Mode  Cnt        Score         Error   Units
NestedMetricsBenchmark.benchmarkForEach                                      avgt   10   949386.155 ±  290036.990   ns/op
NestedMetricsBenchmark.benchmarkForEach:·gc.alloc.rate.norm                  avgt   10  2122032.414 ±       0.121    B/op
NestedMetricsBenchmark.benchmarkGetMetrics                                   avgt   10  1627004.644 ±  361734.975   ns/op
NestedMetricsBenchmark.benchmarkGetMetrics:·gc.alloc.rate.norm               avgt   10  2800648.214 ±      12.257    B/op

After                                                                    Mode  Cnt        Score        Error   Units
NestedMetricsBenchmark.benchmarkForEach                                      avgt   10   214789.191 ±  66852.954   ns/op
NestedMetricsBenchmark.benchmarkForEach:·gc.alloc.rate.norm                  avgt   10   322032.094 ±      0.028    B/op
NestedMetricsBenchmark.benchmarkGetMetrics                                   avgt   10   557286.570 ±  78861.278   ns/op
NestedMetricsBenchmark.benchmarkGetMetrics:·gc.alloc.rate.norm               avgt   10   997784.244 ±      0.039    B/op
```

as you can see, even in the case where we only have a single pre-existing tag, this is a considerable improvement.

j-baker:
This is also only to write 5k metrics - in practice we see many more.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/PrehashedSortedMap.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * A sorted map implementation which prehashes for faster usage in hashmaps.
+ * This is only safe for immutable underlying maps.
+ */
+final class PrehashedSortedMap<K, V> extends ForwardingSortedMap<K, V> {
+    private final SortedMap<K, V> delegate;
+    private final int hashCode;
+
+    PrehashedSortedMap(SortedMap<K, V> delegate) {
```
schlosna: would suggest making immutability explicit in types

```suggestion
    PrehashedSortedMap(ImmutableSortedMap<K, V> delegate) {
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/PrehashedSortedMap.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.google.common.collect.ForwardingSortedMap;
+import java.util.SortedMap;
+
+/**
+ * A sorted map implementation which prehashes for faster usage in hashmaps.
+ * This is only safe for immutable underlying maps.
+ */
+final class PrehashedSortedMap<K, V> extends ForwardingSortedMap<K, V> {
+    private final SortedMap<K, V> delegate;
```
schlosna: would suggest making immutability explicit in types

```suggestion
    private final ImmutableSortedMap<K, V> delegate;
```

carterkozak:
I think we should update the benchmark into a root TaggedMetricRegistry with N tagged metrics, a child wrapped DropwizardMetricSet(MetricRegistry) with N metrics, and a child TaggedMetricRegistry with N metrics (representing SharedTaggedMetricRegistries.getSingleton()).
That should scale in a way that represents production systems.

I'd prefer to put all our eggs in the forEachMetric basket and avoid optimizations on getMetrics to reduce complexity. If we provide a simple MetricName implementation which doesn't pre-compute hashCode, and avoid precomputing hashcode on ExtraEntrySortedMap, we should have strictly less work.

Might be worth special casing adding a single tag to a MetricName with no tags, and adding a simple MetricName implementation around a safeName with no tags for DropwizardTaggedMetricSet.

tritium-registry/src/test/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMapTest.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Map;
+import java.util.SortedMap;
+import java.util.function.Function;
+import net.jqwik.api.Assume;
+import net.jqwik.api.ForAll;
+import net.jqwik.api.Property;
+import net.jqwik.api.constraints.IntRange;
+import net.jqwik.api.constraints.Size;
+
+class ExtraEntrySortedMapTest {
```
CRogers, line 35: I've added this property based test as I wanted to make sure the behaviour of `ExtraEntrySortedMap` was correct but also not write a ton of example based tests. This checks that every method call on `ExtraEntrySortedMap` produces the same result as the same method call on an equivalent guava `ImmutableSortedMap`, for thousands of arbitrary base maps and extra values. It uses [jqwik](https://jqwik.net/docs/current/user-guide.html#creating-a-property).

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
```diff
@@ -0,0 +1,207 @@
...
+@SuppressWarnings({"JdkObsolete", "NullableDereference", "NullAway"})
+final class ExtraEntrySortedMap<K, V> extends AbstractMap<K, V> implements SortedMap<K, V> {
+    private final Ordering<? super K> ordering;
+    private final SortedMap<K, V> base;
+    private final K extraKey;
+    private final V extraValue;
+    private final int extraEntryHashCode;
+
+    ExtraEntrySortedMap(SortedMap<K, V> base, K extraKey, V extraValue) {
+        this.base = base;
```
schlosna: can we assert non-null?
```suggestion
        this.base = Preconditions.checkNotNull(base, "base");
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
```diff
@@ -0,0 +1,207 @@
...
+import java.util.AbstractMap;
+import java.util.AbstractSet;
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+import java.util.SortedMap;
+
+@SuppressWarnings({"JdkObsolete", "NullableDereference", "NullAway"})
```
schlosna: Understand suppressing `JdkObsolete` given `SortedMap`, but would like to not suppress the `NullableDereference ` and `NullAway` here if possible

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
```diff
@@ -0,0 +1,207 @@
...
+final class ExtraEntrySortedMap<K, V> extends AbstractMap<K, V> implements SortedMap<K, V> {
+    private final Ordering<? super K> ordering;
+    private final SortedMap<K, V> base;
+    private final K extraKey;
+    private final V extraValue;
+    private final int extraEntryHashCode;
+
+    ExtraEntrySortedMap(SortedMap<K, V> base, K extraKey, V extraValue) {
+        this.base = base;
+        this.extraKey = extraKey;
```
schlosna: seems like we assume `extraKey` and `extraValue` are non-null, can we assert?
```suggestion
        this.extraKey = Preconditions.checkNotNull(extraKey, "extraKey");
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
```diff
@@ -0,0 +1,207 @@
...
+        this.extraValue = extraValue;
```
schlosna: same as above
```suggestion
        this.extraValue = Preconditions.checkNotNull(extraValue, "extraValue");
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    @Override
+    public boolean isEmpty() {
+        return false;
+    }
+
+    @Override
+    public boolean containsKey(Object key) {
+        return extraKey.equals(key) || base.containsKey(key);
```
schlosna, line 116: if we don't ensure non-null `extraKey` at construction, this would later throw NPE

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
```diff
@@ -0,0 +1,207 @@
...
+            }
+        };
+    }
+
+    @Override
+    public Collection<V> values() {
+        return new AbstractCollection<V>() {
+            @Override
+            public Iterator<V> iterator() {
+                return keySet().stream().map(ExtraEntrySortedMap.this::get).iterator();
```
schlosna: can avoid compiler warning with:

```suggestion
                return Iterators.transform(keySet().iterator(), key -> get(key));
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
```diff
@@ -0,0 +1,207 @@
...
+    public K lastKey() {
+        if (base.isEmpty()) {
+            return extraKey;
+        }
+        return ordering.max(base.lastKey(), extraKey);
+    }
+
+    @Override
+    public int size() {
+        return base.size() + 1;
```
schlosna, line 106: is this true if the extra key is also in the base map?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
```diff
@@ -0,0 +1,207 @@
...
+        return new AbstractSet<K>() {
+            @Override
+            public Iterator<K> iterator() {
+                return Iterables.mergeSorted(ImmutableList.of(base.keySet(), ImmutableList.of(extraKey)), ordering)
+                        .iterator();
+            }
+
+            @Override
+            public int size() {
+                return base.size() + 1;
```
schlosna, line 164: is this true of extraKey is also in base map?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
```diff
@@ -0,0 +1,207 @@
...
+    public Collection<V> values() {
+        return new AbstractCollection<V>() {
+            @Override
+            public Iterator<V> iterator() {
+                return keySet().stream().map(ExtraEntrySortedMap.this::get).iterator();
+            }
+
+            @Override
+            public int size() {
+                return base.values().size() + 1;
```
schlosna, line 179: is this true of extraKey is also in base map?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
```diff
@@ -0,0 +1,207 @@
...
+    public Set<Entry<K, V>> entrySet() {
+        return new AbstractSet<Map.Entry<K, V>>() {
+            @Override
+            public Iterator<Map.Entry<K, V>> iterator() {
+                return Iterators.transform(keySet().iterator(), key -> Maps.immutableEntry(key, get(key)));
+            }
+
+            @Override
+            public int size() {
+                return base.size() + 1;
```
schlosna, line 194: is this true of extraKey is also in base map?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
```diff
@@ -0,0 +1,207 @@
...
+            if (extraKey.equals(otherMap.extraKey) && extraValue.equals(otherMap.extraValue)) {
+                return base.equals(otherMap.base);
+            }
+        }
+        return super.equals(other);
+    }
+
+    @Override
+    public int hashCode() {
+        return base.hashCode() + extraEntryHashCode;
```
schlosna, line 212: do we need to spread the hash more effectively given `String`'s poor hashCode?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/PrehashedSortedMap.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package com.palantir.tritium.metrics.registry;
+
+import com.google.common.collect.ForwardingSortedMap;
+import java.util.SortedMap;
+
+/**
+ * A sorted map implementation which prehashes for faster usage in hashmaps.
+ * This is only safe for immutable underlying maps.
```
schlosna: this assumes not only immutable map (i.e. can't add/remove from map), but also immutable entries such that you can't change the value for a given key, and that the keys/values themselves can't mutate or else the hashcode is no longer valid, and bad things happen

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/RealMetricName.java
```diff
@@ -0,0 +1,88 @@
...
+    private RealMetricName(String safeName, SortedMap<String, String> safeTags) {
+        this.safeName = safeName;
+        this.safeTags = safeTags;
+        this.hashCode = computeHashCode();
+    }
+
+    private int computeHashCode() {
+        int hash = 5381;
+        hash += (hash << 5) + safeName().hashCode();
+        hash += (hash << 5) + safeTags().hashCode();
```
schlosna, line 39: just curious why djb2 hash algorithm?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
CRogers: I wonder if we should check that `extraKey` does not exist in the base map. If this was an ordinary `ImmtuableSortedMap`, it would not allow you to do this, so this PR will open up the ability to "shadow" existing tags, which if people use may make it hard to revert to the previous behaviour, if needed.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
j-baker: Unfortunately ErrorProne is claiming that `ExtraEntrySortedMap.this` is null

j-baker: i can probably add some `@Nullables` to satisfy nullaway

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
j-baker: oops, I meant to validate that one

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
j-baker: `Map` defines the contract of `Map.hashCode` as sum of hashcode of entries (hashCode of entries is key.hashCode ^ value.hashCode)

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/PrehashedSortedMap.java
j-baker: yep - changed comment

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/RealMetricName.java
j-baker: it's the one in the immutables generated code

j-baker: copy-pasted :P

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
j-baker: yeah, done

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/PrehashedSortedMap.java
j-baker: This would require casting in RealMetricName, which I'd like to avoid.

schlosna: not sure I follow that as the only caller of this constructor is `com.palantir.tritium.metrics.registry.RealMetricName#prehash` which explicitly passes in an `ImmutableSortedMap`

```
        return new PrehashedSortedMap<>(ImmutableSortedMap.copyOfSorted(map));
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
j-baker: i solved all these i think

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/PrehashedSortedMap.java
j-baker: oops!

.baseline/checkstyle/checkstyle.xml
```diff
@@ -469,7 +469,7 @@
         <module name="NestedTryDepth"/> <!-- Java Coding Guide: Try/catch blocks: never nested -->
         <module name="NonEmptyAtclauseDescription"/> <!-- Java Style Guide: At-clauses -->
         <module name="ParameterName"> <!-- Java Style Guide: Parameter names -->
-            <property name="format" value="^[a-z][a-zA-Z0-9]+$"/>
+            <property name="format" value="^_?[a-z][a-zA-Z0-9]+$"/>
```
j-baker, line 472: as updated by baseline

schlosna: assume we will also pick this up next time the robots do their rounds in https://github.com/palantir/gradle-baseline/releases/tag/2.7.0 though I don't see a tritium excavator listed in https://github.com/palantir/gradle-baseline/pull/833

schlosna (approved):
LGTM 👍 , thanks James! Defer to you all when you want to merge & cut a release

PR #451, opened 2019-10-03
Report Process CPU time as `process.cpu.time.nanoseconds`

carterkozak:
==COMMIT_MSG==
Report Process CPU time as `process.cpu.time.nanoseconds`
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/OperatingSystemMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/OperatingSystemMetrics.java
@@ -24,8 +24,10 @@ import java.lang.management.OperatingSystemMXBean;
 final class OperatingSystemMetrics {
     private static final MetricName OS_LOAD_NORM_1 = MetricName.builder().safeName("os.load.norm.1").build();
     private static final MetricName OS_LOAD_1 = MetricName.builder().safeName("os.load.1").build();
-    private static final MetricName OS_PROCESS_CPU_UTILIZATION
+    private static final MetricName PROCESS_CPU_UTILIZATION
             = MetricName.builder().safeName("process.cpu.utilization").build();
+    private static final MetricName PROCESS_CPU_TIME
+            = MetricName.builder().safeName("process.cpu.time.nanoseconds").build();
 
     static void register(TaggedMetricRegistry registry) {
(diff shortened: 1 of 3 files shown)


tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/OperatingSystemMetrics.java
```diff
@@ -24,16 +24,19 @@
...
 
     static void register(TaggedMetricRegistry registry) {
         OperatingSystemMXBean osMxBean = ManagementFactory.getOperatingSystemMXBean();
         registry.gauge(OS_LOAD_NORM_1, () -> osMxBean.getSystemLoadAverage() / osMxBean.getAvailableProcessors());
         registry.gauge(OS_LOAD_1, osMxBean::getSystemLoadAverage);
         if (osMxBean instanceof com.sun.management.OperatingSystemMXBean) {
             com.sun.management.OperatingSystemMXBean sunBean = (com.sun.management.OperatingSystemMXBean) osMxBean;
-            registry.gauge(OS_PROCESS_CPU_UTILIZATION, sunBean::getProcessCpuLoad);
+            registry.gauge(PROCESS_CPU_UTILIZATION, sunBean::getProcessCpuLoad);
+            registry.gauge(PROCESS_CPU_TIME, sunBean::getProcessCpuTime);
```
schlosna, line 39: are we getting similar per process CPU time from OS side as well (e.g. `/proc/$PID/stat` contains both user & system time)?

carterkozak: It _should_ be the same because we're pulling from the same place (theoretically)

linux: https://github.com/openjdk/jdk/blob/4084a43111f2e668acdf5e66a6c1d156bb58918e/src/jdk.management/unix/native/libmanagement_ext/OperatingSystemImpl.c#L302-L303
mac: https://github.com/openjdk/jdk/blob/4084a43111f2e668acdf5e66a6c1d156bb58918e/src/jdk.management/unix/native/libmanagement_ext/OperatingSystemImpl.c#L276-L278

I can verify on a linux box later in the day.

carterkozak:
Putting this on hold, I think folks are working on reporting this with node level metrics.

PR #468, opened 2019-10-17
Improved remoting3 tracing fallback handling

schlosna:
## Before this PR
When remoting3 tracing is detected and fallback is required we would noisily log error messages each time a tracer is created.

Ideally all folks would have upgraded off of remoting3 to conjure, but that has not happened everywhere yet, so we should improve our handling of fallback scenarios.

## After this PR
==COMMIT_MSG==
When remoting3 tracing is detected and fallback is required:
  * Memoize fallback determination and tracer to reduce overhead
  * Only log the first time fallback is required to reduce noise
==COMMIT_MSG==

## Possible downsides?
Maintenance burden of remoting3 back-compat and its older, less optimal tracing library.

### DIFF
--- a/tritium-tracing/src/main/java/com/palantir/tritium/tracing/RemotingCompatibleTracingInvocationEventHandler.java
+++ b/tritium-tracing/src/main/java/com/palantir/tritium/tracing/RemotingCompatibleTracingInvocationEventHandler.java
@@ -19,7 +19,9 @@ package com.palantir.tritium.tracing;
 import static com.palantir.logsafe.Preconditions.checkNotNull;
 
 import com.google.common.base.Strings;
+import com.google.common.base.Suppliers;
 import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 import com.palantir.tritium.event.AbstractInvocationEventHandler;
 import com.palantir.tritium.event.DefaultInvocationContext;
 import com.palantir.tritium.event.InstrumentationProperties;
@@ -28,6 +30,8 @@ import com.palantir.tritium.event.InvocationEventHandler;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


tritium-tracing/src/main/java/com/palantir/tritium/tracing/RemotingCompatibleTracingInvocationEventHandler.java
```diff
@@ -38,6 +42,44 @@
...
+    });
+
+    private static final Supplier<Tracer> tracerFactory = Suppliers.memoize(() -> {
+        // Ugly reflection based check to determine what implementation of tracing to use to avoid duplicate
+        // traces as remoting3's tracing classes delegate functionality to tracing-java in remoting 3.43.0+
+        // and remoting3.tracing.Tracers.wrap now returns a "com.palantir.tracing.Tracers" implementation.
+        //
+        // a) remoting3 prior to 3.43.0+ -> use reflection based remoting3 tracer
+        // b) remoting3 3.43.0+ -> use tracing-java
+        // c) no remoting3 -> use tracing-java
```
carterkozak, line 63: At what point do you think we can remove remove the remoting3 implementation?

tritium-tracing/src/main/java/com/palantir/tritium/tracing/RemotingCompatibleTracingInvocationEventHandler.java
```diff
@@ -118,43 +151,18 @@ private static boolean isUsingMultipleTracers(Class<?> tracersClass)
...
-                    logger.error("Multiple tracing implementations detected, expected '{}' but found '{}',"
-                                    + " using legacy remoting3 tracing for backward compatibility",
-                            SafeArg.of("expectedPackage", expectedTracingPackage),
-                            SafeArg.of("actualPackage", actualTracingPackage));
+                    if (shouldLogFallbackError.compareAndSet(false, true)) {
+                        logger.error("Multiple tracing implementations detected, expected '{}' but found '{}',"
+                                        + " using legacy remoting3 tracing for backward compatibility",
+                                SafeArg.of("expectedPackage", expectedTracingPackage),
+                                SafeArg.of("actualPackage", actualTracingPackage),
+                                new SafeIllegalStateException("Multiple tracing implementations detected"));
```
carterkozak, line 159: I'm not sure the exception stack trace will help us debug, maybe the ClassLoader.toString?

tritium-tracing/src/main/java/com/palantir/tritium/tracing/RemotingCompatibleTracingInvocationEventHandler.java
```diff
@@ -48,7 +90,7 @@ public RemotingCompatibleTracingInvocationEventHandler(String component, Tracer
     }
 
     static InvocationEventHandler<InvocationContext> create(String component) {
-        return new RemotingCompatibleTracingInvocationEventHandler(component, createTracer());
+        return new RemotingCompatibleTracingInvocationEventHandler(component, tracerFactory.get());
```
carterkozak, line 93: :+1: very clean

tritium-tracing/src/main/java/com/palantir/tritium/tracing/RemotingCompatibleTracingInvocationEventHandler.java
schlosna: I'd love to remove it, but there are some services with remoting3 (and even remoting2) on classpaths that need to be upgraded first.

carterkozak:
:+1: defer to your judgement on the stack trace, no strong opinion

tritium-tracing/src/main/java/com/palantir/tritium/tracing/RemotingCompatibleTracingInvocationEventHandler.java
iamdanfox: @felixdesouza atlasdb has fully moved off remoting2/3 clients right? All the other usages internally seemed to be just ObjectMappers.

felixdesouza: So we still reference remoting2 `RemoteException` and a few other error decoder related classes for our "custom" feign proxy stuff. We are in the experimental phases of running pure conjure-java-runtime for our clients, and when we get full confidence, we can remove all of that and we'll have no references to it. This is likely on the order of a couple of weeks. @jeremyk-91 @jkozlowski for SA.

carterkozak: I think it's only the `remoting3 prior to 3.43.0` case that we're worried about here, right? Are we aware of any services that satisfy:
1. Use remoting3 prior to 3.43.0
2. Take Tritium upgrades
3. Use tracing

I'm pretty confident that the issue which lead to this PR was in an environment that cannot take advantage of tracing, regardless of the implementation.

PR #293, opened 2019-06-25
Reenable error-prone checks

schlosna:

PR #474, opened 2019-10-18
Add tritium-platform to align versions

pkoenig10:
## Before this PR
It is possible to have misaligned versions of tritium jars on your classpath. This can cause issue when there are unintentional ABI breaks (see https://github.com/palantir/tritium/pull/293#discussion_r298641662).

## After this PR
It is no longer possible to have misaligned versions of tritium jars on your classpath. The platform dependency adds dependency constraints to ensure that all tritium jars are the same version.

See the docs for more info:
https://docs.gradle.org/current/userguide/managing_transitive_dependencies.html#aligning_versions_natively_with_gradle


carterkozak:
I like the consistency provided by the platform, but I'm curious if we can roll it out to everyone with some combination of a baseline plugin and potential consistent-publish-scripts change? That way we solve the problem in all of our projects.

pkoenig10:
Rolling this out to everyone should be pretty low lift, provided that:
1. Dependency upgrade PRs are getting merged
2. They are using GCV

As soon as one dependency requires a version of any tritium jar with the platform dependency, all other tritium jars will be resolved to that version as well. Because Witchcraft has a tritium dependency and we're generally pretty good about keeping Witchcraft up-to-date, we should be able to roll this out widely pretty quickly.

Are there scenarios where you think one of the two conditions listed above is not met?

carterkozak:
Sorry, I don’t think I was very clear. I’d like all of our repos which produce multiple jars to provide a platform locking those jars to the same version. I think we’ve run into similar issues in tracing-java!

pkoenig10:
Is there any downside to merging this now and then migrating to a standardized solution if/when one is created? I don't think creating something like this is currently on the roadmap.

iamdanfox:
Hang on @pkoenig10, I think the last line of [those docs](https://docs.gradle.org/current/userguide/managing_transitive_dependencies.html#aligning_versions_natively_with_gradle) throws a spanner in the works:

> This behavior is enforced for published components only if you use Gradle Module Metadata.

We've had some trouble with module metadata and immutables because it is _so_ expressive and intricate, also GCV didn't play nicely with it, so we've disabled it fleet wide.

This repo has [it disabled](https://github.com/palantir/tritium/blob/8c6bb2896df89a089d0f57b534ec474715fabf0e/gradle/publish-jar.gradle#L52-L55)

iamdanfox:
Going back to the original problem you mention in the PR desc, I think we're safe from ABI breaks now because we have revapi applied everywhere.  My preference would actually be for tritium to actually ship a small number of jars (possibly just one) so that it's _impossible_ to get misalignments.

pkoenig10:
Ah okay, I missed that in the docs.

> My preference would actually be for tritium to actually ship a small number of jars (possibly just one) so that it's impossible to get misalignments.

Are there plans to do this? Seems like it would be difficult to do this in a backwards-compatible way.

carterkozak:
> My preference would actually be for tritium to actually ship a small number of jars (possibly just one) so that it's impossible to get misalignments.

imo that would be more problematic -- I don't want to force all tritium consumers to add a caffeine cache dependency, for example.

PR #480, opened 2019-10-29
Fix Instrumentation for interfaces from classloaders without access to Tritium

carterkozak:
## Before this PR
An interface loaded by a classloader that isn't aware of tritium couldn't be instrumented.

## After this PR
==COMMIT_MSG==
Fix Instrumentation for interfaces in minimal classloaders that cannot access Tritium. If the loader for tritium classes cannot access the interface, an error is logged, but the application can continue.
==COMMIT_MSG==

## Possible downsides?
Executing `isClassLoadable(instrumentationClassLoader, clazz)` is not free, though we generally instrument services once at startup, and in most cases the same classloader loads both tritium and service interfaces, so the check is avoided.


tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentation.java
```diff
@@ -192,6 +204,20 @@ private ByteBuddyInstrumentation() {
...
+        ClassLoader instrumentationClassLoader = ByteBuddyInstrumentation.class.getClassLoader();
+        if (Objects.equals(instrumentationClassLoader, serviceClassLoader)) {
+            return instrumentationClassLoader;
+        }
+        return isClassLoadable(instrumentationClassLoader, clazz)
+                // Prefer the instrumentation classloader when it's broader than the services classloader
+                // to avoid problems accessing tritium from generated instrumentation, and for better
+                // generated class caching.
+                ? instrumentationClassLoader
+                : serviceClassLoader;
```
iamdanfox, line 219: not sure I fully understand the implications of this - is this running with an assumption that if there's more than one classloader involved, there are probably quite a few, so we may as well pick the same `instrumentationClassLoader` every time?

carterkozak: It's possible that serviceClassLoader includes both the service class and tritium, where the instrumentationClassLoader cannot access the service class

iamdanfox: Ah sorry i misread this as a tiebreaking thing where the classloaders are different but the service class is accessible from both.  Totally makes sense that we have to use the classloader that has the target `clazz` in it!!

iamdanfox (approved):

carterkozak:
Verifying from an RC on this branch before merging.

schlosna (approved):
This LGTM, we can should work through adding some tests to prevent future breaks, but think those can be a followup PR.

carterkozak:
Verified this prevents the failure, and logs as a warning using the RC.

PR #501, opened 2019-11-18
Register tagged gauge with replacement

schlosna:
## Before this PR

Registering gauges with `TaggedMetricRegistry` may lead to memory leaks if one intended to actually replace a previously registered gauge under the same metric name.

## After this PR
==COMMIT_MSG==
Register tagged gauge with replacement
==COMMIT_MSG==

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
@@ -61,6 +61,23 @@ public interface TaggedMetricRegistry extends TaggedMetricSet {
     Histogram histogram(MetricName metricName, Supplier<Histogram> histogramSupplier);
 
     /**
+     * Returns existing gauge metric for the specified metric name or null if none has been registered.
+     *
+     * @implNote Implementations should override this method with a more efficient mechanism.
+     *
+     * @param metricName metric name
+     * @return gauge metric or empty
+     */
+    @SuppressWarnings("unchecked")
+    default <T> Optional<Gauge<T>> gauge(MetricName metricName) {
+        Metric metric = getMetrics().get(metricName);
+        if (metric instanceof Gauge) {
+            return Optional.of((Gauge<T>) metric);
+        }
+        return Optional.empty();
+    }
+
+    /**
      * Returns existing or new gauge metric for the specified metric name.
      *
      * @param metricName metric name
@@ -69,6 +86,22 @@ public interface TaggedMetricRegistry extends TaggedMetricSet {
      */
     // This differs from MetricRegistry and takes the Gauge directly rather than a Supplier<Gauge>
     <T> Gauge<T> gauge(MetricName metricName, Gauge<T> gauge);
+
+    /**
+     * Registers and returns the specified gauge, replacing any existing gauge with the specified metric name.
+     *
+     * @param metricName metric name
+     * @param gauge gauge
+     */
+    // This differs from MetricRegistry and takes the Gauge directly rather than a Supplier<Gauge>
+    default <T> void registerOrReplaceGauge(MetricName metricName, Gauge<T> gauge) {
+        Gauge<T> existing = gauge(metricName, gauge);
+        if (existing == gauge) {
+            return;
+        }
+        remove(metricName);
+        gauge(metricName, gauge);
+    }
 
     /**
      * Returns existing or new counter metric for the specified metric name.
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
@@ -22,8 +22,10 @@ import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.Metric;
 import com.codahale.metrics.Timer;
+import com.palantir.logsafe.SafeArg;
 import java.util.Optional;
 import java.util.function.Supplier;
+import org.slf4j.LoggerFactory;
 
 /**
  * Similar to {@link com.codahale.metrics.MetricRegistry} but allows tagging of {@link Metric}s.
@@ -61,12 +63,12 @@ public interface TaggedMetricRegistry extends TaggedMetricSet {
     Histogram histogram(MetricName metricName, Supplier<Histogram> histogramSupplier);
 
     /**
-     * Returns existing gauge metric for the specified metric name or null if none has been registered.
+     * Returns existing gauge metric for the specified metric name or empty if none has been registered.
      *
      * @implNote Implementations should override this method with a more efficient mechanism.
      *
      * @param metricName metric name
-     * @return gauge metric or empty
+     * @return gauge metric or empty if none exists for this name
      */
     @SuppressWarnings("unchecked")
     default <T> Optional<Gauge<T>> gauge(MetricName metricName) {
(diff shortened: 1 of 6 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
```diff
@@ -70,6 +87,22 @@
...
     <T> Gauge<T> gauge(MetricName metricName, Gauge<T> gauge);
 
+    /**
+     * Registers and returns the specified gauge, replacing any existing gauge with the specified metric name.
+     *
+     * @param metricName metric name
+     * @param gauge gauge
+     */
+    // This differs from MetricRegistry and takes the Gauge directly rather than a Supplier<Gauge>
+    default <T> void registerOrReplaceGauge(MetricName metricName, Gauge<T> gauge) {
```
carterkozak: Not sure if the type parameter buys us anything, but I don't have much preference either way:
```suggestion
    default void registerOrReplaceGauge(MetricName metricName, Gauge<?> gauge) {
```

carterkozak: I think we should either log a warning when gauges are replaced, or return an ignorable boolean to allow callers to warn. Thoughts?

tritium-registry/src/test/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistryTest.java
```diff
@@ -90,15 +90,73 @@ void testSuppliedCounter(TaggedMetricRegistry registry) {
...
+                .isSameAs(gauge2);
+        assertThat(registry.getMetrics().get(METRIC_2))
+                .isSameAs(registry.gauge(METRIC_2).get())
+                .isSameAs(gauge2);
+        Gauge<Integer> gauge4 = intGauge(4);
+        registry.registerOrReplaceGauge(METRIC_2, gauge4);
+        assertThat(registry.getMetrics().get(METRIC_2))
+                .isSameAs(registry.gauge(METRIC_2).get())
+                .isSameAs(gauge4);
+    }
```
carterkozak, line 148: Great test coverage

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
```diff
@@ -60,16 +62,58 @@
...
+     *
+     * In most cases, one typically wants {@link #registerWithReplacement(MetricName, Gauge)} as gauges may
+     * retain references to large backing data structures (e.g. queue or cache), and if a gauge is being registered
+     * with the same name, one is likely replacing that previous data structure and the registry should not retain
+     * references to the previous version. If lookup of an existing gauge is desired, one should use
+     * {@link #gauge(MetricName)}.
      */
     // This differs from MetricRegistry and takes the Gauge directly rather than a Supplier<Gauge>
+    // @Deprecated
     <T> Gauge<T> gauge(MetricName metricName, Gauge<T> gauge);
```
schlosna, line 97: @carterkozak interested in your thoughts here -- I can't currently deprecate this method yet because of the cyclical dependency between `tritium`  & `metric-schema` as metric schema will generate code that uses this deprecated method (e.g. `com.palantir.tritium.metrics.jvm.OsMetrics#loadNorm1`) so the build will fail with deprecation warnings. We could release tritium without deprecating this method, then update metric-schema to have a minimum required tritium version that provides `registerWithReplacement` and use that, though we have to be careful both metric-schema & tritium versions are bumped in concert. Another option would be to move metric-schema into tritium to avoid the cyclical project dependency and version them together (slightly related to #474)

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
```diff
@@ -60,16 +62,58 @@
 
     Histogram histogram(MetricName metricName, Supplier<Histogram> histogramSupplier);
 
+    /**
+     * Returns existing gauge metric for the specified metric name or empty if none has been registered.
+     *
+     * @implNote Implementations should override this method with a more efficient mechanism.
```
carterkozak, line 68: `@implNote` I wasn't aware of the new java8 javadoc tags

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
carterkozak: Holding off on deprecation sounds reasonable to me. I'd be okay with moving the java generator into tritium, but I like the separation between the schema itself and other components -- arguably the schema should be versioned independently of the generator implementations in the same way we manage conjure.

If we moved the plugin it would likely change the maven coordinates, which can be difficult. While we have a circular dependency between repositories, I don't think we have a terribly problematic circular dependency because the tritium-metrics and tritium-metrics-jvm depend on both metric-schema and tritium-registry.

We disable a lot of static analysis in generated sources, but I'm not aware of a way to disable deprecation warnings in generated code, this caused some issues for conjure-undertow generated endpoints where we code-gen the callers.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
```diff
@@ -60,16 +62,58 @@
...
+     */
+    // This differs from MetricRegistry and takes the Gauge directly rather than a Supplier<Gauge>
+    @SuppressWarnings("deprecation") // explicitly using as desired
+    default void registerWithReplacement(MetricName metricName, Gauge<?> gauge) {
+        Gauge<?> existing = gauge(metricName, gauge);
+        if (existing == gauge) {
+            return;
+        }
+        remove(metricName).ifPresent(removed -> LoggerFactory.getLogger(getClass())
+                .debug("Removed previously registered gauge {}", SafeArg.of("metricName", metricName)));
```
carterkozak, line 113: Can we add this logging to the abstract registry implementation as well?

carterkozak (commented):
oh goodness, I forgot to post my responses. Sorry!

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
carterkozak: Actually, in this case do we want to deprecate the method, or annotate it with `@CheckReturnValue`? In most cases I agree that this isn't the right method to use, but there are scenarios where the return value is used, where a `putIfAbsent` operation may be preferable.

If we go this route, we should probably wait until after I've had a chance to roll out a refactor to the new method.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
schlosna: its used throughout openjdk, but haven't seen it in a ton of other places

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
schlosna: I'm fine with staging the rollout & deprecation, mainly just don't want to cause unnecessary pain and breaks, especially going into the end of year.

tritium-registry/src/test/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistryTest.java
schlosna: thanks, may clean this up a bit more later, but not going to block merge

carterkozak:
:+1: Thanks @schlosna!

carterkozak:
This will need an approval from you as well because I pushed a commit.

schlosna (commented):
👍

schlosna:
I can't self approve as submitter and don't think policybot is happy with just a 👍 . I could force it through as admin, but would prefer to have @iamdanfox or someone else review the new API.

schlosna:
interesting, policybot and bulldozer may have joined together in the singularity

carterkozak:
bulldozer attempts to yolomerge (without admin override), once all the robots are happy the attempts become successful :-)

PR #502, opened 2019-11-18
Migrate several metrics to metric-schema

carterkozak:
==COMMIT_MSG==
Migrate several metrics to metric-schema
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
@@ -318,7 +318,7 @@ public final class MetricRegistries {
             String name) {
         return new TaggedMetricsScheduledExecutorService(
                 checkNotNull(delegate, "delegate"),
-                checkNotNull(registry, "registry"),
+                ExecutorMetrics.of(registry),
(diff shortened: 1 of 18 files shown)


carterkozak:
We're good to go

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
```diff
@@ -318,7 +318,7 @@ public static ScheduledExecutorService instrument(
             String name) {
         return new TaggedMetricsScheduledExecutorService(
                 checkNotNull(delegate, "delegate"),
-                checkNotNull(registry, "registry"),
+                ExecutorMetrics.of(registry),
```
ferozco, line 321: should we still check non-null?

ferozco (approved):

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
carterkozak: I didn't retain that null check because our generated utility does the same check internally. No strong preference either way.

PR #505, opened 2019-11-20
Remove some scheduled executor metrics

a-k-g:
## Before this PR
~Each scheduled executor reports a bunch of metrics that we don't actually use~
~- `executor.scheduled.once` meter~
~- `executor.scheduled.overrun` counter~
~- `executor.scheduled.repetitively` meter~
~- `executor.scheduled.percent_of_period` histogram~

Each scheduled executor reports the `executor.scheduled.percent_of_period` histogram that we don't actually use.

## After this PR
Scheduled executors no longer report this metric.

## Possible downsides?
Might break someone depending on this metric.


carterkozak:
:-1:

Pretty sure we do depend on these, I’ve certainly used them to debug outages.

-ck

iamdanfox:
@carterkozak was your usage in a regular datadog dashboard or somewhere else (like prometheus?)

We've got an up-to-date dump of DataDog dashboards yesterday: https://internal-papaya/workspace/data-integration/dataset/preview/ri.foundry.main.dataset.9263b89b-cd44-46e1-970b-0036f3d6700c/master so could have a look through to identify how many usages.

Also would be nice to put a dollar figure out how much this would save!

carterkozak:
100% Prometheus ;-)

iamdanfox:
@a-k-g was the plan to just drop these from DD in filter-config?

a-k-g:
@iamdanfox plan was to filter out as well as stop emitting it if no one is using it.

PR #508, opened 2019-11-21
Instrumentation wrapper classes have their own loggers

carterkozak:
No longer have to reach into package-private
ByteBuddyInstrumentationAdvice, which can
result in access exceptions.

==COMMIT_MSG==
Instrumentation wrapper classes have their own loggers
==COMMIT_MSG==

```
[main] WARN com.palantir.tritium.proxy.InstrumentedRunnable$5 - Failure occurred handling 'preInvocation' invocation on: com.google.common.util.concurrent.Runnables$1@5829e4f4
java.lang.RuntimeException
	at com.palantir.tritium.proxy.InstrumentedRunnable$5.run(Unknown Source)
	at org.assertj.core.api.ThrowableAssert.catchThrowable(ThrowableAssert.java:62)
	at org.assertj.core.api.AssertionsForClassTypes.catchThrowable(AssertionsForClassTypes.java:739)
	at org.assertj.core.api.AssertionsForClassTypes.assertThatCode(AssertionsForClassTypes.java:710)
```

PR #510, opened 2019-11-21
Tritium instrumentation handles null InvocationContexts 

carterkozak:
==COMMIT_MSG==
Tritium instrumentation handles null InvocationContexts.
A behavior change was introduced in #379 which resulted in handlers
which return a null InvocationContext being considered disabled,
which prevented the `onSuccess` and `onFailure` methods from
being invoked at all.
==COMMIT_MSG==

### DIFF
--- a/tritium-api/src/main/java/com/palantir/tritium/event/InvocationEventHandler.java
+++ b/tritium-api/src/main/java/com/palantir/tritium/event/InvocationEventHandler.java
@@ -48,6 +48,9 @@ public interface InvocationEventHandler<C extends InvocationContext> {
      * interface method takes no arguments. Arguments of primitive types are
      * wrapped in instances of the appropriate primitive wrapper class, such as
      * {@code java.lang.Integer} or {@code java.lang.Boolean}.
+     *
+     * @return the current invocation context. Null values are not recommended but
+     * are supported.
      */
     C preInvocation(@Nonnull Object instance, @Nonnull Method method, @Nonnull Object[] args);
 
--- a/tritium-api/src/main/java/com/palantir/tritium/event/InvocationEventHandler.java
+++ b/tritium-api/src/main/java/com/palantir/tritium/event/InvocationEventHandler.java
@@ -54,7 +54,7 @@ public interface InvocationEventHandler<C extends InvocationContext> {
     /**
      * Invoked with the result of the invocation when it is successful.
      *
-     * @param context the current invocation context.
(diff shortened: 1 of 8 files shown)


carterkozak:
Note that this PR includes #508

tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
```diff
@@ -32,6 +32,36 @@
 
     private static final Logger logger = LoggerFactory.getLogger(CompositeInvocationEventHandler.class);
 
+    // A sentinel value is used to differentiate null contexts returned by handlers from
+    // invocations on disabled handlers.
+    private static final InvocationContext DISABLED_HANDLER_SENTINEL = new InvocationContext() {
```
schlosna: nit: should this be a singleton enum shared with ByteBuddy instrumentation?

carterkozak: It can be if we're okay with making it public. It's an implementation detail which isn't necessary in every case, so I'd prefer if it wasn't possible to reference externally.

carterkozak: +1 for singleton enum

carterkozak:
I think I should update behavior slightly to differentiate when `preInvocation` throws as opposed to when the handler is disabled, as they're currently the same. If `preInvocation` throws, I think we want to return a null context.

schlosna (approved):
👍 LGTM, thanks @carterkozak

carterkozak:
@schlosna Thanks for the review, I've pushed one more commit to this PR based on my last comment. Feel free to add `merge when ready` after you've taken a look at it if you're still happy with the change.

tritium-api/src/main/java/com/palantir/tritium/event/InvocationEventHandler.java
```diff
@@ -54,7 +54,7 @@
     /**
```
schlosna, line 57: should we specify `@return` docs for `preInvocation`?

carterkozak: Good call. I don't want to recommend returning null from `preInvocation` since our logging all suggests it's symptomatic of an error, and we've never annotated the method as `@Nullable`

schlosna (approved):

PR #512, opened 2019-11-21
Migrate `jvm.attribute` metrics to metric-schema

carterkozak:
==COMMIT_MSG==
Include `jvm.attribute` metrics in metric-schema
==COMMIT_MSG==

These are a bit gnarly because we don't define the gauges ourselves, but I think its worthwhile to provide documentation in our schema.

### DIFF
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
@@ -16,16 +16,21 @@
 
 package com.palantir.tritium.metrics.jvm;
 
+import com.codahale.metrics.Gauge;
 import com.codahale.metrics.JvmAttributeGaugeSet;
+import com.codahale.metrics.Metric;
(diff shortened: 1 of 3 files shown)


tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -57,6 +63,22 @@ public static void register(TaggedMetricRegistry registry) {
...
     }
 
+    private static void registerAttributes(InternalJvmMetrics metrics) {
+        Map<String, Metric> jvmAttributes = new JvmAttributeGaugeSet().getMetrics();
+        metrics.attributeName(gauge(jvmAttributes, "name"));
+        metrics.attributeVendor(gauge(jvmAttributes, "vendor"));
+        metrics.attributeUptime(gauge(jvmAttributes, "uptime"));
+    }
+
+    private static Gauge<?> gauge(Map<String, Metric> metrics, String name) {
```
ferozco, line 73: nice

ferozco (approved):

schlosna (approved):

PR #516, opened 2019-11-26
jvm.attribute.uptime tags includes jvm vendor and version info

iamdanfox:
## Before this PR

I'm currently rolling out a new JRE, adoptopenjdk 11.0.5. I've got it installed on a couple of canary deployments, but I currently don't have a good way of looking at a deployment and establishing how long services have actually been _running_ using my new JVM. 

Servers will pick up my new JVM when they restart, but I don't have the ability to just go and restart everything. Even if I did, I'd still like to look some kind of graph to figure out how much confidence I have in a particular jvm version.

cc @CRogers 

## After this PR
==COMMIT_MSG==
The `jvm.attribute.uptime` metric now has several additional tags corresponding to the JVM system properties `java.specification.version`, `java.version`, `java.version.date`, `java.runtime.version`, `java.vendor.version`, `java.vm.vendor`, allowing observability of how long a particular JVM version has been up for.
==COMMIT_MSG==

Obviously when a service restarts its uptime goes back to zero, but I actually want to add together its previous uptime with its current uptime... I think I do this by using the `per_second` and then `cum_sum` to get the graph I want.

These system properties are not expected to change for the lifetime of a service, so while this PR adds lots of tags, it shouldn't actually add any more unique _series_ in datadog (therefore I think the cost is $0).

Future work: I'm intending to slurp this stuff back from DD -> our internal foundry stack so that excavator can make decisions based on it

## Possible downsides?

- writing more bytes to disk
- there is now inconsistency between metrics `jvm` namespace, as most are untagged but `attribute.uptime` is super tagged.

## Alternatives

Could I just get this from skylab?

Not easily - skylab currently gives services both JAVA_8_HOME and JAVA_11_HOME environment variables and then lets them decide which one they want to use when they start up.

### DIFF
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
@@ -68,12 +68,12 @@ public final class JvmMetrics {
         metrics.attributeName(gauge(jvmAttributes, "name"));
         metrics.attributeVendor(gauge(jvmAttributes, "vendor"));
         metrics.attributeUptime()
-                .javaSpecificationVersion(System.getProperty("java.specification.version", "null"))
-                .javaVersion(System.getProperty("java.version", "null"))
-                .javaVersionDate(System.getProperty("java.version.date", "null"))
-                .javaRuntimeVersion(System.getProperty("java.runtime.version", "null"))
-                .javaVendorVersion(System.getProperty("java.vendor.version", "null"))
-                .javaVmVendor(System.getProperty("java.vm.vendor", "null"))
+                .javaSpecificationVersion(System.getProperty("java.specification.version", "unknown"))
+                .javaVersion(System.getProperty("java.version", "unknown"))
+                .javaVersionDate(System.getProperty("java.version.date", "unknown"))
+                .javaRuntimeVersion(System.getProperty("java.runtime.version", "unknown"))
+                .javaVendorVersion(System.getProperty("java.vendor.version", "unknown"))
+                .javaVmVendor(System.getProperty("java.vm.vendor", "unknown"))
                 .build(gauge(jvmAttributes, "uptime"));
     }
 
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
@@ -65,8 +65,6 @@ public final class JvmMetrics {
 
     private static void registerAttributes(InternalJvmMetrics metrics) {
         Map<String, Metric> jvmAttributes = new JvmAttributeGaugeSet().getMetrics();
-        metrics.attributeName(gauge(jvmAttributes, "name"));
-        metrics.attributeVendor(gauge(jvmAttributes, "vendor"));
         metrics.attributeUptime()
                 .javaSpecificationVersion(System.getProperty("java.specification.version", "unknown"))
                 .javaVersion(System.getProperty("java.version", "unknown"))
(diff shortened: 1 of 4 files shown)


tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -67,7 +67,14 @@ private static void registerAttributes(InternalJvmMetrics metrics) {
         Map<String, Metric> jvmAttributes = new JvmAttributeGaugeSet().getMetrics();
         metrics.attributeName(gauge(jvmAttributes, "name"));
         metrics.attributeVendor(gauge(jvmAttributes, "vendor"));
-        metrics.attributeUptime(gauge(jvmAttributes, "uptime"));
+        metrics.attributeUptime()
+                .javaSpecificationVersion(System.getProperty("java.specification.version", "null"))
```
carterkozak: I think `unknown` has more meaning than `null` when we consume the results.
```suggestion
                .javaSpecificationVersion(System.getProperty("java.specification.version", "unknown"))
```

tritium-metrics-jvm/src/main/metrics/metrics.yml
```diff
@@ -37,3 +37,10 @@ namespaces:
       attribute.uptime:
         type: gauge
         docs: Provides the Java virtual machine uptime value in milliseconds.
+        tags:
+        - javaSpecificationVersion # 11
+        - javaVersion # 11.0.3
+        - javaVersionDate # 11.0.3
+        - javaRuntimeVersion # 11.0.3+7-LTS
+        - javaVendorVersion # Zulu11.31+11-CA
+        - javaVmVendor # Azul Systems, Inc.
```
carterkozak: Might as well list examples in the docs field so everyone gets the benefits.

iamdanfox: I found these by looking through `java -XshowSettings:all -version` on a few different java installations I have locally - these seemed the most interesting.

tritium-metrics-jvm/src/main/metrics/metrics.yml
```diff
@@ -37,3 +37,10 @@ namespaces:
       attribute.uptime:
```
carterkozak, line 31: In a follow up we should probably get rid of `attribute.name` and `attribute.vendor` since they're provided by our uptime gauge.

tritium-metrics-jvm/src/main/metrics/metrics.yml
iamdanfox: Sure thing - I wonder if it would actually make sense to allow documentation on tags themselves? currently it's just a `list<string>`

https://github.com/palantir/metric-schema/blob/develop/metric-schema-api/src/main/conjure/metric-schema-api.yml#L25

carterkozak: I had considered that but ultimately decided against it because I've found the conjure format with multiple value structures confusing. I think we need docs for tags infrequently enough that its worth the simplicity to include documentation in the metric docs field.
I may have gotten it wrong though ¯\_(ツ)_/¯ What do you think?

tritium-metrics-jvm/src/main/metrics/metrics.yml
iamdanfox: actually have just tacked it on here, as we're conceptually moving this info from a `Gauge<String>` (which isn't representable in DD) to tags on the numerical uptime gauge.

carterkozak: I like it. We should probably flag this as a break, despite even though it could only break manual workflows.

tritium-metrics-jvm/src/main/metrics/metrics.yml
iamdanfox: yeah I think it's probably infrequent enough that there's not really any point. If we weren't committed to a conjure def here I think we could probably do some kind of fun yaml abuse like:

```yml
tags:
- javaRuntimeVersion : docs for this tag
- javaVendorVrrsion
- etc
```

carterkozak:
lgtm, thanks @iamdanfox :+1:

PR #519, opened 2019-11-27
Deprecate SharedTaggedMetricRegistries#getSingleton

pkoenig10:
We'd like to encourage people not to use the global singleton. Marking the method as deprecated signals this intent to potential users.

It's relatively easy for callers to add `SuppressWarnings("deprecation")` if it is not currently feasible to migrate away from the global singleton.

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/SharedTaggedMetricRegistries.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/SharedTaggedMetricRegistries.java
@@ -37,8 +37,7 @@ public final class SharedTaggedMetricRegistries {
      * Single shared instance for use when it is infeasible to plumb through a user supplied instance.
      *
      * @deprecated avoid using the global singleton
-     *
-     * */
+     */
     @Deprecated
     public static TaggedMetricRegistry getSingleton() {
         return DEFAULT;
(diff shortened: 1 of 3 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/SharedTaggedMetricRegistries.java
```diff
@@ -33,7 +33,13 @@
      */
     private static final TaggedMetricRegistry DEFAULT = new SlidingWindowTaggedMetricRegistry(35, TimeUnit.SECONDS);
 
-    /** Single shared instance for use when it is infeasible to plumb through a user supplied instance. */
+    /**
+     * Single shared instance for use when it is infeasible to plumb through a user supplied instance.
+     *
+     * @deprecated avoid using the global singleton
+     *
```
schlosna:
```suggestion
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/SharedTaggedMetricRegistries.java
```diff
@@ -33,7 +33,13 @@
...
+     * */
```
schlosna:
```suggestion
     */
```

schlosna (approved):
Seems reasonable to me.

PR #553, opened 2019-12-16
Downgrade error_prone_annotations 2.3.4 -> 2.3.3

schlosna:
## Before this PR
tritium imposed error_prone_annotations 2.3.4 on consumers

## After this PR
==COMMIT_MSG==
tritium no longer imposes error_prone_annotations 2.3.4 on consumers,
to avoid triggering a baseline bug (https://github.com/palantir/gradle-baseline/issues/1113)
==COMMIT_MSG==


carterkozak:
:+1:

PR #555, opened 2019-12-17
RealMetricName provides a toString implementation

carterkozak:
## Before this PR
`RealMetricName@012abc`

## After this PR
==COMMIT_MSG==
RealMetricName provides a toString implementation matching ImmutableMetricName.
==COMMIT_MSG==

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/RealMetricName.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/RealMetricName.java
@@ -51,6 +51,11 @@ final class RealMetricName implements MetricName {
     }
 
     @Override
+    public String toString() {
+        return "MetricName{safeName=" + safeName + ", safeTags=" +  safeTags + '}';
+    }
+
+    @Override
     public int hashCode() {
         return hashCode;
     }
--- a/tritium-registry/src/test/java/com/palantir/tritium/metrics/registry/MetricNameTest.java
+++ b/tritium-registry/src/test/java/com/palantir/tritium/metrics/registry/MetricNameTest.java
@@ -39,23 +39,30 @@ public class MetricNameTest {
         MetricName three = RealMetricName.create(two);
         MetricName four = RealMetricName.create(
                 MetricName.builder()
-                .safeName("test")
-                .putSafeTags("key2", "value2")
-                .putSafeTags("key", "value")
+                        .safeName("test")
+                        .putSafeTags("key2", "value2")
+                        .putSafeTags("key", "value")
                         .build(),
                 "key1",
                 "value1");
+        MetricName five = ImmutableMetricName.copyOf(four);
 
-        assertThat(one).isEqualTo(two).isEqualTo(three).isEqualTo(four);
-        assertThat(two).isEqualTo(one).isEqualTo(three).isEqualTo(four);
-        assertThat(three).isEqualTo(one).isEqualTo(two).isEqualTo(four);
-        assertThat(four).isEqualTo(one).isEqualTo(two).isEqualTo(three);
+        assertThat(one).isEqualTo(two).isEqualTo(three).isEqualTo(four).isEqualTo(five);
+        assertThat(two).isEqualTo(one).isEqualTo(three).isEqualTo(four).isEqualTo(five);
+        assertThat(three).isEqualTo(one).isEqualTo(two).isEqualTo(four).isEqualTo(five);
+        assertThat(four).isEqualTo(one).isEqualTo(two).isEqualTo(three).isEqualTo(five);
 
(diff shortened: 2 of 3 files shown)


tritium-registry/src/test/java/com/palantir/tritium/metrics/registry/MetricNameTest.java
```diff
@@ -36,11 +36,13 @@ public void compareSame() {
...
+        MetricName three = RealMetricName.create(two);
 
-        assertThat(one).isEqualTo(two);
-        assertThat(two).isEqualTo(one);
+        assertThat(one).isEqualTo(two).isEqualTo(three);
+        assertThat(two).isEqualTo(one).isEqualTo(three);
+        assertThat(three).isEqualTo(one).isEqualTo(two);
 
-        assertThat(one.toString()).isEqualTo(two.toString());
+        assertThat(one.toString()).isEqualTo(two.toString()).isEqualTo(three.toString());
```
schlosna: minor nit: do we want to verify the actual contents of the `toString` (i.e. there's an unlikely alternate universe where `one == two == three` if we had a flyweight cache behind builder & RealMetricName)?

```suggestion
        assertThat(one.toString())
                .isEqualTo(two.toString())
                .isEqualTo(three.toString())
                .isEqualTo("MetricName{safeName=test, safeTags={key=value, key1=value1, key2=value2}}");
```

tritium-registry/src/test/java/com/palantir/tritium/metrics/registry/MetricNameTest.java
```diff
@@ -36,11 +36,13 @@ public void compareSame() {
                 .putSafeTags("key", "value")
                 .putSafeTags("key1", "value1")
                 .build();
+        MetricName three = RealMetricName.create(two);
 
-        assertThat(one).isEqualTo(two);
-        assertThat(two).isEqualTo(one);
+        assertThat(one).isEqualTo(two).isEqualTo(three);
+        assertThat(two).isEqualTo(one).isEqualTo(three);
+        assertThat(three).isEqualTo(one).isEqualTo(two);
```
schlosna: Does assertj have any equivalent to perform the full equals/hashCode comparison reflexive, symmetric, transitive permutations like `com.google.common.testing.EqualsTester` [`org.assertj.core.test.EqualsHashCodeContractAssert`](https://github.com/joel-costigliola/assertj-core/blob/master/src/test/java/org/assertj/core/test/EqualsHashCodeContractAssert.java) seems to be what we'd want, but not sure why its not in main assertj module.

Might also want to verify `hashCode` while we're at it.
```
        MetricName four = ImmutableMetricName.copyOf(three);

        assertThat(one)
                .hasSameHashCodeAs(two)
                .hasSameHashCodeAs(three)
                .hasSameHashCodeAs(four);
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/RealMetricName.java
```diff
@@ -50,6 +50,11 @@ public String safeName() {
         return safeTags;
     }
 
+    @Override
+    public String toString() {
+        return "MetricName{safeName=" + safeName + ", safeTags=" +  safeTags + '}';
```
schlosna, line 55: should we condense this, just make it JSON (like) for simplicity?

tritium-registry/src/test/java/com/palantir/tritium/metrics/registry/MetricNameTest.java
carterkozak: good call

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/RealMetricName.java
carterkozak: This matches ImmutableMetricName, I'd prefer to have RealMetricName match the other implementation for now, I'm not sure there's enough benefit to changing the structure to something json-like for it to be worth the change. Thoughts?

schlosna: as-is is fine by me

tritium-registry/src/test/java/com/palantir/tritium/metrics/registry/MetricNameTest.java
carterkozak: > Does assertj have any equivalent to perform the full equals/hashCode comparison reflexive, symmetric, transitive permutations

Not that I'm aware of, I can file an assertj FR upstream though.

> Might also want to verify hashCode while we're at it.

Good idea, done.

schlosna (approved):

PR #583, opened 2020-01-23
Upgrades gradle wrapper to 6.1

schlosna:
Fixes issues seen in #573 upgrading to Gradle 6.1 via gradle-baseline 2.49.2 upgrade to pick up https://github.com/palantir/gradle-baseline/pull/1176


changelog/@unreleased/pr-583.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: improvement
+improvement:
+  description: |-
+    Upgrades gradle wrapper to 6.1
+
+    Fixes issues seen in #573 upgrading to Gradle 6.1 via gradle-baseline 2.49.2 upgrade to pick up https://github.com/palantir/gradle-baseline/pull/1176
+  links:
+  - https://github.com/palantir/tritium/pull/583
```
iamdanfox, line 8: FWIW I would use the `no changelog` for these because consumers of tritium probably don't care (and this doesn't warrant a release IMO)

iamdanfox (approved):
LGTM, but you don't have to do this manually anymore @schlosna, excavator should get everyone onto this later today!

PR #589, opened 2020-01-28
Handshake debug logging describing the negotiated protocol and cipher

carterkozak:
## Before this PR
Only metrics are recorded when handshakes complete, but in some scenarios standard logging is more helpful.

## After this PR
==COMMIT_MSG==
Add TLS handshake debug logging describing the negotiated protocol and cipher
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/HandshakeInstrumentation.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/HandshakeInstrumentation.java
@@ -1,0 +1,48 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


tritium-metrics/src/main/java/com/palantir/tritium/metrics/HandshakeInstrumentation.java
```diff
@@ -0,0 +1,48 @@
...
+    static void record(TlsMetrics metrics, String contextName, String cipherSuite, String protocol) {
+        metrics.handshake()
+                .context(contextName)
+                .cipher(cipherSuite)
+                .protocol(protocol)
+                .build()
+                .mark();
+        if (log.isDebugEnabled()) {
+            log.debug(
+                    "SSL Handshake completed for context {}, cipher {}, protocol {}",
```
carterkozak:
```suggestion
                    "TLS Handshake completed for context {}, cipher {}, protocol {}",
```

schlosna (approved):

PR #656, opened 2020-03-03
Reduce jqwik dependency to jqwik-api

schlosna:
## Before this PR
jqwik testing dependency was pulling in more than necessary

## After this PR
==COMMIT_MSG==
Reduce jqwik dependency to jqwik-api
==COMMIT_MSG==


versions.props
```diff
@@ -13,6 +13,7 @@ com.squareup.okhttp3:okhttp = 4.0.0
 com.uber.nullaway:nullaway = 0.7.9
 io.undertow:undertow-core = 2.0.20.Final
 net.bytebuddy:byte-buddy = 1.10.8
+net.jqwik:* = 1.2.4
```
carterkozak, line 16: Should we add this to our managed artifacts to get automatic upgrades?

schlosna: yes, PR to update the robots submitted.

tritium-registry/build.gradle
```diff
@@ -20,7 +20,7 @@ dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter-params'
     testImplementation 'org.mockito:mockito-core'
     testImplementation 'org.mockito:mockito-junit-jupiter'
-    testImplementation 'net.jqwik:jqwik'
+    testImplementation 'net.jqwik:jqwik-api'
```
schlosna, line 23: we don't need the full library, so updated to just use api

schlosna:
will rebase after #657 merges

carterkozak (approved):

versions.props
schlosna: this is now just sorting all the versions

PR #672, opened 2020-03-12
Implement equals and hashcode on instrumented TLS components

carterkozak:
Some libraries pool connections with a check for SSLSocketFactory
equality. If we provide two connections from the same factory
wrapped with the same name in different places. While it's
safer to pool connections from a constant factory object, not
sharing a pool between client instances, we shouldn't
unexpectedly break this case.

## After this PR
==COMMIT_MSG==
Implement equals and hashcode on instrumented TLS components
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/InstrumentedSslContext.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/InstrumentedSslContext.java
@@ -18,6 +18,7 @@ package com.palantir.tritium.metrics;
 
 import java.security.KeyManagementException;
 import java.security.SecureRandom;
+import java.util.Objects;
 import javax.net.ssl.KeyManager;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLContextSpi;
@@ -42,6 +43,23 @@ final class InstrumentedSslContext extends SSLContext {
     @Override
     public String toString() {
         return "InstrumentedSSLContext{delegate=" + context + ", name=" + name + '}';
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        if (this == other) {
+            return true;
+        }
+        if (other == null || getClass() != other.getClass()) {
+            return false;
+        }
+        InstrumentedSslContext that = (InstrumentedSslContext) other;
+        return context.equals(that.context) && name.equals(that.name);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(context, name);
     }
 
     private static final class InstrumentedSslContextSpi extends SSLContextSpi {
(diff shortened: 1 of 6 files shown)


tritium-metrics/src/main/java/com/palantir/tritium/metrics/InstrumentedSslContext.java
```diff
@@ -44,6 +45,23 @@ public String toString() {
...
+        if (other == null || getClass() != other.getClass()) {
+            return false;
+        }
+        InstrumentedSslContext that = (InstrumentedSslContext) other;
+        return context.equals(that.context) && name.equals(that.name);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(context, name);
```
schlosna, line 62: not sure how much this gets called in the pooling and maybe premature optimization, but should we avoid the varargs allocation?

tritium-metrics/src/main/java/com/palantir/tritium/metrics/InstrumentedSslEngine.java
```diff
@@ -277,6 +278,23 @@ public String toString() {
         return getClass().getSimpleName() + "{name=" + name + ", delegate=" + engine + '}';
     }
 
+    @Override
+    public boolean equals(Object other) {
+        if (this == other) {
+            return true;
+        }
+        if (other instanceof InstrumentedSslEngine) {
```
schlosna, line 286: curious why instanceof vs `.getClass()` check used above (noting that `InstrumentedSslEngine` is not final while `InstrumentedSslContext` is final)

schlosna (commented):
seems reasonable, couple questions

tritium-metrics/src/main/java/com/palantir/tritium/metrics/InstrumentedSslContext.java
carterkozak: I'm not aware of anything that hashes these objects, but I can replace these with an allocation-free implementation.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/InstrumentedSslEngine.java
carterkozak: error-prone complained about getClass in a non-final equals method. I considered suppressing because this type is effectively sealed, but opted to avoid suppressions when this works as well.
If `other` uses the java9 implementation, the delegate engine equality check will fail, the type of our InstrumentedSslEngine depends directly on the fields available on the delegate instance.

schlosna: this should be fine, I just don't recall why we didn't make `InstrumentedSslEngine` final as well.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/InstrumentedSslContext.java
schlosna: I'm fine merging as-is for now and we can adjust later if necessary

schlosna (approved):

tritium-metrics/src/main/java/com/palantir/tritium/metrics/InstrumentedSslEngine.java
carterkozak: I had thought I'd documented this but apparently not! Added a comment explaining the design here: #673

PR #689, opened 2020-04-01
Warn registering Caffeine cache without recording stats

schlosna:
## Before this PR
If a caller of `CaffeineCacheStats.registerCache` had does not enable stats recording on their Caffeine cache via `Caffeine.newBuilder().recordStats()`, the resulting cache metrics are always 0, which can be surprising.

## After this PR
==COMMIT_MSG==
When registering a Caffeine cache that does not have stats recording
enabled, log a warning and increment a "cache.stats.disabled" counter to
notify consumers of that their caches will not produce expected metrics.

Unfortunately Guava Caches do not currently expose a mechanism to
determine this, so this only applies to Caffeine caches (which should
generally be preferred over Guava caches anyway, see
<https://github.com/palantir/gradle-baseline/pull/317>).
==COMMIT_MSG==

### DIFF
--- a/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
+++ b/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
@@ -18,19 +18,26 @@ package com.palantir.tritium.metrics.caffeine;
 
 import static com.palantir.logsafe.Preconditions.checkNotNull;
 
+import com.codahale.metrics.Counter;
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
 import com.github.benmanes.caffeine.cache.Cache;
 import com.google.common.collect.ImmutableMap;
 import com.palantir.logsafe.Safe;
+import com.palantir.logsafe.SafeArg;
 import com.palantir.tritium.metrics.InternalCacheMetrics;
 import com.palantir.tritium.metrics.MetricRegistries;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @SuppressWarnings("WeakerAccess") // public API
 public final class CaffeineCacheStats {
+
+    private static final Logger log = LoggerFactory.getLogger(CaffeineCacheStats.class);
+    private static final String STATS_DISABLED = "cache.stats.disabled";
 
     private CaffeineCacheStats() {}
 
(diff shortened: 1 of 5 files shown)


tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
```diff
@@ -62,8 +79,22 @@ public static void registerCache(TaggedMetricRegistry registry, Cache<?, ?> cach
...
+        } else {
+            warnNotRecordingStats(
+                    name,
+                    registry.counter(InternalCacheMetrics.taggedMetricName(name).apply(STATS_DISABLED)));
+        }
+    }
 
-        CaffeineCacheTaggedMetrics.create(cache, name).getMetrics().forEach(registry::registerWithReplacement);
+    private static void warnNotRecordingStats(@Safe String name, Counter counter) {
+        counter.inc();
```
carterkozak, line 92: nit: might prefer a meter because this value will only ever increase. Alternatively we could log the warning without adding a metric, though that's more difficult to search broadly.

chris05atm:
I wonder if this should just be a gradle-baseline check?

schlosna:
@carterkozak has written a lot more error-prone checks, so curious if he knows if there is a good way to validate that the constructed `Cache` passed to `registerCache` had `recordStats()` called during the cache build phase.

carterkozak:
bluf: Ideally we implement both

It's straightforward to validate assuming the builder is completely fluent -- if Cache.Builder instances are stored to variables it becomes more difficult to follow the flow, and more difficult still if the builder is passed into or returned from another method. Other edge cases to consider are caches we wrap that are created by other libraries, where compile-time validation can do relatively little.

I think having both would be a good idea, a simple error-prone rule could cover 80% of more of cases. Are we confident the minor performance hit is worthwhile across the board including caches that aren't instrumented? I can buy the argument that all caches should be instrumented, they're often the source of memory pressure. Automation to attach cache construction to a TaggedMetricRegistry instance is nontrivial.

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
carterkozak: I think we should fail tests in this case. Preconditions will be heavy handed and could potentially break existing services, an assert might be reasonable here to catch issues premerge without upgrade risk.

carterkozak:
I wonder if we should provide a StatsCounter implementation backed by non-gauge metrics and solve the problems that arise when multiple cache instances are created with the same name? Using the `Caffeine.newBuilder().recordStats(statsCounterSupplier)` API makes it impossible to register tritium instrumentation without enabling stats. It would require a change for both metric producers (new tritium api) and consumers (gauge metrics have different suffixes than other metric types).

schlosna:
@carterkozak seems reasonable to do both. Having a tagged metrics `StatsCounter` factory seems useful as well, though a big migration to shift over consumers. I don't think I have the cycles for the error-prone checks or stats counter and migration right now, so thoughts on just this to help in the interim?

FWIW there's a [standard dropwizard metric registry stats counter in Caffeine's examples: MetricsStatsCounter](https://github.com/ben-manes/caffeine/blob/master/examples/stats-metrics/src/main/java/com/github/benmanes/caffeine/examples/stats/metrics/MetricsStatsCounter.java)

carterkozak:
Absolutely, I think this provides a much better experience than silently reporting no data!

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
carterkozak: We don't have much prior art for asserts, and the warning/metric are as strict improvement. Happy merging this as is.

carterkozak:
👍

PR #690, opened 2020-04-02
Migrate class loading gauges to metric-schema

carterkozak:
==COMMIT_MSG==
Migrate class loading gauges to metric-schema
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
@@ -17,13 +17,13 @@
 package com.palantir.tritium.metrics.jvm;
 
 import com.codahale.metrics.jvm.BufferPoolMetricSet;
-import com.codahale.metrics.jvm.ClassLoadingGaugeSet;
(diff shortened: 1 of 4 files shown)


tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -17,13 +17,13 @@
 package com.palantir.tritium.metrics.jvm;
 
 import com.codahale.metrics.jvm.BufferPoolMetricSet;
-import com.codahale.metrics.jvm.ClassLoadingGaugeSet;
 import com.codahale.metrics.jvm.MemoryUsageGaugeSet;
 import com.codahale.metrics.jvm.ThreadStatesGaugeSet;
```
schlosna, line 21: Just curious, @carterkozak are you planning to replace all these gauge sets with metric-schema?

schlosna (approved):
👍

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
carterkozak: Yep, just slowly pushing through them between other things

PR #697, opened 2020-04-13
Migrate JVM thread gauges to metric-schema

carterkozak:
==COMMIT_MSG==
Migrate JVM thread gauges to metric-schema for more thorough generated documentation.
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
@@ -19,6 +19,7 @@ package com.palantir.tritium.metrics.jvm;
 import com.codahale.metrics.jvm.BufferPoolMetricSet;
 import com.codahale.metrics.jvm.MemoryUsageGaugeSet;
 import com.codahale.metrics.jvm.ThreadDeadlockDetector;
+import com.google.common.base.Suppliers;
 import com.google.common.collect.Maps;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.tritium.metrics.MetricRegistries;
@@ -28,6 +29,11 @@ import java.lang.management.ManagementFactory;
 import java.lang.management.RuntimeMXBean;
 import java.lang.management.ThreadInfo;
 import java.lang.management.ThreadMXBean;
+import java.util.EnumMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.function.BiFunction;
+import java.util.function.Supplier;
 
 /** {@link JvmMetrics} provides a standard set of metrics for debugging java services. */
 public final class JvmMetrics {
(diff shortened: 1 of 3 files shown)


tritium-metrics-jvm/src/main/metrics/metrics.yml
```diff
@@ -47,3 +47,30 @@ namespaces:
...
+        docs: Number of live threads in the `BLOCKED` state.
+      threads.waiting.count:
+        type: gauge
+        docs: Number of live threads in the `WAITING` state.
+      threads.timed-waiting.count:
+        type: gauge
+        docs: Number of live threads in the `TIMED_WAITING` state.
+      threads.terminated.count:
+        type: gauge
+        docs: Number of live threads in the `TERMINATED` state.
```
carterkozak, line 76: Ideally all of these except for the deadlock metric would be combined into a single tagged metric:

```
threads.count:
  type: gauge
  tags: [state, deamon]
```

I'm not sure the daemon count gives us much helpful information, so that could likely be removed as well, but I'd prefer not to make breaks coupled to this migration.

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -77,6 +79,37 @@ private static void registerClassLoading(InternalJvmMetrics metrics) {
...
+        metrics.threadsRunnableCount(() -> threadsByState(threads, Thread.State.RUNNABLE));
+        metrics.threadsBlockedCount(() -> threadsByState(threads, Thread.State.BLOCKED));
+        metrics.threadsWaitingCount(() -> threadsByState(threads, Thread.State.WAITING));
+        metrics.threadsTimedWaitingCount(() -> threadsByState(threads, Thread.State.TIMED_WAITING));
+        metrics.threadsTerminatedCount(() -> threadsByState(threads, Thread.State.TERMINATED));
+    }
+
+    private static int threadsByState(ThreadMXBean threads, Thread.State requestedState) {
+        // max-depth zero to avoid creating stack traces, we're only interested in high level metadata
+        final ThreadInfo[] loadedThreadInfo = threads.getThreadInfo(threads.getAllThreadIds(), 0);
```
schlosna: this might be premature optimization, but should we share the thread info across all thread state gauges?

carterkozak: I'd thought about this, but I'm not sure how helpful it would be in practice. We can memoize a supplier for the array of threads, but that makes it less likely the array will be quickly garbage collected, though I'm not confident escape analysis will be able to help us out much in this case.

There's no guarantee that all state gauges will be called in quick succession, so sharing the underlying thread info can be a bit complicated to do cleanly.

carterkozak: I suppose memoizing an enum map of state to count would work nicely

carterkozak: Updated.

schlosna (approved):

PR #698, opened 2020-04-13
Implement a tagged instrumented ThreadFactory

carterkozak:
This is based on the codahale InstrumentedThreadFactory, but
uses tagged metrics for reusable dashboarding and metric-schema
for auto-generated documentation.

## After this PR
==COMMIT_MSG==
Implement a tagged instrumented ThreadFactory
==COMMIT_MSG==

## Possible downsides?
metrics aren't free. But neither are compute resources, and metrics are cheaper.

### DIFF
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsThreadFactory.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsThreadFactory.java
@@ -1,0 +1,84 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.metrics;
+
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.Meter;
+import com.palantir.logsafe.Preconditions;
+import java.util.concurrent.ThreadFactory;
+
+/**
+ * An instrumented {@link ThreadFactory} based on the codahale {@link com.codahale.metrics.InstrumentedThreadFactory}.
+ */
+final class TaggedMetricsThreadFactory implements ThreadFactory {
+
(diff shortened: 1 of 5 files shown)


tritium-metrics/src/main/metrics/metrics.yml
```diff
@@ -86,3 +86,16 @@ namespaces:
...
+        tags: [executor]
+        docs: Rate that new threads are created for this executor.
+      threads.terminated:
+        type: meter
+        tags: [executor]
+        docs: Rate that executor threads are terminated.
+      threads.running:
+        type: counter
+        tags: [executor]
+        docs: Number of live threads created by this executor.
```
carterkozak, line 101: ThreadFactory metrics are in the executor section, which isn't necessarily accurate, but it's where we tend to use them. For metric consumers I think this might be easier to follow than separate components and `thread-factory` vs `executor` tag names, but I don't have a strong opinion.

Looking for feedback on this choice.

ferozco: The tagging might be a bit confusing for direct consumers but for those that get instrumentation for free through witchcraft I think this choice will make thing simpler. I expect the later to be the common case so I'm ok with it as is.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsThreadFactory.java
```diff
@@ -0,0 +1,84 @@
...
+    public Thread newThread(Runnable runnable) {
+        Thread result = delegate.newThread(new InstrumentedTask(
+                Preconditions.checkNotNull(runnable, "Runnable is required"), running, terminated));
+        created.mark();
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return "TaggedMetricsThreadFactory{delegate=" + delegate + '}';
```
ferozco, line 53: should we include the name in the toString?

carterkozak: I could go either way, we don't keep the name string around otherwise so I didn't include it here. I think the important part is to surface delegate.toString via the instrumented objects toString.

Happy to plumb it through if you think that would be helpful

ferozco: eh its probably fine as is

ferozco:
👍 Looks good! thanks @carterkozak

tritium-metrics/src/main/metrics/metrics.yml
carterkozak: I'm going to hold off merging for a bit to give @schlosna a chance to weigh in on this.

schlosna: I'm fine with using `executor` name in this context for consistency as I'd expect all use of instrumented thread factory via WC or similar executor factory.

schlosna (commented):
👍 LGTM

PR #712, opened 2020-05-04
Report metrics describing the finalizer queue size

carterkozak:
This recently took down a stack, but required full heap dumps to discover the root cause. Now we can keep an eye on (and once we have sufficient data, alert from) this value.
==COMMIT_MSG==
Report the finalizer queue size
==COMMIT_MSG==


tritium-metrics-jvm/src/test/java/com/palantir/tritium/metrics/jvm/JvmMetricsTest.java
```diff
@@ -71,7 +72,7 @@
             "jvm.memory.non-heap.max",
             "jvm.memory.heap.max",
             "jvm.safepoint.time",
-            "jvm.threads.timed_waiting.count",
+            "jvm.threads.timed-waiting.count",
```
carterkozak, line 75: This caused tests to fail locally, I'm not sure how it worked on develop.

schlosna: that's quiet odd

schlosna (approved):

iamdanfox:
Nice

PR #715, opened 2020-05-06
Deprecated non-tagged caffeine cache metrics

pkoenig10:
## Before this PR
The `CaffeineCacheStats#registerStats` method that takes a `MetricRegistry` is not deprecated.

## After this PR
The `CaffeineCacheStats#registerStats` method that takes a `MetricRegistry` is deprecated.

We want to encourage users to use tagged metrics because that allows using the same metrics for every cache.

### DIFF
--- a/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
+++ b/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
@@ -34,7 +34,6 @@ import java.util.function.Function;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-@SuppressWarnings("WeakerAccess") // public API
 public final class CaffeineCacheStats {
 
     private static final Logger log = LoggerFactory.getLogger(CaffeineCacheStats.class);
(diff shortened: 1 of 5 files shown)


tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
```diff
@@ -48,10 +47,13 @@ private CaffeineCacheStats() {}
...
      * {@code Caffeine.newBuilder().recordStats()} otherwise there are no cache metrics to register.
      *
+     * @deprecated use {@link #registerCache(TaggedMetricRegistry, Cache, String)}
+     *
      * @param registry metric registry
      * @param cache cache to instrument
      * @param name cache name
      */
+    @Deprecated
     public static void registerCache(MetricRegistry registry, Cache<?, ?> cache, String name) {
```
schlosna, line 57: We should probably deprecate the Guava cache version as well https://github.com/palantir/tritium/blob/aa6942273cce940270ddbc334f94c0e080dd3584/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java#L227

schlosna (commented):
I'm open to this to nudge and force folks toward `TaggedMetricRegistry`, but I'm a little worried about this blocking automated upgrades for projects enforce deprecation as compile error. @pkoenig10 do you or others have cycles to coordinate rollout through WC & all the downstream services?

pkoenig10:
I can help shepherd this upgrade and make sure the deprecation isn't blocking automatic upgrades.

> but I'm a little worried about this blocking automated upgrades for projects enforce deprecation as compile error

I'm generally not too worried about this. Projects that enforce deprecation as a compile error are/should be actively maintained. Projects in maintenance mode should not be enforcing deprecations as compile errors.

pkoenig10:
@schlosna any other concerns here?

schlosna (approved):

schlosna:
@pkoenig10 LGTM, will let you mark when ready to merge, release, and shepherd this through the ecosystem 👍

PR #732, opened 2020-05-26
Implement a simple metric-schema based handler

carterkozak:
Changes from the existing tagged metric invocation handler:
* Metrics use constant safe-names.
* There is no global failure metric, failures are all recorded
  using the standard service-name,method-name tuple.
* Failures are not tagged with the exception type in order to
  reduce cardinality. Other observability components exist to
  discover failure causes.
* Metrics are tagged with the provided service name, but no longer
  provide a tag including the declaring class as the two are generally
  either equivalent or interchangeable.

These changes are aimed at leveraging auto-generated documentation
from metric-schema, and allowing general purpose dashboards to take
advantage of tritium instrumentation metrics.

## Before this PR
No documentation for instrumentation proxy metrics through metric-schema. No reasonable way to build a generalized dashboard for instrumented proxies.

## After this PR
==COMMIT_MSG==
Implement a simple constant-safe-name metric-schema based metric invocation handler replacement for `TaggedMetricsServiceInvocationEventHandler`
==COMMIT_MSG==

## Possible downsides?

TODO: Build a reasonable migration path from existing tagged metrics
to those defined using metric-schema. Ideas?

### DIFF
--- a/tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
+++ b/tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
@@ -29,6 +29,7 @@ import com.palantir.tritium.event.InvocationEventHandler;
 import com.palantir.tritium.event.log.LoggingInvocationEventHandler;
 import com.palantir.tritium.event.log.LoggingLevel;
 import com.palantir.tritium.event.metrics.MetricsInvocationEventHandler;
+import com.palantir.tritium.event.metrics.TaggedMetricInvocationEventHandler;
 import com.palantir.tritium.event.metrics.TaggedMetricsServiceInvocationEventHandler;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.util.Collections;
@@ -146,10 +147,12 @@ public final class Instrumentation {
          * @param prefix - Metrics name prefix to be used
          * @return - InstrumentationBuilder
          */
+        @SuppressWarnings("deprecation") // TaggedMetricsServiceInvocationEventHandler is provided for backcompat
         public Builder<T, U> withTaggedMetrics(TaggedMetricRegistry metricRegistry, String prefix) {
             checkNotNull(metricRegistry, "metricRegistry");
             String serviceName = Strings.isNullOrEmpty(prefix) ? interfaceClass.getName() : prefix;
             this.handlers.add(new TaggedMetricsServiceInvocationEventHandler(metricRegistry, serviceName));
+            this.handlers.add(TaggedMetricInvocationEventHandler.of(metricRegistry, serviceName));
             return this;
         }
 
--- a/tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
+++ b/tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
@@ -144,16 +144,16 @@ public final class Instrumentation {
          * are chosen based off of the interface name and invoked method.
          *
          * @param metricRegistry - TaggedMetricsRegistry used for this application.
-         * @param prefix - Metrics name prefix to be used
-         * @return - InstrumentationBuilder
+         * @param serviceName - Service name with which to associate metrics.
+         * @return - Instrumentation builder
          */
         @SuppressWarnings("deprecation") // TaggedMetricsServiceInvocationEventHandler is provided for backcompat
-        public Builder<T, U> withTaggedMetrics(TaggedMetricRegistry metricRegistry, String prefix) {
+        public Builder<T, U> withTaggedMetrics(TaggedMetricRegistry metricRegistry, String serviceName) {
             checkNotNull(metricRegistry, "metricRegistry");
             this.handlers.add(new TaggedMetricsServiceInvocationEventHandler(
-                    metricRegistry, Strings.isNullOrEmpty(prefix) ? interfaceClass.getName() : prefix));
+                    metricRegistry, Strings.isNullOrEmpty(serviceName) ? interfaceClass.getName() : serviceName));
             this.handlers.add(TaggedMetricInvocationEventHandler.of(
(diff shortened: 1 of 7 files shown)


carterkozak:
woop, meant to open this as a draft. Curious if folks have ideas on migrating metrics.

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/TaggedMetricInvocationEventHandler.java
```diff
@@ -0,0 +1,96 @@
...
+ * described in {@code invocation-metrics.yml}.
+ */
+public final class TaggedMetricInvocationEventHandler extends AbstractInvocationEventHandler<InvocationContext> {
+
+    private final String serviceName;
+    private final ConcurrentMap<Method, Timer> timerCache = new ConcurrentHashMap<>();
+    private final Function<Method, Timer> onSuccessTimerMappingFunction;
+    private final InvocationMetrics metrics;
+
+    public static InvocationEventHandler<?> of(TaggedMetricRegistry taggedMetricRegistry, String serviceName) {
```
schlosna: do we need to document `serviceName` as `@Safe`?

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/TaggedMetricInvocationEventHandler.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.function.BooleanSupplier;
+import java.util.function.Function;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * An implementation of {@link AbstractInvocationEventHandler} whose purpose is to provide tagged metrics as
+ * described in {@code invocation-metrics.yml}.
+ */
+public final class TaggedMetricInvocationEventHandler extends AbstractInvocationEventHandler<InvocationContext> {
```
schlosna, line 42: I'm assuming you'd wire this in downstream in WC?

tritium-metrics/src/main/metrics/invocation-metrics.yml
```diff
@@ -0,0 +1,15 @@
+options:
+  javaPackage: com.palantir.tritium.event.metrics
+  javaVisibility: packagePrivate
+namespaces:
+  invocation:
+    docs: Tritium invocation proxy metrics for components wrapped using `Tritium.instrument` or `Instrumentation.builder()` with tagged metrics handlers.
+    metrics:
+      success:
+        type: timer
+        tags: [service-name, method-name]
```
schlosna, line 10: we'll have to eat cost migrating from `endpoint` to `method-name` until everything is migrated, and we'll need to kill off the old ones to avoid long tail cost of duplicate metrics

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/TaggedMetricInvocationEventHandler.java
carterkozak: This is the piece I haven't ironed out yet, long term I'd prefer if `Tritium.instrument` used this instead of the existing tagged metrics. As an intermediate step we could report both, but I'm not sure how much impact that would have on dd.

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/TaggedMetricInvocationEventHandler.java
carterkozak: Good idea

tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/TaggedMetricInvocationEventHandler.java
carterkozak: Updated to wire both for now.

tritium-metrics/src/main/metrics/invocation-metrics.yml
carterkozak: The most difficult part is discovering places that rely on the existing metrics. `service-name` and `endpoint` tags match those used by our per-endpoint server metrics, and the names are completely dynamic based on the interface name.
I'm a bit worried about keeping both tagged metric handlers around for too long because they're the heaviest handlers, and timers are expensive. I think an announcement email is the way to go once we've allowed a version to percolate out with the new metrics before we stop wiring the old metrics by default.

tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
```diff
@@ -143,13 +144,16 @@ private Builder(Class<T> interfaceClass, U delegate) {
...
-        public Builder<T, U> withTaggedMetrics(TaggedMetricRegistry metricRegistry, String prefix) {
+        @SuppressWarnings("deprecation") // TaggedMetricsServiceInvocationEventHandler is provided for backcompat
+        public Builder<T, U> withTaggedMetrics(TaggedMetricRegistry metricRegistry, String serviceName) {
             checkNotNull(metricRegistry, "metricRegistry");
-            String serviceName = Strings.isNullOrEmpty(prefix) ? interfaceClass.getName() : prefix;
-            this.handlers.add(new TaggedMetricsServiceInvocationEventHandler(metricRegistry, serviceName));
+            this.handlers.add(new TaggedMetricsServiceInvocationEventHandler(
+                    metricRegistry, Strings.isNullOrEmpty(serviceName) ? interfaceClass.getName() : serviceName));
+            this.handlers.add(TaggedMetricInvocationEventHandler.of(
+                    metricRegistry, Strings.isNullOrEmpty(serviceName) ? interfaceClass.getSimpleName() : serviceName));
```
carterkozak, line 156: Note: I've opted for the interface simple name here over the fqcn. Most UIs don't handle full class names very gracefully, and I don't think there's much risk of duplication especially when we filter by deployment and product first.

PR #754, opened 2020-06-16
TaggedMetricRegistry with baked-in libraryName and libraryVersion tags

iamdanfox:
## Before this PR

In Dialogue, I've found it incredibly beneficial to have the `dialogueVersion` tag on all my metrics, because when looking at a graph I can tell whether they're on latest or an older version. I've been using this `VersionedTaggedMetricRegistry` class to achieve this: https://github.com/palantir/dialogue/blob/develop/dialogue-core/src/main/java/com/palantir/dialogue/core/VersionedTaggedMetricRegistry.java. I had intended to generalize this and make the pattern available for other libraries e.g. atlasdb (cc @jkong) and witchcraft.

A more urgent reason to pursue this came up recently with regard to our DD $ spend, as our spend reports are currently broken down by `product` tag, which makes it hard to see what proportion of the $ spend our various _libraries_ are responsible for. Apparently we can get cost breakdowns made for _any_ given tag, there just isn't a consistent one across libraries yet. (cc @samrogerson)

## After this PR
==COMMIT_MSG==
Library authors can now construct a TaggedMetricRegistry that augments all created metrics with a `libraryName` and `libraryVersion` tag:
```java
MetricRegistries.wrapWithLibraryInfo()
                .registry(registry)
                .libraryName("my-library")
                .libraryVersionFromClass(MyLibrary.class)
                .build()
```
==COMMIT_MSG==

- this approach means that when we use `Tritium.instrument` to add metrics to our WC clients, we should be able to correctly attribute this to 'witchcraft' rather than tritium (which is less helpful)
- an alternative approach was to implement this in metric-schema, but @ferozco didn't really like that approach (and it also wouldn't catch _all_ atlasdb metrics).

## Possible downsides?

- actually _using_ this may be a little bit fiddly, because in WC we'll need to construct one of these and carefully pass around this `wcInternalTaggedMetricRegistry` without exposing it to users.

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AugmentedTaggedMetricRegistry.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AugmentedTaggedMetricRegistry.java
@@ -178,14 +178,14 @@ public final class AugmentedTaggedMetricRegistry implements TaggedMetricRegistry
     }
 
     @Override
-    public boolean equals(Object o) {
-        if (this == o) {
+    public boolean equals(Object obj) {
+        if (this == obj) {
             return true;
         }
-        if (o == null || getClass() != o.getClass()) {
+        if (obj == null || getClass() != obj.getClass()) {
             return false;
         }
-        AugmentedTaggedMetricRegistry that = (AugmentedTaggedMetricRegistry) o;
+        AugmentedTaggedMetricRegistry that = (AugmentedTaggedMetricRegistry) obj;
         return delegate.equals(that.delegate) && tagName.equals(that.tagName) && tagValue.equals(that.tagValue);
     }
 
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AugmentedTaggedMetricRegistry.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AugmentedTaggedMetricRegistry.java
@@ -47,7 +47,7 @@ public final class AugmentedTaggedMetricRegistry implements TaggedMetricRegistry
         this.tagValue = Preconditions.checkNotNull(tagValue, "tagValue");
     }
 
-    static AugmentedTaggedMetricRegistry create(
+    public static AugmentedTaggedMetricRegistry create(
             TaggedMetricRegistry delegate, @Safe String tagName, @Safe String tagValue) {
         if (delegate instanceof AugmentedTaggedMetricRegistry) {
             AugmentedTaggedMetricRegistry other = (AugmentedTaggedMetricRegistry) delegate;
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AugmentedTaggedMetricRegistry.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AugmentedTaggedMetricRegistry.java
@@ -24,6 +24,8 @@ import com.codahale.metrics.Metric;
 import com.codahale.metrics.Timer;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.Safe;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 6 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AugmentedTaggedMetricRegistry.java
```diff
@@ -0,0 +1,202 @@
...
+     * .
+     * @deprecated not implemented
+     */
+    @Deprecated
+    @Override
+    public void addMetrics(String _safeTagName, String _safeTagValue, TaggedMetricSet _metrics) {
+        throw new UnsupportedOperationException(
+                "Operations involving transforming TaggedMetricSet are not supported, please interact with the "
+                        + "delegate directly");
+    }
```
iamdanfox, line 156: i'm not 100% sure what the desired behaviour is in this case- perhaps we need an AugmentedTaggedMetricSet too? Passing it straight down unmodified feels like it might surprise users.

carterkozak:
Could we do something more automatic using metric-schema to explicitly add these tags based on the generated metrics package implementation version, and perhaps the repository name (or similar heuristic)?

I'd prefer if we could make this happen automatically without requiring code changes to opt into it, if we can make that happen.

iamdanfox:
Haha I pitched that approach in #team-foundry-infra this afternoon

carterkozak:
I thought we wanted to use that design instead of this one because it would allow us to include the extra tags in our docs, but I may have misinterpreted!

iamdanfox:
I don't really have a strong preference either way tbh - althought I guess one difference between the two approaches is for metrics like the [`tls.handshake`](https://github.com/palantir/tritium/blob/develop/tritium-metrics/src/main/metrics/metrics.yml#L49-L55), we might want to say it's always attributed to `tritium` or we might want to say the client side is attributed to `dialogue` and the server side is `witchcraft`. It's sort of a question of whether attribution should go to the library author or to the person who chose to use the library?

iamdanfox:
Another example of this is the [executor](https://github.com/palantir/tritium/blob/b561ad836f36a6a0d4bd8b1b030f064bfe285bb2/tritium-metrics/src/main/metrics/metrics.yml#L56) metrics, where I think it might be quite helpful to attribute these to atlasdb as they're the people who control how many threads they make... don't really want to penalize Dave for making such a widely relied-on library :)

carterkozak:
What happens when another library wraps dialogue and applies a lib/version?

iamdanfox:
Hmm at the moment it [just throws](https://github.com/palantir/tritium/pull/754/files#diff-9d8c1a2b321a81ccbc8c2ec1cc8da021R44-R52), but maybe we should actually switch the attribution to that library?

(I don't want to lose my version tag in that case though - maybe I'll change that to say `dialogueVersion`, `atlasdbVersion` etc?)

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AugmentedTaggedMetricRegistry.java
```diff
@@ -0,0 +1,202 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.BiConsumer;
+import java.util.function.Supplier;
+
+/**
+ * Any metric created on this registry will be saved into the delegate {@link TaggedMetricRegistry} with an extra tag
+ * added. This ensures any reads from the underlying delegate will have the desired 'augmented' tag.
+ */
+public final class AugmentedTaggedMetricRegistry implements TaggedMetricRegistry {
```
ferozco, line 39: what other purposes do you imagine for an `AugmentedTaggedMetricRegistry` besides adding library name/version? I wonder if we could just simplify to having a single   `ProvenanceTaggedMetricRegistry` (its a working title) which takes a name, version and a delegate. This would allow us to do away with a a bunch of boilerplate in the implementation and could make passing registries across libraries safer since you easily access and re-wrap the delegate of another `ProvenanceTaggedMetricRegistry`.

ferozco:
I think it is helpful to have the provenance live outside of metric schema since it gives you additional flexibility in attribution and it also does not require  converting all metrics to get complete attribution.

carterkozak:
Given the executor case, what library owns the executors created using the WC executor factory? Tritium for the instrumentation, wc as the library, my library that requests the executor from wc, or a service that uses my library?
I think matching the library that defines a metric (same as metric-schema documentation) is reasonable enough, and allows us to avoid adding complexity for other developers.
Let’s chat tomorrow :-)

carterkozak:
> does not require converting all metrics to get complete attribution.

But does require code change

iamdanfox:
Closing out as the plan is to go with the metric-schema implementation: https://github.com/palantir/metric-schema/pull/288

PR #785, opened 2020-07-11
Setup dependabot

schlosna:
## Before this PR
Excavator manages dependency upgrades

## After this PR

==COMMIT_MSG==
Dependabot and excavator manage dependency upgrades
==COMMIT_MSG==

## Possible downsides?

Dependably may not work with version.props. Dependabot may create conflicting/duplicate PRs compared to excavator.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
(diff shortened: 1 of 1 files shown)


carterkozak:
Will this work with gradle-consistent-versions/nebula style versions.props?

schlosna:
Honestly not sure if it will work with the version.props and locks. I don’t think there’s a way to test it until it merges, so I was just going to test it out.

carterkozak (approved):
Let's see what happens :+1:

schlosna:
So far it looks like dependabot does only hit dependencies with full `group:module:version` coordinates, but it did flag a couple interesting gradle plug-in deps in #786 and #787

PR #790, opened 2020-07-14
Add `instrument.tls.socket` property to disable handshake metrics using sockets

carterkozak:
Socket handshake listeners result in a `HandshakeCompleteNotify-Thread`
being created to run the callback rather than running on the same
thread.

Threads are expensive. So are handshakes, we shouldn't do either frequently, but it's possible the latter is occurring enough that the former compounds the problem.

==COMMIT_MSG==
Add `instrument.tls.socket` property to disable handshake metrics using sockets
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-790.v2.yml
+++ b/changelog/@unreleased/pr-790.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Add `instrument.tls.socket` property to disable handshake metrics using
+    sockets
+  links:
+  - https://github.com/palantir/tritium/pull/790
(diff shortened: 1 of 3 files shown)


carterkozak:
Note that this will not impact our server-side instrumentation which uses nio apis and SSLEngine, our dashboards tend to use the server-side metrics more than client reported metrics.

schlosna (approved):
The `HandshakeCompletedListenter` thread creation bottleneck seems to have existed for quite a while per https://mail.openjdk.java.net/pipermail/security-dev/2013-January/006390.html . I'm curious if we should consider reviving that thread after 7 years of dormancy as it is still there:

https://github.com/AdoptOpenJDK/openjdk-jdk14u/blob/4c6b476f2be142954d9823cc8a2e55a95d7433f4/src/java.base/share/classes/sun/security/ssl/TransportContext.java#L620

carterkozak:
once loom becomes available I imagine the fix will be relatively simple, but I don't want to plan around new jdk features being released...

PR #792, opened 2020-07-15
Socket handshake metrics are not recorded by default

carterkozak:
==COMMIT_MSG==
Socket handshake metrics are not recorded by default to prevent new threads for each handshake

Metrics and logging can be enabled by either enabling debug logging
on `com.palantir.tritium.metrics.HandshakeInstrumentation` or
by setting the `instrument.tls.socket` property to `true`.
==COMMIT_MSG==

## Possible downsides?

Reduced insight into how the system is behaving.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -116,7 +116,7 @@ allprojects {
 
     tasks.withType(Test) {
         useJUnitPlatform {
-            includeEngines 'jqwik'
+            includeEngines 'jqwik', 'junit-jupiter'
         }
         maxParallelForks Runtime.getRuntime().availableProcessors()
         testLogging {
--- a/changelog/@unreleased/pr-792.v2.yml
+++ b/changelog/@unreleased/pr-792.v2.yml
@@ -1,0 +1,10 @@
+type: improvement
+improvement:
+  description: |-
+    Socket handshake metrics are not recorded by default to prevent new threads for each handshake
+
+    Metrics and logging can be enabled by either enabling debug logging
+    on `com.palantir.tritium.metrics.HandshakeInstrumentation` or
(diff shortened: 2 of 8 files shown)


schlosna (approved):
Follow up to https://github.com/palantir/tritium/pull/790 . I think it is probably worth reviving the JDK thread about handshake event listener threads to see if we can improve upstream https://mail.openjdk.java.net/pipermail/security-dev/2013-January/006390.html

carterkozak:
PR doesn't appear to reflect commits I've pushed to fix tests that were in a silently broken state. -- trying to close+reopen

build.gradle
```diff
@@ -116,7 +116,7 @@ allprojects {
 
     tasks.withType(Test) {
         useJUnitPlatform {
-            includeEngines 'jqwik'
+            includeEngines 'jqwik', 'junit-jupiter'
```
carterkozak, line 119: This allows all our tests to run

iamdanfox:
Hang on does this mean we'll lose information about what cipher suites all our tls.handshakes are doing fleet-wide?

carterkozak:
> Hang on does this mean we'll lose information about what cipher suites all our tls.handshakes are doing fleet-wide?

only for clients, servers are not impacted because they don't use the handshake listener which spawns a thread per handshake. SSLEngines are still instrumented normally.

carterkozak:
reopening seemed to get things into the right state, this pr doesn't reflect my rebase... reopening again.

carterkozak:
well, it won't let me reopen again, so I've created a new PR. #793
GH is a little grumpy at the moment... https://www.githubstatus.com/ actions is degraded

PR #795, opened 2020-07-15
Fix build config to run junit tests

carterkozak:
Fixed several tests which weren't updated to allow library name
and version tags.

==COMMIT_MSG==
Fix build config to run junit tests
==COMMIT_MSG==


schlosna (approved):

carterkozak:
I've pushed commits to this branch that fix all the tests, but GH hasn't noticed them yet. I'll come back to this after GH is feeling better.

PR #851, opened 2020-09-15
Use error-prone 2.4.0

schlosna:
## Before this PR
`version.props` was unnecessarily pinning older error-prone 2.3.4 and disabling dependency bump excavator.

## After this PR

==COMMIT_MSG==
Use error-prone 2.4.0
==COMMIT_MSG==

## Possible downsides?
There should be no downside as https://github.com/palantir/tritium/pull/770 bumped to error-prone 2.4.0 in June, and versions.lock has had `com.google.errorprone:error_prone_annotations:2.4.0` since then.


carterkozak (approved):
Lgtm

PR #860, opened 2020-09-23
Implement a lock-free exponential decaying reservoir

carterkozak:
This solves both the nanotime issue reported in
https://github.com/dropwizard/metrics/pull/1654 as well
as correctness issues when the system clock changes in the original
implementation as this one only uses the precise clock.

More importantly, unexpected errors (OME, ThreadDeath, etc) cannot leave
this reservoir in a blocked state because it has no locks. If a thread
is descheduled while rescaling, another thread may complete the task
first.

This Reservoir allows us to specify a rescale threshold rather than
the hard coded 1-hour threshold.

## After this PR

==COMMIT_MSG==
Implement a lock-free exponential decaying reservoir
==COMMIT_MSG==

## Possible downsides?
More code to maintain, could have bugs, but it can't block entire applications!

## Benchmarks

```
Benchmark                                 (reservoirType)  Mode  Cnt     Score      Error  Units
ReservoirBenchmarks.updateReservoir            EXPO_DECAY  avgt    5  8235.861 ± 1306.404  ns/op
ReservoirBenchmarks.updateReservoir  LOCK_FREE_EXPO_DECAY  avgt    5   758.315 ±   36.916  ns/op
```

### DIFF
--- a/tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/ReservoirBenchmarks.java
+++ b/tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/ReservoirBenchmarks.java
@@ -56,7 +56,7 @@ public class ReservoirBenchmarks {
         LOCK_FREE_EXPO_DECAY() {
             @Override
             Reservoir create() {
-                return new LockFreeExponentiallyDecayingReservoir();
+                return LockFreeExponentiallyDecayingReservoir.builder().build();
             }
         };
 
--- a/tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/ReservoirBenchmarks.java
+++ b/tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/ReservoirBenchmarks.java
@@ -1,0 +1,77 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.microbenchmarks;
+
+import com.codahale.metrics.ExponentiallyDecayingReservoir;
+import com.codahale.metrics.Reservoir;
+import com.palantir.tritium.metrics.registry.LockFreeExponentiallyDecayingReservoir;
+import java.util.concurrent.TimeUnit;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Fork(1)
+@Threads(32)
+@State(Scope.Benchmark)
+@SuppressWarnings({"designforextension", "NullAway"})
+public class ReservoirBenchmarks {
+
+    @Param({"EXPO_DECAY", "LOCK_FREE_EXPO_DECAY"})
+    private ReservoirType reservoirType;
+
+    public enum ReservoirType {
+        EXPO_DECAY() {
+            @Override
+            Reservoir create() {
+                return new ExponentiallyDecayingReservoir();
+            }
+        },
+        LOCK_FREE_EXPO_DECAY() {
+            @Override
+            Reservoir create() {
+                return new LockFreeExponentiallyDecayingReservoir();
+            }
+        };
+
+        abstract Reservoir create();
+    }
+
+    private Reservoir reservoir;
+
+    @Setup
+    public void before() {
+        reservoir = reservoirType.create();
+    }
+
+    @Benchmark
+    public void updateReservoir() {
+        reservoir.update(1L);
+    }
+}
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
@@ -16,14 +16,13 @@
 
 package com.palantir.tritium.metrics.registry;
 
-import com.codahale.metrics.ExponentiallyDecayingReservoir;
 import com.google.auto.service.AutoService;
 
 @AutoService(TaggedMetricRegistry.class)
 public final class DefaultTaggedMetricRegistry extends AbstractTaggedMetricRegistry {
 
     public DefaultTaggedMetricRegistry() {
-        super(ExponentiallyDecayingReservoir::new);
+        super(() -> LockFreeExponentiallyDecayingReservoir.builder().build());
     }
 
     /**
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
@@ -22,9 +22,9 @@ import com.codahale.metrics.Snapshot;
 import com.codahale.metrics.WeightedSnapshot;
 import com.codahale.metrics.WeightedSnapshot.WeightedSample;
 import com.google.common.annotations.Beta;
+import java.time.Duration;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.ThreadLocalRandom;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
@@ -41,7 +41,7 @@ import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 public final class LockFreeExponentiallyDecayingReservoir implements Reservoir {
     private static final int DEFAULT_SIZE = 1028;
     private static final double DEFAULT_ALPHA = 0.015;
-    private static final long DEFAULT_RESCALE_THRESHOLD = TimeUnit.HOURS.toNanos(1);
+    private static final Duration DEFAULT_RESCALE_THRESHOLD = Duration.ofHours(1);
 
     private static final AtomicReferenceFieldUpdater<LockFreeExponentiallyDecayingReservoir, State> stateUpdater =
             AtomicReferenceFieldUpdater.newUpdater(LockFreeExponentiallyDecayingReservoir.class, State.class, "state");
@@ -109,15 +109,15 @@ public final class LockFreeExponentiallyDecayingReservoir implements Reservoir {
         this(DEFAULT_SIZE, DEFAULT_ALPHA, DEFAULT_RESCALE_THRESHOLD);
     }
 
-    public LockFreeExponentiallyDecayingReservoir(int size, double alpha, long rescaleThresholdNanos) {
-        this(size, alpha, rescaleThresholdNanos, Clock.defaultClock());
+    public LockFreeExponentiallyDecayingReservoir(int size, double alpha, Duration rescaleThreshold) {
+        this(size, alpha, rescaleThreshold, Clock.defaultClock());
     }
 
-    public LockFreeExponentiallyDecayingReservoir(int size, double alpha, long rescaleThresholdNanos, Clock clock) {
+    public LockFreeExponentiallyDecayingReservoir(int size, double alpha, Duration rescaleThreshold, Clock clock) {
         this.alpha = alpha;
         this.size = size;
         this.clock = clock;
-        this.rescaleThresholdNanos = rescaleThresholdNanos;
+        this.rescaleThresholdNanos = rescaleThreshold.toNanos();
         this.state = new State(clock.getTick(), new AtomicLong(), new ConcurrentSkipListMap<>());
     }
 
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
@@ -51,7 +51,6 @@ public final class LockFreeExponentiallyDecayingReservoir implements Reservoir {
     private final long rescaleThresholdNanos;
     private final Clock clock;
 
-    // state
     private volatile State state;
 
     private final class State {
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
@@ -122,10 +122,6 @@ public final class LockFreeExponentiallyDecayingReservoir implements Reservoir {
         rescaleIfNeeded(now).update(value, now);
     }
 
-    private State rescaleIfNeeded() {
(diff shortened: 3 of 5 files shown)


tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/ReservoirBenchmarks.java
```diff
@@ -0,0 +1,77 @@
...
+    private Reservoir reservoir;
+
+    @Setup
+    public void before() {
+        reservoir = reservoirType.create();
+    }
+
+    @Benchmark
+    public void updateReservoir() {
+        reservoir.update(1L);
```
blacksmithgu, line 75: Is it worth changing the rescale time to test rescale performance as well?

blacksmithgu (approved):
I went ahead and compared the implementation against the reference (`ExponentiallyDecayingReservoir`), and it looks good. Given the long rescale window (1 hour), the potential race condition between concurrent updates during a rescale seems quite tame.

tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/ReservoirBenchmarks.java
carterkozak: It's difficult to compare to the upstream version because they do not allow the interval to be configured. I think it's infrequent enough that we shouldn't need to worry about it.

iamdanfox: +1 for benchmarks :)

The upstream dropwizard-metrics implementation allows passing in a clock - could you inject a mutable clock which you increment by a second or so after every update? 

```
public ExponentiallyDecayingReservoir(int size, double alpha, Clock clock) {
```

or perhaps you'd even put a for loop in this `@Benchmark` method to do exactly 3600 updates and guarantee you get a rescale on every invocation?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -0,0 +1,166 @@
...
+        State(long startTick, AtomicLong count, ConcurrentSkipListMap<Double, WeightedSample> values) {
+            this.startTick = startTick;
+            this.values = values;
+            this.count = count;
+        }
+
+        private void update(long value, long timestampNanos) {
+            final double itemWeight = weight(timestampNanos - startTick);
+            final WeightedSample sample = new WeightedSample(value, itemWeight);
+            final double priority = itemWeight / ThreadLocalRandom.current().nextDouble();
```
iamdanfox: Ah yes this was the bit of randomness that I wasn't able to inject in dialogue :) I see in dw-metrics they actually used some `ThreadLocalRandomProxy` thingy, but presumably that's just to support jdks <7 before ThreadLocalRandom appeared??

tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/ReservoirBenchmarks.java
carterkozak: We can increment a clock, but things get a little wonky because we're running on several threads simultaneously. I think it boils down to what data we want to distill from the benchmark. The worst case scenario for this implementation is when all N threads concurrently do the work to rescale rather than parking while a single thread holding a lock does the work -- in the single-thread case the rescale itself is equivalent.

Relatedly, if we're worried about threads which produce metric running into issues, we could update this implementation such that it only rescales on `getSnapshot()`, but I'm not confident it's a good idea as long as the rescale interval is somewhat configurable.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
carterkozak: That's correct.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -0,0 +1,166 @@
...
+
+    public LockFreeExponentiallyDecayingReservoir() {
+        this(DEFAULT_SIZE, DEFAULT_ALPHA, DEFAULT_RESCALE_THRESHOLD);
+    }
+
+    public LockFreeExponentiallyDecayingReservoir(int size, double alpha, long rescaleThresholdNanos) {
+        this(size, alpha, rescaleThresholdNanos, Clock.defaultClock());
+    }
+
+    public LockFreeExponentiallyDecayingReservoir(int size, double alpha, long rescaleThresholdNanos, Clock clock) {
```
carterkozak:
```suggestion
    public LockFreeExponentiallyDecayingReservoir(int size, double alpha, Duration rescaleThreshold, Clock clock) {
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -0,0 +1,166 @@
...
+            this.count = count;
+        }
+
+        private void update(long value, long timestampNanos) {
+            final double itemWeight = weight(timestampNanos - startTick);
+            final WeightedSample sample = new WeightedSample(value, itemWeight);
+            final double priority = itemWeight / ThreadLocalRandom.current().nextDouble();
+
+            final long newCount = count.incrementAndGet();
+            if (newCount <= size || values.isEmpty()) {
```
iamdanfox, line 84: ok so the dw implementation gets away without this `values.isEmpty()` check... ~is it just because when the `values` ConcurrentSkipListMap is truly empty, then the `size` variable will never be zero? So values.firstKey() down below will always return non-null?~

nvm I don't think I fully understand how the existing dw implementation avoids NPEs, cos `size` is final!

carterkozak: the dw implementation uses exactly this check :-)

https://github.com/dropwizard/metrics/blob/0313a104bf785e87d7d14a18a82026225304c402/metrics-core/src/main/java/com/codahale/metrics/ExponentiallyDecayingReservoir.java#L102-L112

iamdanfox: Aha I had been looking at the last 3.x released version (3.2.6) https://github.com/dropwizard/metrics/blob/v3.2.6/metrics-core/src/main/java/com/codahale/metrics/ExponentiallyDecayingReservoir.java#L101

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -0,0 +1,166 @@
...
+            ConcurrentSkipListMap<Double, WeightedSample> newValues = new ConcurrentSkipListMap<>();
+            if (Double.compare(scalingFactor, 0) != 0) {
+                values.forEach((key, sample) -> {
+                    double newWeight = sample.weight * scalingFactor;
+                    if (Double.compare(newWeight, 0) == 0) {
+                        return;
+                    }
+                    WeightedSample newSample = new WeightedSample(sample.value, newWeight);
+                    if (newValues.put(key * scalingFactor, newSample) == null) {
+                        newCount.incrementAndGet();
```
iamdanfox: why increment the `newCount` AtomicLong on every forEach lambda call rather than just set it once at the end to `newValues.size()` like they did? https://github.com/dropwizard/metrics/blob/release/4.1.x/metrics-core/src/main/java/com/codahale/metrics/ExponentiallyDecayingReservoir.java#L184-L185

carterkozak: Size isn't trivial, and we're already iterating through elements.

carterkozak:
Updated to use Duration and a builder.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
iamdanfox: boils down to a longAdder sum? 

I guess I don't know of the top of my head whether a longadder sum vs incrementing a non-atomic singleton long array vs incrementing an AtomicLong is quicker 🤷

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -0,0 +1,197 @@
...
+
+    @Override
+    public void update(long value) {
+        long now = clock.getTick();
+        rescaleIfNeeded(now).update(value, now);
+    }
+
+    private State rescaleIfNeeded() {
+        return rescaleIfNeeded(clock.getTick());
+    }
```
iamdanfox: Now only used once, inline?

tritium-registry/src/test/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoirTest.java
```diff
@@ -0,0 +1,450 @@
...
+import com.codahale.metrics.Timer.Context;
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.jupiter.api.Test;
+
+/**
+ * {@link LockFreeExponentiallyDecayingReservoirTest} is based closely on the codahale
+ * <a href="https://github.com/dropwizard/metrics/blob/a314e903d7ba3c5ec47ee7c9103d73ed09a20bd6/metrics-core/src/test/java/com/codahale/metrics/ExponentiallyDecayingReservoirTest.java">
```
iamdanfox, line 34: :+1: for the links - I guess the alternative approach is to use the codahale version as a 'golden master' reference impl, like we did for the dialogue futures https://github.com/palantir/dialogue/blob/develop/dialogue-futures/src/test/java/com/palantir/dialogue/futures/DialogueFuturesTest.java, but I think I'm OK with this :)

iamdanfox (approved):
I'd think I'd want to make sure @schlosna is happy with this before replacing the default reservoir for our taggedmetricregistries, but in generally pretty pumped about chopping out some locks from such a common codepath as updating metrics, especially as they've given up on biasedlocking in java15!

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -0,0 +1,193 @@
...
+import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+
+/**
+ * {@link LockFreeExponentiallyDecayingReservoir} is based closely on the codahale
+ * <a href="https://github.com/dropwizard/metrics/blob/0313a104bf785e87d7d14a18a82026225304c402/metrics-core/src/main/java/com/codahale/metrics/ExponentiallyDecayingReservoir.java">
+ * ExponentiallyDecayingReservoir.java</a>, however it provides looser guarantees (updates which race rescale may be
+ * lost) while completely avoiding locks.
```
schlosna: I think we probably want to elaborate on the tradeoffs here about possible lost writes, especially if we increase the rescale frequency.

schlosna: If I'm reading correctly, in the case of racing rescales, the winner would have its new state installed and the losers would perform a volatile read of the latest state (which under extremely small rescale threshold may actually be newer than the former winner) and return that state, which is then updated. We might want to consider enforcing a lower bound for a reasonable rescale threshold may be specified (say 1 second which even seems low?).

carterkozak: Your analysis is correct. I’m not sure I’d want to strictly enforce a minimum interval because there are so many variables, perhaps stronger documentation describing the potential pitfalls would be sufficient?

I think losing a cas and updating a newer-than-expected state object is roughly equivalent to the calling thread being descheduled for a moment as it calls into reservoir.update.

j-baker:
What's the perf like for the expected more common uncontended case?

carterkozak:
> What's the perf like for the expected more common uncontended case?

Improves from 80ns/op to 40ns/op, though at that level I don't expect any real world impact in our consumers. I can rerun benchmarks with 1 thread later, or you can try it out locally if you like.

carterkozak:
1 thread:

```
Benchmark                                 (reservoirType)  Mode  Cnt   Score    Error  Units
ReservoirBenchmarks.updateReservoir            EXPO_DECAY  avgt    5  92.845 ± 36.478  ns/op
ReservoirBenchmarks.updateReservoir  LOCK_FREE_EXPO_DECAY  avgt    5  49.168 ±  1.033  ns/op
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -0,0 +1,218 @@
...
+                    }
+                }
+            }
+        }
+
+        State rescale(long newTick) {
+            long durationNanos = newTick - startTick;
+            double durationSeconds = durationNanos / 1_000_000_000D;
+            double scalingFactor = Math.exp(-alpha * durationSeconds);
+            AtomicLong newCount = new AtomicLong();
```
schlosna:
```suggestion
            long newCount = 0;
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
schlosna:
```suggestion
                        newCount++;
```

more a micro perf nit, but `newCount` could just be a `long` as we don't need the volatile/JMM side effects in this single thread until we create the new state, e.g. something like follows (and related suggested changes):
```
            long newCount = 0;
            ConcurrentSkipListMap<Double, WeightedSample> newValues = new ConcurrentSkipListMap<>();
            if (Double.compare(scalingFactor, 0) != 0) {
                values.forEach((key, sample) -> {
                    double newWeight = sample.weight * scalingFactor;
                    if (Double.compare(newWeight, 0) == 0) {
                        return;
                    }
                    WeightedSample newSample = new WeightedSample(sample.value, newWeight);
                    if (newValues.put(key * scalingFactor, newSample) == null) {
                        newCount++;
                    }
                });
            }
            return new State(newTick, new AtomicLong(newCount), newValues);
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -0,0 +1,218 @@
...
+                    if (Double.compare(newWeight, 0) == 0) {
+                        return;
+                    }
+                    WeightedSample newSample = new WeightedSample(sample.value, newWeight);
+                    if (newValues.put(key * scalingFactor, newSample) == null) {
+                        newCount.incrementAndGet();
+                    }
+                });
+            }
+            return new State(newTick, newCount, newValues);
```
schlosna, line 110:
```suggestion
            return new State(newTick, new AtomicLong(newCount), newValues);
```

carterkozak:
Merging to develop and cutting a release candidate. If things don't look good, we can revert the change to DefaultTaggedMetricRegistry before cutting a full release.

schlosna (approved):
overall LGTM, micro perf nit around the rescale count comments (could be follow up PR)

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java
```diff
@@ -16,14 +16,13 @@
...
 
-import com.codahale.metrics.ExponentiallyDecayingReservoir;
 import com.google.auto.service.AutoService;
 
 @AutoService(TaggedMetricRegistry.class)
 public final class DefaultTaggedMetricRegistry extends AbstractTaggedMetricRegistry {
 
     public DefaultTaggedMetricRegistry() {
-        super(ExponentiallyDecayingReservoir::new);
+        super(() -> LockFreeExponentiallyDecayingReservoir.builder().build());
```
schlosna, line 25: chatted with @carterkozak , I'm ok with changing this, cut an RC to bump a couple canaries and verify on realistic workloads

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
carterkozak: We'd either have to allocate a final container for the non-volatile value, or replace forEach with an iterator. What do you think?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
carterkozak: Updated using a concrete class instead of a lambda in forEach to produce this result.

PR #862, opened 2020-09-29
Enable JDK 11 and 15 test runs

schlosna:
## Before this PR
Generated circle config includes end-of-life JDK 14 unit tests, but doesn't run tests on JDK 11 or 15.

## After this PR

==COMMIT_MSG==
Enable JDK 11 and 15 test runs
==COMMIT_MSG==

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,6 +96,57 @@ jobs:
       - store_test_results: { path: ~/junit }
       - store_artifacts: { path: ~/artifacts }
 
+  unit-test-11:
+    docker: [{ image: 'circleci/openjdk:11-node' }]
+    resource_class: large
+    environment:
+      CIRCLE_TEST_REPORTS: /home/circleci/junit
+      CIRCLE_ARTIFACTS: /home/circleci/artifacts
+      GRADLE_OPTS: -Dorg.gradle.jvmargs='-XX:MaxMetaspaceSize=256m' -Dorg.gradle.workers.max=4
+      _JAVA_OPTIONS: -XX:ActiveProcessorCount=4 -Xmx1177m -XX:MaxMetaspaceSize=512m  -XX:ErrorFile=/home/circleci/artifacts/hs_err_pid%p.log -XX:HeapDumpPath=/home/circleci/artifacts
+    steps:
+      - checkout
+      - restore_cache: { key: 'gradle-wrapper-v2-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}' }
+      - restore_cache: { key: 'unit-test-11-gradle-cache-v2-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}' }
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -196,5 +253,5 @@ workflows:
           filters: { branches: { ignore: develop } }
 
       - publish:
-          requires: [ unit-test, unit-test-14, check, trial-publish ]
+          requires: [ unit-test, unit-test-11, unit-test-15, unit-test-14, check, trial-publish ]
```
iamdanfox, line 256: I just re-rendered this by running `REPOSITORY_OWNER=palantir REPOSITORY_NAME=tritium /Volumes/git/circle-templates/excavator/circleci/render.sh oss`... might actually just flip the `unit-test-14` thing from being default on to unit-test-15 being default on

.circleci/config.yml
```diff
@@ -96,6 +96,57 @@ jobs:
...
+          key: 'unit-test-11-gradle-cache-v2-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}'
+          paths: [ ~/.gradle/caches ]
+      - run:
+          command: mkdir -p ~/junit && find . -type f -regex ".*/build/.*TEST.*xml" -exec cp --parents {} ~/junit/ \;
+          when: always
+      - store_test_results: { path: ~/junit }
+      - store_artifacts: { path: ~/artifacts }
+
+  unit-test-15:
+    docker: [{ image: 'circleci/openjdk:11-node' }]
```
iamdanfox, line 122: maybe we should actually swap this out for a pre-baked jdk... looks like there's `openjdk:15-buster-node` available https://circleci.com/docs/2.0/docker-image-tags.json

schlosna: @iamdanfox I assume you mean update the template for `unit-test-15` to use `circleci/openjdk:15-buster-node` which [builds off `circleci/openjdk:15-jdk-buster`](https://github.com/CircleCI-Public/circleci-dockerfiles/blob/master/openjdk/images/15-jdk-buster/Dockerfile) which [builds off `openjdk:15-jdk-buster`](https://github.com/docker-library/openjdk/blob/master/15/jdk/buster/Dockerfile) which uses the OpenJDK builds from Oracle.

In general, I think we want to be using the Zulu OpenJDK builds for these projects, so might be worthwhile us contributing some Zulu OpenJDK 15 Dockerfiles to https://github.com/zulu-openjdk/zulu-openjdk to spin up some new docker images.

.circleci/config.yml
schlosna: yeah, I would prefer to drop JDK 14 when we can given it is end of life

.circleci/config.yml
schlosna: For now, I'm inclined to keep building & testing with Zulu OpenJDK 15

carterkozak (approved):
lgtm

PR #865, opened 2020-10-05
Update the LockFreeExponentiallyDecayingReservoir based on upstream feedback

carterkozak:
Before:
```
ReservoirBenchmarks.updateReservoir  LOCK_FREE_EXPO_DECAY  avgt    5   758.315 ±   36.916  ns/op
```

After:
```
ReservoirBenchmarks.updateReservoir  LOCK_FREE_EXPO_DECAY  avgt    5    72.031 ±    2.393  ns/op
```

This improvement is a consequence of avoiding CAS operations on the
counter in the fast path, when sample priority is too low to track.

==COMMIT_MSG==
Update the LockFreeExponentiallyDecayingReservoir based on upstream feedback
==COMMIT_MSG==

Upstream PR: https://github.com/dropwizard/metrics/pull/1656

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
@@ -57,7 +57,7 @@ import java.util.function.BiConsumer;
 @Beta
 public final class LockFreeExponentiallyDecayingReservoir implements Reservoir {
 
-    private static final double SECONDS_PER_NANO = .000000001D;
+    private static final double SECONDS_PER_NANO = .000_000_001D;
     private static final AtomicReferenceFieldUpdater<LockFreeExponentiallyDecayingReservoir, State> stateUpdater =
             AtomicReferenceFieldUpdater.newUpdater(LockFreeExponentiallyDecayingReservoir.class, State.class, "state");
 
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
@@ -157,9 +157,6 @@ public final class LockFreeExponentiallyDecayingReservoir implements Reservoir {
     private LockFreeExponentiallyDecayingReservoir(int size, double alpha, Duration rescaleThreshold, Clock clock) {
         // Scale alpha to nanoseconds
         this.alphaNanos = alpha * SECONDS_PER_NANO;
-        if (Double.compare(alphaNanos, 0) == 0) {
-            throw new IllegalArgumentException("Alpha value " + alpha + " is to small to be scaled to nanoseconds");
-        }
         this.size = size;
         this.clock = clock;
         this.rescaleThresholdNanos = rescaleThreshold.toNanos();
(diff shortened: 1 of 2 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -54,10 +57,11 @@
 @Beta
 public final class LockFreeExponentiallyDecayingReservoir implements Reservoir {
 
+    private static final double SECONDS_PER_NANO = .000000001D;
```
schlosna: format nit to add separators:

```suggestion
    private static final double SECONDS_PER_NANO = .000_000_001D;
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -66,6 +70,7 @@
 
     private final class State {
         private final long startTick;
+        // Count is updated after samples are successfully added to the map.
         private final AtomicLong count;
```
schlosna, line 74: @carterkozak Are you looking to pull in the changes from https://github.com/carterkozak/metrics/pull/3 here as well or separately?

carterkozak: It doesn’t make a significant difference so I haven’t pulled it in here. If it merges to the main PR branch, I’ll pull it into tritium as well.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -135,7 +155,11 @@ public void accept(Double priority, WeightedSample sample) {
     }
 
     private LockFreeExponentiallyDecayingReservoir(int size, double alpha, Duration rescaleThreshold, Clock clock) {
-        this.alpha = alpha;
+        // Scale alpha to nanoseconds
+        this.alphaNanos = alpha * SECONDS_PER_NANO;
+        if (Double.compare(alphaNanos, 0) == 0) {
+            throw new IllegalArgumentException("Alpha value " + alpha + " is to small to be scaled to nanoseconds");
```
schlosna: Under which circumstances would we now throw here? Is this a case we need we should test? Should we throw a SafeIllegalArgumentException?

carterkozak: I was worried about the Double.MIN_VALUE^2 case where the input is small enough that the conversion updates it to zero.
In retrospect I don't think the validation is meaningful since alpha=0 was allowed for the exponential decay reservoir. In the case alpha is close to zero, decay would take much longer than any jvm will be alive.

carterkozak: I'll update the upstream pr as well

schlosna (approved):
LGTM. Do you want to cut another RC and roll out to a few places?

carterkozak:
> Do you want to cut another RC and roll out to a few places?

That's the plan. After a day or two on the staging stack I'd like to cut a full release since we haven't had issues yet.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
carterkozak: https://github.com/carterkozak/metrics/pull/3 has been approved by an upstream contributor, would you like to review it before I PR it into tritium? I'll stage it here separately since this PR is already reviewed.

schlosna: sounds good

PR #869, opened 2020-10-07
Fix LockFreeExponentiallyDecayingReservoir potential race

carterkozak:
Fix a potential race in which the values map could grow beyond the
upper bound when it is rescaled. This is unlikely, however over
the course of months/years it's possible it could increase memory
pressure.

==COMMIT_MSG==
Fix LockFreeExponentiallyDecayingReservoir potential race allowing additional samples to be retained on rescale
==COMMIT_MSG==

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
@@ -132,15 +132,19 @@ public final class LockFreeExponentiallyDecayingReservoir implements Reservoir {
         State rescale(long newTick) {
             long durationNanos = newTick - startTick;
             double scalingFactor = Math.exp(-alphaNanos * durationNanos);
-            final int newCount;
+            int newCount = 0;
             ConcurrentSkipListMap<Double, WeightedSample> newValues = new ConcurrentSkipListMap<>();
             if (Double.compare(scalingFactor, 0) != 0) {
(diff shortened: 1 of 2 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -132,15 +132,19 @@ private void addSample(double priority, long value, double itemWeight, boolean b
...
                 // make sure the counter is in sync with the number of stored samples.
                 newCount = consumer.count;
-            } else {
-                newCount = 0;
+            }
+            // It's possible that more values were added while the map was scanned, those with the
+            // minimum priorities are removed.
+            while (newCount > size) {
+                newValues.pollFirstEntry();
+                newCount--;
```
schlosna: pedantic nit, I think we're guaranteed that the values won't be null since newValues won't be empty, but to make it clear, would suggest the following

```suggestion
                if (newValues.pollFirstEntry() != null) {
                    newCount--;
                }
```

carterkozak: :+1:

carterkozak: Perhaps we should throw in the null case, otherwise we might spin indefinitely.

schlosna (approved):

PR #870, opened 2020-10-07
[DO NOT MERGE] Update to dropwizard metrics 4.x

carterkozak:
Cannot merge yet, exploratory proof of concept for now.

4.x added a MovingAverages concept used by meters which we can take advantage of to optimize the common case in which averages are calculated later in the pipeline.

Before this could be considered for a merge, LazilyInitializedMovingAverages needs test coverage, and our spark version must be updated (in progress).


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LazilyInitializedMovingAverages.java
```diff
@@ -0,0 +1,96 @@
+/*
```
carterkozak, line 1: Not a great idea to couple this change with a library upgrade, however it's the piece I'm most interested in leveraging

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AbstractTaggedMetricRegistry.java
```diff
@@ -96,7 +106,7 @@ public AbstractTaggedMetricRegistry(Supplier<Reservoir> reservoirSupplier) {
     @Nonnull
     @SuppressWarnings("NoFunctionalReturnType") // metric factory
     protected Supplier<Timer> timerSupplier() {
-        return () -> new Timer(createReservoir());
+        return () -> new Timer(meterSupplier.get(), histogramSupplier.get(), Clock.defaultClock());
```
carterkozak, line 109: Dependency from timerSupplier onto histogram and meter suppliers may come back to bite me...

PR #872, opened 2020-10-09
HasBeenUpdatedCounter allows differentiating the getCount() = 0 case

iamdanfox:
## Before this PR

Internally, we try to reduce the DD$ cost by not writing down metrics until they've recorded an interaction using the following snippet:

```java
    static boolean metricHasNeverBeenUpdated(Metric metric) {
        if (metric instanceof Meter || metric instanceof Histogram || metric instanceof Timer) {
            return ((Counting) metric).getCount() == 0L;
        } else {
            // intentionally not including Counters because .inc().dec() is indistinguishable from never updated
            // we always want to report gauges
            return false;
        }
    }
```

In Dialogue, we're considering lazily constructing Counters to achieve this same outcome, of avoiding writing down `count=0` lines until the counter is actually interacted with.  I figured it might be worth just applying this optimization to _all_ counters.

## After this PR
==COMMIT_MSG==
Creating a counter in a TaggedMetricRegistry will now provide a `HasBeenUpdatedCounter` instance, which allows users to differentiate whether the counter has ever been updated and avoid writing down the count=0 case.
==COMMIT_MSG==

Hopefully this should save some DD $ for every counter that is never used, while also allowing us to write slightly simpler code in dialogue.

## Possible downsides?
- teeny performance hit, every time you interact with a counter it will now involve a write to a boolean.
- Like our other meters, histograms and timers, this PR might make graphs of Counters look a bit funny across restarts, because there won't be a count=0 point immediately after startup, so DD will have to draw a line from the last non-zero point to the first non-zero point.

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/HasBeenUpdatedCounter.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/HasBeenUpdatedCounter.java
@@ -23,9 +23,7 @@ import javax.annotation.concurrent.ThreadSafe;
 public final class HasBeenUpdatedCounter extends Counter {
     private volatile boolean hasBeenUpdated = false;
 
-    HasBeenUpdatedCounter() {
-        super();
-    }
+    HasBeenUpdatedCounter() {}
 
     /**
      * Returns false if this counter has never been interacted with, allowing users to differentiate whether
@@ -42,9 +40,9 @@ public final class HasBeenUpdatedCounter extends Counter {
     }
 
     @Override
-    public void inc(long n) {
+    public void inc(long num) {
         hasBeenUpdated = true;
-        super.inc(n);
+        super.inc(num);
     }
 
     @Override
@@ -53,8 +51,8 @@ public final class HasBeenUpdatedCounter extends Counter {
     }
 
     @Override
-    public void dec(long n) {
+    public void dec(long num) {
         hasBeenUpdated = true;
-        super.dec(n);
+        super.dec(num);
     }
 }
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/HasBeenUpdatedCounter.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/HasBeenUpdatedCounter.java
@@ -17,18 +17,17 @@
 package com.palantir.tritium.metrics.registry;
 
 import com.codahale.metrics.Counter;
-import javax.annotation.concurrent.ThreadSafe;
 
-@ThreadSafe
 public final class HasBeenUpdatedCounter extends Counter {
-    private volatile boolean hasBeenUpdated = false;
+
+    private boolean hasBeenUpdated = false;
 
     HasBeenUpdatedCounter() {}
 
     /**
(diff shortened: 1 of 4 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/HasBeenUpdatedCounter.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+package com.palantir.tritium.metrics.registry;
+
+import com.codahale.metrics.Counter;
+import javax.annotation.concurrent.ThreadSafe;
+
+@ThreadSafe
+public final class HasBeenUpdatedCounter extends Counter {
+    private volatile boolean hasBeenUpdated = false;
```
carterkozak: Is volatile necessary here? If a metric is updated prior to the reporting thread executing, happens-before relationship is fine, and parallel first use with the reporter can be reported with the next batch.

iamdanfox: Fair point, it does seem like a very low downside to race a call to `.inc()` with `.hasBeenUpdated()`!

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/HasBeenUpdatedCounter.java
```diff
@@ -0,0 +1,57 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.metrics.registry;
+
+import com.codahale.metrics.Counter;
+
+public final class HasBeenUpdatedCounter extends Counter {
```
iamdanfox, line 21: very open to better names here - I also considered just calling it `com.palantir.tritium.metrics.registry.Counter`, or `Counter2` or `LazyCounter` or `UsageTrackingCounter`

schlosna:
Hey @iamdanfox curious if it would be worth creating a delegate wrapper for each `Metric<T>` type that tracks use?

iamdanfox:
Mmm interesting, are you thinking something along the lines of:
```java 
interface MetricUsageTracker {
  boolean hasBeenUsed();
}
```
~And then our sls-logging snippet would just drop any metric which reports `hasBeenUsed() = false`?~ EDIT actually I think we'd need to preserve the old logic to apply the same optimization to plain codehale MetricRegistries, in addition to tritium TaggedMetricRegistries.

I didn't do this originally because the `.getCount() = 0` thing seemed very intuitive and understandable... also it feels a bit dirty extending the concrete Counter class already (if they ever make this final we'd be completely boned - thankfully hasn't happened [yet](https://github.com/dropwizard/metrics/blob/release/4.1.x/metrics-core/src/main/java/com/codahale/metrics/Counter.java)).

I guess the big motivation here would be if we wanted to make Gauges that could 'switch on' at some point in the future?  Have you seen use-cases for this? (I know some of our gauges currently report the same value almost 100% of the time, e.g. number of IO threads)

PR #901, opened 2020-11-09
Simplify error-prone configuration

schlosna:
## Before this PR
The error-prone configuration was overly verbose and could be simplified using the [`gradle-errorprone-plugin` DSL](https://github.com/tbroyer/gradle-errorprone-plugin#migration-from-versions-00x).

## After this PR

==COMMIT_MSG==
Simplify error-prone configuration
==COMMIT_MSG==

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,12 +76,6 @@ allprojects {
                 disableWarningsInGeneratedCode = true
                 excludedPaths = '.*(generated).*'
 
-                errorproneArgs.addAll([
-                    '-Werror',
-                    '-Xlint:deprecation',
-                    '-Xlint:unchecked',
-                ])
-
                 option('NullAway:AnnotatedPackages', 'com.palantir')
 
                 // warnings not explicitly provided by error-prone
--- a/build.gradle
+++ b/build.gradle
@@ -71,46 +71,37 @@ allprojects {
                 '-Xlint:deprecation',
                 '-Xlint:unchecked',
             ]
-            options.errorprone.disableWarningsInGeneratedCode = true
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -71,46 +71,37 @@ allprojects {
...
-                check -> warningsToPromote << check.canonicalName()
+            options.errorprone {
+                disableWarningsInGeneratedCode = true
+                excludedPaths = '.*(generated).*'
+
+                errorproneArgs.addAll([
+                    '-Werror',
+                    '-Xlint:deprecation',
+                    '-Xlint:unchecked',
+                ])
```
carterkozak: I wouldn't expect these to have any impact, shouldn't error-prone inherit them from the java compiler settings?
excludedPaths and disableWarningsInGeneratedCode should be inherited from baseline, however I think the `excludedPaths` here is a little broader than the default.

schlosna: I seem to recall there was some point in time where the `options.compilerArgs` didn't get inherited and we weren't actually catching some compiler warnings as errors so I will need to test adding in an explicit break to test that would fire if we remove these.

schlosna: updated, did a quick test locally after removing these and error-prone compiler properly flagged an `unchecked` warning and treated it as a compilation error.

carterkozak: I'd imagine these could be removed as well:
```
disableWarningsInGeneratedCode = true
excludedPaths = '.*(generated).*'
```

carterkozak (approved):
lgtm, one potential nit but it's not blocking.

PR #922, opened 2020-11-25
Use JDK 11 for building

schlosna:
## Before this PR
Using JDK 8 for building, which is incompatible with [gradle-java-format 2.0.0](https://github.com/palantir/palantir-java-format/releases/tag/2.0.0).

## After this PR

==COMMIT_MSG==
Use JDK 11 for building, maintain JDK 8 source & target compatibility for now.
Use circleci/openjdk:11-node for all circle tasks except JDK 8 unit test.
==COMMIT_MSG==


iamdanfox:
LGTM, but I think @fawind was looking into excavating this everywhere (as we're gonna have to do this to all our libs - tracing-java etc etc!)

schlosna:
Cool we can hold off on this. I noticed https://github.com/palantir/tritium/pull/920 which @fawind closed out. 

This change should still maintain Java 8 source/binary compatibility, it just switches over to running everything but `unit-test` which still uses a Java 8 runtime to validate back-compat.

schlosna:
Just for SA, this will work for now and temporarily validates JDK8 back-compat, but once we do bump `gradle-java-format` to 2.0.0, the `unit-test` on Java 8 will blow up per https://app.circleci.com/pipelines/github/palantir/tritium/899/workflows/a0fd7628-7c18-4e0d-bbd0-8fc4c99da6ad/jobs/11519 as gradle won't be able to load the JDK11 class versions. We'll have to go all in on JDK 11 for compile & test, but should still be relatively safe for JDK8 back-compat as long as we don't use any JDK9+ features (e.g. `Optional.isEmpty()`).

fawind:
Yep, putting together an excavator right now. But also happy to just merge this in, given that the code is already there.

> We'll have to go all in on JDK 11 for compile & test, but should still be relatively safe for JDK8 back-compat as long as we don't use any JDK9+ features (e.g. Optional.isEmpty()).

Do you think its worth having an escape hatch to still run tests on java 8? E.g. something like not applying the formatter plugin if on jdk 8?

fawind (approved):

schlosna:
I suspect `gradle-java-format` is the first of many to bump its target bytecode to JDK11+, so I expect others will follow. There are still a few projects where we need java 8 runtime compatibility, so I'm planning to keep tritium at source/target 1.8 compatibility for a little while longer (probably until at least early 2021).

We'll get some transitive JDK8 test coverage with AtlasDB as well https://github.com/palantir/atlasdb/blob/1135a240f7942d1b63e21c8b2dac668686b091cc/gradle/shared.gradle#L9

schlosna:
Deferring for now, https://github.com/palantir/tritium/pull/929 landed the template changes to keep building on JDK8

schlosna:
superseded by https://github.com/palantir/tritium/pull/1087

PR #934, opened 2020-12-02
Format gradle

schlosna:
## Before this PR
Gradle files formatting not standardized

## After this PR

==COMMIT_MSG==
Format gradle files

Automated via `./gradlew format -Pcom.palantir.baseline-format.gradle-files=true` per https://github.com/palantir/gradle-baseline/releases/tag/3.55.0
==COMMIT_MSG==


iamdanfox (approved):
I should probably enable this on all our OSS repos too!

Issue #482, opened 2019-10-31
Reflective error in SafepointMetrics when running in Java 11


mglazer:
## What happened?

I see this error when running with Java 11:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.palantir.tritium.metrics.jvm.SafepointMetrics (file:/opt/palantir/.binaries/tiles/4.38.0/tiles-4.38.0/service/lib/tritium-metrics-jvm-0.15.8-rc1.jar) to method sun.management.ManagementFactoryHelper.getHotspotRuntimeMBean()
WARNING: Please consider reporting this to the maintainers of com.palantir.tritium.metrics.jvm.SafepointMetrics
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

## What did you want to happen?

Not display that error since in a future release this may break more.

iamdanfox:
Think this is actually in the open-source palantir/tritium repo: https://github.com/palantir/tritium/blob/develop/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/SafepointMetrics.java

iamdanfox:
To make this go away, I think either tritium needs a `module-info.java` file declaring that it is going to reflectively access the `sun.management.ManagementFactoryHelper` class (which is in the `java.management` module), or we'd need to add some jvm args like:

```
--add-opens java.management/sun.management=ALL-UNNAMED
```

to all running WC services.

PR #989, opened 2021-02-09
Safepoint metrics no longer return warnings

j-baker:
Still cross-jvm compatible due to usage of inner class. Generally this is a bit heinous as a tactic, though I don't really see a better way without hacks. Java won't let us compile against the code with our javac settings, while using reflection directly will trigger the detection mechanisms.

### DIFF
--- a/changelog/@unreleased/pr-989.v2.yml
+++ b/changelog/@unreleased/pr-989.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Safepoint metrics no longer return illegal access warnings
+  links:
+  - https://github.com/palantir/tritium/pull/989
--- a/tritium-metrics-jvm/build.gradle
+++ b/tritium-metrics-jvm/build.gradle
@@ -14,6 +14,7 @@ dependencies {
     implementation 'com.palantir.safe-logging:preconditions'
     implementation 'com.palantir.safe-logging:safe-logging'
     implementation 'org.slf4j:slf4j-api'
+    implementation 'net.bytebuddy:byte-buddy'
 
     testImplementation 'org.assertj:assertj-core'
     testImplementation 'org.junit.jupiter:junit-jupiter'
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/SafepointMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/SafepointMetrics.java
@@ -103,7 +103,7 @@ final class SafepointMetrics {
                     .newInstance();
             gaugeImplementation.getValue();
             return Optional.of(gaugeImplementation);
-        } catch (ReflectiveOperationException e) {
+        } catch (ReflectiveOperationException | NoClassDefFoundError e) {
             log.info("Could not get the total safepoint time, these metrics will not be registered.", e);
             return Optional.empty();
         }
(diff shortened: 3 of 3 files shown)


carterkozak:
I've begun the process of unblocking a proper fix via module-info:
https://github.com/palantir/safe-logging/pull/515
https://github.com/palantir/gradle-baseline/pull/1632

j-baker:
@carterkozak your conclusion was that this does not in fact unblock a proper fix, right?

schlosna:
@j-baker Will this address the runtime warnings in #482 as we're still reflectively getting the management bean?

carterkozak:
I don't think there's a proper way to access safepoint time in jdk9+ without warnings, and 16+ will throw unless we specify `--illegal-access=permit` or `--add-opens <value>`. I'm not as familiar with the module-info type as I should be, but I haven't found any options to retain reflective access to internals without modifying jvm args.

carterkozak:
This branch doesn't appear to resolve the problem on jdk16, however it does avoid illegal reflective access warnings on older builds.

```
java.lang.IllegalAccessError: class net.bytebuddy.renamed.java.lang.Object$ByteBuddy$5fCJ5exq (in unnamed module @0x3336e6b6) cannot access class sun.management.ManagementFactoryHelper (in module java.management) because module java.management does not export sun.management to unnamed module @0x3336e6b6

	at net.bytebuddy.renamed.java.lang.Object$ByteBuddy$5fCJ5exq.<clinit>(Unknown Source)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:78)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
	at com.palantir.tritium.metrics.jvm.SafepointMetrics.getGauge(SafepointMetrics.java:106)
```

carterkozak:
I think this PR has the right idea, but instead of using byte-buddy to generate non-reflective accesses we should use java8 javac. Setup will be a little odd, so I'd propose a new project which publishes a jar with a minimal API around accessing `sun.management.HotspotRuntimeMBean` when it's available.

We use the safepoint time internally in our server project, so it will be helpful to share the extraction code instead of duplicating it.

Thoughts?

schlosna:
@carterkozak curious which JRE compatibility versions you're thinking about targeting. Would we still target a JRE 8 for now? Would we create a multi-release JAR with JRE 8 and 9+ (with proper module-info declarations) classes?

carterkozak:
I haven’t been able to get module-info to work at all when encapsulation is enforced because the sun packages are not exposed. When we need module-info definitions we’ll have to decide between bumping the minimum version to java 11, or multi-release jars. I hope that by the time it’s necessary we’ll be able to safely require java 11, mr jars are difficult to test.

I can build a jar using java 8 which references the classes without reflection and it resolves the reflective access warnings, however java 16 still fails.

j-baker:
@carterkozak maybe an easier fix is to have bytebuddy generate java 8 bytecode?

carterkozak:
I think we can close this out now that #999 has merged, now we use java 8 bytecode from https://github.com/palantir/jvm-diagnostics

carterkozak:
Closing, this has been resolved

PR #1001, opened 2021-02-23
Pin caffeine 2.9.0 for JDK 8 compatibility

schlosna:
## Before this PR

Excavator https://github.com/palantir/tritium/pull/998 bumping caffeine to 3.0 fails because tritium still requires JDK 8 while Caffeine 3.0 requires JDK 11+, see https://github.com/ben-manes/caffeine/releases/tag/v3.0.0

See also the discussion in https://github.com/palantir/tritium/pull/998#discussion_r580677931

## After this PR

==COMMIT_MSG==
Pin caffeine 2.9.0 for JDK 8 compatibility

Caffeine 3.0 requires JDK 11+, see
https://github.com/ben-manes/caffeine/releases/tag/v3.0.0
==COMMIT_MSG==

### DIFF
--- a/versions.props
+++ b/versions.props
@@ -1,5 +1,4 @@
 com.fasterxml.jackson.*:* = 2.11.3
-com.github.ben-manes.caffeine:caffeine = 2.9.0
 com.google.auto.service:auto-service = 1.0-rc4
 com.google.auto:auto-common = 0.11
 com.google.code.findbugs:jFormatString = 3.0.0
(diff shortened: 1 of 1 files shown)


robert3005 (approved):

carterkozak (approved):
Thanks!

ben-manes:
You might consider support for more dynamic restrictions in your `gradle-consistent-versions` plugin. The PR's setting would ignore any future 2.x releases, which will continue to be maintained. A resolutionStrategy can evaluate to reject versions, e.g. see the [gradle-versions-plugin](https://github.com/ben-manes/gradle-versions-plugin) which generates a report and allows for a custom rules that take into account the current and candidate versions. For example you might restrict upgrades to 2.+ rather than disable updates entirely.

schlosna:
Good flag @ben-manes as we would want to pick up any 2.x.x releases. We'll look into the best mechanism to support that automation.

PR #1006, opened 2021-03-04
Use SOURCE retention for immutables style annotations

pkoenig10:
See https://github.com/immutables/immutables/issues/291.

Example of similar fix in https://github.com/palantir/tracing-java/pull/691.

This eliminates compilation warnings for consumers that do not directly depend on immutables.


carterkozak (approved):
Thanks!

PR #1009, opened 2021-03-05
Apply AutoCloseableMustBeClosed

schlosna:
Apply `AutoCloseableMustBeClosed` from https://github.com/palantir/gradle-baseline/pull/1673

## Before this PR
Methods returning `AutoCloseable` were not marked as `@MustBeClosed`

## After this PR

==COMMIT_MSG==
Methods returning `AutoCloseable` marked as `@MustBeClosed`

./gradlew clean compileJava compileTestJava -PerrorProneApply=AutoCloseableMustBeClosed
./gradlew format
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/InstrumentedSslServerSocketFactory.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/InstrumentedSslServerSocketFactory.java
@@ -16,6 +16,7 @@
 
 package com.palantir.tritium.metrics;
 
+import com.google.errorprone.annotations.MustBeClosed;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.ServerSocket;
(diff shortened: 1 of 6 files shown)


tritium-metrics/src/main/java/com/palantir/tritium/metrics/InstrumentedSslServerSocketFactory.java
```diff
@@ -53,21 +54,25 @@
     }
 
     @Override
+    @MustBeClosed
```
carterkozak: Perhaps we should disable these on overrides unless the overridden method is also annotated with `@MustBeClosed`?

schlosna: Yeah, that would help avoid some noise as this rolls out, then we can ratchet it up.

schlosna: fixed by https://github.com/palantir/gradle-baseline/pull/1685

PR #1012, opened 2021-03-08
Allow executor queuedDuration metrics to be disabled

carterkozak:
## Before this PR
Some common executor configurations never queue work so it's
not necessary to report and index metrics describing the scenario.

## After this PR
==COMMIT_MSG==
Allow executor queuedDuration metrics to be disabled
==COMMIT_MSG==

## Possible downsides?
The flag could be used incorrectly

### DIFF
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
@@ -54,6 +54,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
+import javax.annotation.CheckReturnValue;
 import javax.annotation.Nullable;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLEngine;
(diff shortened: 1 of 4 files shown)


schlosna (approved):

tritium-metrics/src/main/java/com/palantir/tritium/metrics/MetricRegistries.java
```diff
@@ -355,11 +356,7 @@ public static ScheduledExecutorService instrument(
      * @return instrumented executor service
      */
     public static ExecutorService instrument(TaggedMetricRegistry registry, ExecutorService delegate, String name) {
-        if (delegate instanceof ScheduledExecutorService) {
-            return instrument(registry, (ScheduledExecutorService) delegate, name);
-        }
-        return new TaggedMetricsExecutorService(
-                checkNotNull(delegate, "delegate"), ExecutorMetrics.of(registry), checkNotNull(name, "name"));
+        return executor().registry(registry).name(name).executor(delegate).build();
```
jkozlowski, line 359: I think the builder works, but to play devils advocate against adding configurable points: I wonder if instead of this builder, we should just check for "delegate instanceof ThreadPoolExecutor". That thing also has BlockingQueue#getQueue which would let you check the type of queue it is, if you really wanted to?

Or add a minimum value below which we don't report the timing information, so that it would basically not be reported?

carterkozak: > I wonder if instead of this builder, we should just check for "delegate instanceof ThreadPoolExecutor"

I expect this will never succeed, we don't use ThreadPoolExecutor in most places in favor of the jboss-threads EnhancedQueueExecutor, and in many cases the executor is already wrapped with Tracing.

> Or add a minimum value below which we don't report the timing information, so that it would basically not be reported?

This gets difficult, the queue may churn quickly despite keeping the thread pool full. If we get minimums wrong, we lose signal.

PR #1018, opened 2021-03-16
Instrument only public interface methods by default

schlosna:
## Before this PR
Instrumented types were creating additional overhead observing methods that are not publicly exposed on the specified instrumented interface type. 

For example, a pre-invocation context was being generated for private method (e.g. `foo.bar.Test.somePrivateMethod(FooBar)`) invocations on an instrumented instance, even when there is no need to track these invocations.

```
InvocationContext com.palantir.tritium.event.CompositeInvocationEventHandler.preInvocation(Object, Method, Object[])
   boolean com.palantir.tritium.proxy.InstrumentedFilterPredicate$9.match(FooBar)
      boolean foo.bar.Test.somePrivateMethod(FooBar)
         void foo.bar.Test.process(FooBar)
            void com.palantir.tritium.proxy.InstrumentedTest$10.process(Collection)
```

## After this PR

Methods that are not part of specified interface will not be instrumented by default.

==COMMIT_MSG==
Instrument only public interface methods by default
==COMMIT_MSG==

## Possible downsides?

This is a behavior change in that previously instrumented methods that were not part of interface being instrumented would produce instrumentation events (e.g. metrics, logs, etc.) this is generally not what one might want as the instrumentation should be for the public facing methods of a given interface.

### DIFF
--- a/tritium-core/src/main/java/com/palantir/tritium/event/InstrumentationFilters.java
+++ b/tritium-core/src/main/java/com/palantir/tritium/event/InstrumentationFilters.java
@@ -18,6 +18,7 @@ package com.palantir.tritium.event;
 
 import com.palantir.tritium.api.event.InstrumentationFilter;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import javax.annotation.Nonnull;
 
 public enum InstrumentationFilters implements InstrumentationFilter {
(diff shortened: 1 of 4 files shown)


schlosna:
https://openjdk.java.net/jeps/213

tritium-core/src/main/java/com/palantir/tritium/event/InstrumentationFilters.java
```diff
@@ -36,6 +37,16 @@ public boolean shouldInstrument(@Nonnull Object _instance, @Nonnull Method _meth
...
         }
+    },
+
+    /** Instrument public interface methods. */
+    INSTRUMENT_PUBLIC_INTERFACE_METHODS {
+        @Override
+        public boolean shouldInstrument(@Nonnull Object _instance, @Nonnull Method method, @Nonnull Object[] _args) {
+            return Modifier.isPublic(method.getModifiers())
+                    && method.getDeclaringClass().isInterface()
+                    && !Modifier.isStatic(method.getModifiers());
```
carterkozak, line 48: I would have expected this behavior to be identical to `INSTRUMENT_ALL`, would it make more sense to update INSTRUMENT_ALL instead? I don't think the dynamic proxy codepath will ever give us private or static methods, the byte-buddy implementation may need to be updated, but I might consider that an implementation detail of the byte-buddy instrumentation. What do you think?

PR #1021, opened 2021-03-19
Don't include performance logging in the Tritium.instrument method

mglazer:
## Before this PR

Java9+ have a performance regression when constructing loggers
with dynamic names. Until the upstream Log4j bug is fixed
we probably want to remove this.

## After this PR

==COMMIT_MSG==

Don't include performance logging in the Tritium.instrument method
==COMMIT_MSG==

## Possible downsides?

This performance logging was used and needed by some user.

### DIFF
--- a/changelog/@unreleased/pr-1021.v2.yml
+++ b/changelog/@unreleased/pr-1021.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Don't include performance logging in the Tritium.instrument method
+  links:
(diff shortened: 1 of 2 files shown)


carterkozak:
Some context here https://github.com/apache/logging-log4j2/pull/475
and https://issues.apache.org/jira/browse/LOG4J2-2940

I think this is a reasonable path forward, we should file a ticket to re-enable logging once the upstream fix has been released.

schlosna (approved):

PR #1028, opened 2021-03-28
Create v1 APIs

schlosna:
## Before this PR
Packages and classes clashed across modules, presenting challenges for a JDK 9+ modules world.

## After this PR

==COMMIT_MSG==
Create v1 APIs and deprecate previous APIs
==COMMIT_MSG==

## Possible downsides?

Still refactoring/cleanup needed for `tritium-metrics` and `tritium-lib`. There is excess duplicate code until we can excavate cleanup and post v1.0 where we can remove deprecated methods.


settings.gradle
```diff
@@ -1,12 +1,16 @@
...
+//include 'tritium-api-v0'
 include 'tritium-caffeine'
+//include 'tritium-core-v0'
 include 'tritium-core'
 include 'tritium-jmh'
 include 'tritium-lib'
+//include 'tritium-lib-v0'
 include 'tritium-proxy'
 include 'tritium-metrics'
+//include 'tritium-metrics-v0'
```
schlosna: - [ ] TODO: possibly create backward compatibility modules

schlosna (commented):
There's a lot of code to maintain full binary backward compatibility, how long do we want to keep it? Do we want to produce a separate backward compatibility module? Should we combine some of the modules into just one or two (tritium-caffeine being the main outlier as not every library/service adds caches).

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
```diff
@@ -47,24 +38,14 @@ private CaffeineCacheStats() {}
...
      *
-     * @deprecated use {@link #registerCache(TaggedMetricRegistry, Cache, String)}
-     *
      * @param registry metric registry
      * @param cache cache to instrument
      * @param name cache name
+     * @deprecated use {@link CaffeineCacheMetrics#registerCache(TaggedMetricRegistry, Cache, String)}
      */
     @Deprecated
     public static void registerCache(MetricRegistry registry, Cache<?, ?> cache, String name) {
```
carterkozak, line 47: Should we remove support for non-tagged registries in 1.0?

schlosna: I'm not sure we're quite ready to completely remove the `MetricRegistry` usages quite yet. I'm also wondering which module & package should contain the `TaggedMetricRegistry` APIs.

The way I'm currently thinking is creating a 1.0 with Java 11 source/target compatibility and with all of these older APIs deprecated. Then cleanup the remaining consumers using deprecated types like all the original invocation event handlers and `MetricRegistry`, and then cutting 2.0 at that point.

carterkozak: If we're renaming packages, should we use new maven coordinates as well to allow old and new to live side-by-side?

> Then cleanup the remaining consumers using deprecated types like all the original invocation event handlers and MetricRegistry, and then cutting 2.0 at that point.

Would we rename packages again from `v1` to `v2`?

We may want to avoid modifying `tritium-registry` given the breadth of dependencies, and the static state. Several consumers depend on data written into the SharedTaggedMetricRegistries singleton getting recorded by our server library.  I could see a case for extracting the metric registry API and implementation into a separate repository where we can retain java8 support longer.

schlosna: Right now this is mostly a side project, so wanted to throw this draft up for thoughts.

I'm using v1 as a placeholder right now just to get a sense of what the intended exported APIs should be. Definitely interested in thoughts on how best to approach the modules & packaging as I know there are pros/cons to whether to include a major version in fully qualified names. 

As you mentioned, the `TaggedMetricRegistry` & `MetricName` bits are the ones that will be most difficult to evolve and I'm open to pulling those out to a separate module/repo as needed.

schlosna:
closing out for now

PR #1041, opened 2021-04-01
Add ability to register custom cache stats

pkoenig10:
After profiling an internal product, we noticed that a non-negligible amount of CPU time was spent updating the `LongAdder` fields in Caffeine's `ConcurrentStatsCounter`. To improve performance, we changed the implementation to count stats per-request in a non-thread-safe way and then batch update the registered counters after handling the request.

Unfortunately this required using some methods in the internal but public `InternalCacheMetrics`. We needed to implement `Stats` and use the static methods to register our metrics with the canonical names.

For details, see https://g.p.b/foundry/multipass/pull/11551

This PR is an attempt to expose the ability to use custom `Stats` implementations without using internal classes.

There are a lot of cache-related classes in Tritium and the division of responsibility is not obvious. I attempted to implement this in a way I thought was most inline with the current design, but please let me know if you'd prefer to structure this differently - I have no real preference on the design here.

### DIFF
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/CacheMetrics.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/CacheMetrics.java
@@ -1,0 +1,93 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.metrics;
+
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.RatioGauge;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+import java.util.Optional;
+import java.util.function.BiConsumer;
+
+public final class CacheMetrics {
+
+    private CacheMetrics() {}
+
+    public static void register(TaggedMetricRegistry taggedMetrics, Stats stats, String cacheName) {
+        InternalCacheMetrics.createMetrics(stats, InternalCacheMetrics.taggedMetricName(cacheName))
+                .forEach(taggedMetrics::registerWithReplacement);
+    }
+
+    static void forEachMetric(CacheMetrics.Stats stats, BiConsumer<String, Gauge<?>> consumer) {
+        consumer.accept("cache.estimated.size", stats.estimatedSize());
+        consumer.accept("cache.request.count", stats.requestCount());
+        consumer.accept("cache.hit.count", stats.hitCount());
+        consumer.accept("cache.hit.ratio", stats.hitRatio());
+        consumer.accept("cache.miss.count", stats.missCount());
+        consumer.accept("cache.miss.ratio", stats.missRatio());
+        consumer.accept("cache.eviction.count", stats.evictionCount());
+        consumer.accept("cache.load.success.count", stats.loadSuccessCount());
+        consumer.accept("cache.load.failure.count", stats.loadFailureCount());
+        consumer.accept("cache.load.average.millis", stats.loadAverageMillis());
+        stats.maximumSize().ifPresent(maximumSizeGauge -> consumer.accept("cache.maximum.size", maximumSizeGauge));
+        stats.weightedSize().ifPresent(weightedSizeGauge -> consumer.accept("cache.weighted.size", weightedSizeGauge));
+    }
+
+    public interface Stats {
+
+        Gauge<Long> estimatedSize();
+
+        Optional<Gauge<Long>> weightedSize();
+
+        Optional<Gauge<Long>> maximumSize();
+
+        Gauge<Long> requestCount();
+
+        Gauge<Long> hitCount();
+
+        Gauge<Long> missCount();
+
+        Gauge<Long> evictionCount();
+
+        Gauge<Long> loadSuccessCount();
+
+        Gauge<Long> loadFailureCount();
+
+        Gauge<Double> loadAverageMillis();
+
+        default Gauge<Double> hitRatio() {
+            return new RatioGauge() {
+                @Override
+                protected Ratio getRatio() {
+                    return Ratio.of(
+                            hitCount().getValue().doubleValue(),
+                            requestCount().getValue().doubleValue());
+                }
+            };
+        }
+
+        default Gauge<Double> missRatio() {
+            return new RatioGauge() {
+                @Override
+                protected Ratio getRatio() {
+                    return Ratio.of(
+                            missCount().getValue().doubleValue(),
+                            requestCount().getValue().doubleValue());
+                }
+            };
+        }
+    }
+}
--- a/.palantir/revapi.yml
+++ b/.palantir/revapi.yml
@@ -1,0 +1,14 @@
+acceptedBreaks:
+  "0.18.8":
+    com.palantir.tritium:tritium-metrics:
+    - code: "java.class.removed"
+      old: "interface com.palantir.tritium.metrics.InternalCacheMetrics.Stats"
+      justification: "Changed internal methods"
+    - code: "java.method.parameterTypeChanged"
+      old: "parameter <K> com.google.common.collect.ImmutableMap<K, com.codahale.metrics.Gauge<?>>\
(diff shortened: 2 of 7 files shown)


ben-manes:
Just to double check given my lack of knowledge on these projects, but you were able to replace Caffeine's `ConcurrentStatsCounter` with your own `StatsCounter`, correct?

Naively, it looks like this metrics binder exposes stats by querying the cache which delegates to `StatsCounter` for the instance. If that is the case, then would any changes be needed here? I'd think that Tritium would report a stale snapshot until your StatsCounter produced an up to date one.

If you went about another route then you might want to look at a writing a custom `StatsCounter` (optional param for `recordStats()`) that does your more efficient updates.

carterkozak:
+1 for a custom StatsCounter implementation. I've been thinking about updating Tritium to provide a custom StatsCounter which can be used rather than reading from the existing StatsCounter, this would solve a few problems:

1. It's impossible to register one without cache stats being enabled
2. While not ideal, we still get _some_ data if two caches are created with the same name because most stats can safely be accumulated to counters/meters/timers rather than extracted via Gauges. We do this more than you'd think/hope
3. Allows us to tune for performance, we have more leverage to collect the specific data we need when we provide the accumulator rather than accessing data stored within another accumulator.

I haven't gotten around to it because it would be nontrivial to automate migration from registering cache instances to the new StatsCounter, and the metric names would be different (as well perhaps as the metrics themselves). I think it's a perfect fit your your use case though, @pkoenig10.

pkoenig10:
> Just to double check given my lack of knowledge on these projects, but you were able to replace Caffeine's `ConcurrentStatsCounter` with your own `StatsCounter`, correct?

Not quite. We have a global (long-lived, concurrently accessed) cache that is used by individual request threads. We don't need the cache stats to update immediately when the cache is used, so each request manually keeps track of its own hits and misses and updates our custom `Stats` implementation after the request is finished. We're no longer using Caffeine's stats mechanisms.

It not clear to me how a custom `StatsCounter` would solve the problem, since we only have a single cache which _is_ being accessed concurrently. We just don't want to pay the cost of a ton of CAS operations when we don't need super low latency updates to our cache metrics.

For reference, our implementation looks something like this:
```java
public final class ManualCacheStats implements Stats {

    private final Cache<?, ?> cache;

    private final LongAdder numHits = new LongAdder();
    private final LongAdder numMisses = new LongAdder();

    public ManualCacheStats(Cache<?, ?> cache) {
        this.cache = cache;
    }

    // We create a new CacheStatsRecorder at the start of each request and close it at the end of the request
    public CacheStatsRecorder recorder() {
        return new CacheStatsRecorder();
    }

    // Implement the Stats methods, with zero values for ones we can't implement like evictionCount
    ...

    @NotThreadSafe
    private final class CacheStatsRecorder implements AutoCloseable {

        private long hits;
        private long misses;

        public void markHit() {
            hits++;
        }

        public void markMiss() {
            misses++;
        }

        @Override
        public void close() {
            numHits.add(hits);
            numMisses.add(misses);
        }
    }
}
```

ben-manes:
You could use a thread local within your StatsCounter and aggregate the results on `snapshot()`. This way the thread never suffers CAS conflicts and that per-thread version can be optimized. As threads can die, you probably want to summarize their results back into a global snapshot, which the example below does using weak references. The penalty of this approach is the memory overhead depending on how many threads you have.

This can be optimized further as for convenience this used a per-thread `ConcurrentStatsCounter` which is heavy but good enough for a proof-of-concept demonstration.

```java
public final class ThreadLocalStatsCounter implements StatsCounter {
  private final ThreadLocal<StatsCounter> threadLocalStatsCounter;
  private final Cache<Thread, StatsCounter> cache;

  private volatile CacheStats stats;

  public ThreadLocalStatsCounter() {
    cache = Caffeine.newBuilder()
        .weakKeys()
        .removalListener((Thread t, StatsCounter statsCounter, RemovalCause cause) -> {
          synchronized (this) {
            stats = stats.plus(statsCounter.snapshot());
          }
        }).build();
    threadLocalStatsCounter = ThreadLocal.withInitial(() -> {
      var statsCounter = new ConcurrentStatsCounter();
      cache.put(Thread.currentThread(), statsCounter);
      return statsCounter;
    });
    stats = CacheStats.empty();
  }

  @Override public void recordHits(int count) {
    threadLocalStatsCounter.get().recordHits(count);
  }
  @Override public void recordMisses(int count) {
    threadLocalStatsCounter.get().recordMisses(count);
  }
  @Override public void recordLoadSuccess(long loadTime) {
    threadLocalStatsCounter.get().recordLoadSuccess(loadTime);
  }
  @Override public void recordLoadFailure(long loadTime) {
    threadLocalStatsCounter.get().recordLoadFailure(loadTime);
  }
  @Override public void recordEviction(int weight, RemovalCause cause) {
    threadLocalStatsCounter.get().recordEviction(weight, cause);
  }

  @Override
  public CacheStats snapshot() {
    var snapshot = stats;
    for (var statsCounter : cache.asMap().values()) {
      snapshot = snapshot.plus(statsCounter.snapshot());
    }
    return snapshot;
  }
}
```

pkoenig10:
This is certainly a reasonable approach. I considered using thread locals but then figured we could avoid the thread local map lookup by just recording the stats ourselves, since we only really care about hits and misses.

Although if we're already doing a map lookup in the cache, another thread local map lookup probably isn't too bad - I'd imagine it's better than the multiple CAS operations required by `LongAddr`.

pkoenig10:
@carterkozak one downside of the custom `StatsCounter` approach is that it can't provide the size metrics, which I tend to find fairly valuable.

ben-manes:
The polling approach seems fine and does offer richer options. I meant naive in terms of my understanding of this code; sorry if that came across as towards the implementation usage.

I think that a custom, low overhead StatsCounter should solve the original problem without needing any code changes to this project. This PR might be good regardless, I can’t judge that. Of course replacing the reporting to a cheaper subset also makes a lot of sense given the performance critical nature where the default stats were too expensive.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/CacheMetrics.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private CacheMetrics() {}
+
+    public static void register(TaggedMetricRegistry taggedMetrics, Stats stats, String cacheName) {
+        InternalCacheMetrics.createMetrics(stats, InternalCacheMetrics.taggedMetricName(cacheName))
+                .forEach(taggedMetrics::registerWithReplacement);
+    }
+
+    static void forEachMetric(CacheMetrics.Stats stats, BiConsumer<String, Gauge<?>> consumer) {
+        consumer.accept("cache.estimated.size", stats.estimatedSize());
```
pkoenig10, line 35: Would be nice to make all of these `Optional` so we don't have to report metrics we know are garbage.

pkoenig10:
Another downside of the thread local approach is that I think we'd be forced to at least use volatile longs in our per-thread stats counters to ensure that updates are read by the thread taking the stats snapshots.

ben-manes:
Yes, a positive aspect is that only one thread increments those volatile counters so it can be more lenient in the semantics. 

|Benchmark|          Score|
|---|---|
|plainIncrement|     595M ops/s
|racyIncrement|      119M ops/s
|atomicIncrement|    120M ops/s
|opaqueIncrement|       531M ops/s
|releaseIncrement|      591M ops/s

So for example an increment 5x slower, though still fast for a single thread. If you relaxed that by piggybacking on other memory barriers then it can be equivalent. However, I don't think you'll observe a meaningful difference and should prefer an atomic increment for simplicity.

schlosna:
I haven't had cycles to dig in on this and the perf & scalability pieces of where the default `LongAdder` overhead is too high for MP's cache utilization. I think you all have a workaround for now, but happy to discuss more.

Directionally I think we want folks using `Caffeine#recordStats(Supplier<? extends StatsCounter>)` so I pushed a draft of a basic version in https://github.com/palantir/tritium/pull/1075 for thoughts.

pkoenig10:
> Directionally I think we want folks using Caffeine#recordStats(Supplier<? extends StatsCounter>) so I pushed a draft of a basic version in #1075 for thoughts.

That's one option. Although a `StatsCounter` implementation that supports what we need here requires maintaining a per-thread map of local stats counters (as described above) and volatile longs in the local stats counters (to support reading from the metric recording thread).

At that point, I'm not really convinced that the nicety of using `StatsCounter` outweighs the cost when compared to just manually updating the metrics ourselves after finishing a request.

For reference, here's our manual implementation:
https://g.p.b/foundry/mp/pull/11551

And here's a commit demonstrating what a `StatsCounter` implementation would look like:
https://g.p.b/foundry/mp/commit/c237f4d6ad

ben-manes:
On a related note, I do find it concerning that you narrowed down to `LongAdder` being a bottleneck. That is quite surprising since it has a contention avoidance design, and merely increments a counter whereas application work will be much costlier (e.g. I/O). While there is overhead, that typically is noise compared to everything else the application is doing.

LongAdder tries to rebalance threads to per-core counters, so there is no CAS conflicts when they perform an update. As threads may be relocated over time the thread may rehash to an alternative location as it searches. We use a similar idea for finding ring buffers to append to, in order to record and replay the cache events onto the eviction policy.

PR #1072, opened 2021-04-30
Tritium SSLEngine handshake instrumentation avoids reflection

carterkozak:
Checking for older java 8 releases was necessary when ALPN support
was initially backported, however at this point it's incredibly
dangerous to run outdated runtimes, and our environments upgraded
long ago. Furthermore, the risk of introducing bugs when using
an old JRE is incredibly low, and requires a framework which
attempts to reflectively discover SSLEngine methods.

## After this PR
==COMMIT_MSG==
Tritium SSLEngine handshake instrumentation avoids reflection
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/InstrumentedSslEngine.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/InstrumentedSslEngine.java
@@ -16,16 +16,11 @@
 
 package com.palantir.tritium.metrics;
 
-import com.palantir.logsafe.exceptions.SafeIllegalStateException;
-import java.lang.reflect.Method;
 import java.nio.ByteBuffer;
-import java.security.AccessController;
-import java.security.PrivilegedAction;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiFunction;
-import javax.annotation.Nullable;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLEngineResult;
 import javax.net.ssl.SSLException;
@@ -34,7 +29,7 @@ import javax.net.ssl.SSLSession;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-class InstrumentedSslEngine extends SSLEngine {
+final class InstrumentedSslEngine extends SSLEngine {
 
     private static final Logger log = LoggerFactory.getLogger(InstrumentedSslEngine.class);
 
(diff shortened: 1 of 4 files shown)


carterkozak:
I have a PR internally which updates circleci templates for oss java-lib projects, I've verified it produces identical results to what I've committed here.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/InstrumentedSslEngine.java
```diff
@@ -46,34 +41,17 @@
...
+     * ALPN API methods {@link #getApplicationProtocol()}, {@link #getHandshakeApplicationProtocol()},
+     * {@link #setHandshakeApplicationProtocolSelector(BiFunction)}, {@link #getHandshakeApplicationProtocolSelector()}.
+     * Using this method on an older jre8 release will not fail outright, but some callers may assume
+     * ALPN is supported by this implementation when calls to the delegate fail. Reflection is not
+     * used here because it increases risk of incorrect behavior using graalvm native-images.
+     *
+     * @see <a href="https://openjdk.java.net/jeps/244">JEP 244</a>
+     * @see <a href="https://bugs.openjdk.java.net/browse/JDK-8230977">JDK-8230977</a>
      */
     static SSLEngine instrument(SSLEngine engine, TlsMetrics metrics, String name) {
```
schlosna, line 54: Given ALPN was backported to 8u252 per https://bugs.openjdk.java.net/browse/JDK-8230977 so should we do something where we don't instrument on JDK 1.8 < 8u252, e.g. checking `java.version` system property rather than the reflection checks?

```suggestion
     * Note that this assumes up-to-date JVMs, older Java 8 releases did not provide
     * ALPN API methods {@link #getApplicationProtocol()}, {@link #getHandshakeApplicationProtocol()},
     * {@link #setHandshakeApplicationProtocolSelector(BiFunction)}, {@link #getHandshakeApplicationProtocolSelector()}.
     * Using this method on an older jre8 release will not fail outright, but some callers may assume
     * ALPN is supported by this implementation when calls to the delegate fail. Reflection is not
     * used here because it increases risk of incorrect behavior using graalvm native-images.
     *
     * @see <a href="https://openjdk.java.net/jeps/244">JEP 244</a>
     * @see <a href="https://bugs.openjdk.java.net/browse/JDK-8230977">JDK-8230977</a>
     */
    static SSLEngine instrument(SSLEngine engine, TlsMetrics metrics, String name) {
        if (hasAlpn()) {
            return engine;
        }
```

and

```
    private static final Pattern java8MinorVersionRegex = Pattern.compile("1.8.0_([0-9]+)");

    private static boolean hasAlpn() {
        OptionalInt minorUpdateVersion = parseMinorUpdateVersion(System.getProperty("java.version"));
        if (minorUpdateVersion.isPresent()) {
            // ALPN support exists in JDK 8u252+ https://bugs.openjdk.java.net/browse/JDK-8230977
            return minorUpdateVersion.getAsInt() >= 252;
        }
        return true; // ALPN exists in JDK 9+ https://bugs.openjdk.java.net/browse/JDK-8051498
    }

    private static OptionalInt parseMinorUpdateVersion(String javaVersion) {
        if (javaVersion != null) {
            Matcher matcher = java8MinorVersionRegex.matcher(javaVersion);
            if (matcher.matches()) {
                try {
                    return OptionalInt.of(Integer.parseInt(matcher.group(1)));
                } catch (NumberFormatException e) {
                    return OptionalInt.empty();
                }
            }
        }
        return OptionalInt.empty();
    }
```

carterkozak: I considered a conditional here, we've written [similar checks before](https://github.com/apache/httpcomponents-core/pull/215), but I'm not sure it's worthwhile at this point for a few reasons:
* Our use of java 8 is waning
* Java 8 builds which don't support ALPN are considered insecure
* We supported conditionally implementing these methods to work sympathetically with other libraries, however those libraries have been updated to make that unnecessary: https://github.com/undertow-io/undertow/pull/977

carterkozak: Happy to add the switch if you prefer that approach though, I don't have a terribly strong opinion.

schlosna: yeah, I think I'm fine if we assume at least 8u252+ and if folks hit `NoSuchMethodError` or other odd behavior the solution is upgrade to supported JDK runtime.

schlosna (approved):

PR #1075, opened 2021-05-03
CaffeineCacheStats.record for Caffeine stats recording

schlosna:
## Before this PR
Setting up Caffeine stats recording was slightly awkward and required enabling `Caffeine#recordStats()` then `CaffeineCacheStats#registerCache`.

## After this PR
Similar to https://github.com/palantir/tritium/pull/1041

==COMMIT_MSG==
CaffeineCacheStats.record for Caffeine stats recording

This streamlines providing a `StatsCounter` with tagged metrics for use with `Caffeine#recordStats(Supplier<? extends StatsCounter>)` during cache building, for example

```
Caffeine.newBuilder()
    .recordStats(() -> CaffeineCacheStats.record(taggedMetricRegistry, "cacheName")
    .build()
```

==COMMIT_MSG==

## Possible downsides?
Haven't provided similar for Guava caches, but those should be migrated to Caffeine anyway

### DIFF
--- a/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
+++ b/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
@@ -110,7 +110,6 @@ public final class CaffeineCacheStats {
      * @return stats counter
      */
     public static StatsCounter record(TaggedMetricRegistry registry, @Safe String name) {
-        //        return CaffeineCacheTaggedMetrics.statsCounter(registry, name);
         Function<String, MetricName> nameFunction = InternalCacheMetrics.taggedMetricName(name);
         return new TaggedCaffeineStatsCounter(
                 registry.counter(nameFunction.apply("cache.hit.count")),
--- a/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
+++ b/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
@@ -21,13 +21,20 @@ import static com.palantir.logsafe.Preconditions.checkNotNull;
 import com.codahale.metrics.Counter;
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Snapshot;
+import com.codahale.metrics.Timer;
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.RemovalCause;
+import com.github.benmanes.caffeine.cache.stats.CacheStats;
+import com.github.benmanes.caffeine.cache.stats.StatsCounter;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.palantir.logsafe.Safe;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.tritium.metrics.InternalCacheMetrics;
 import com.palantir.tritium.metrics.MetricRegistries;
+import com.palantir.tritium.metrics.registry.MetricName;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
```diff
@@ -90,6 +97,34 @@ public static void registerCache(TaggedMetricRegistry registry, Cache<?, ?> cach
...
+     * Caffeine.newBuilder()
+     *     .recordStats(() -> CaffeineCacheStats.record(taggedMetricRegistry, "name")
+     *     .build()
+     * }
+     * @param registry tagged metric registry
+     * @param name safe cache name
+     * @return stats counter
+     */
+    public static StatsCounter record(TaggedMetricRegistry registry, @Safe String name) {
+        //        return CaffeineCacheTaggedMetrics.statsCounter(registry, name);
```
schlosna:
```suggestion
```

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
```diff
@@ -90,6 +97,34 @@ public static void registerCache(TaggedMetricRegistry registry, Cache<?, ?> cach
...
+                registry.counter(nameFunction.apply("cache.hit.count")),
+                registry.counter(nameFunction.apply("cache.miss.count")),
+                registry.timer(nameFunction.apply("cache.load.success")),
+                registry.timer(nameFunction.apply("cache.load.failure")),
+                registry.counter(cause(nameFunction.apply("cache.eviction.count"), "total")),
+                registry.counter(cause(nameFunction.apply("cache.eviction.count"), "explicit")),
+                registry.counter(cause(nameFunction.apply("cache.eviction.count"), "replaced")),
+                registry.counter(cause(nameFunction.apply("cache.eviction.count"), "collected")),
+                registry.counter(cause(nameFunction.apply("cache.eviction.count"), "expired")),
+                registry.counter(cause(nameFunction.apply("cache.eviction.count"), "size")));
```
carterkozak, line 124: Thoughts on using metric-schema for these?

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
```diff
@@ -90,6 +97,34 @@ public static void registerCache(TaggedMetricRegistry registry, Cache<?, ?> cach
         }
     }
 
+    /**
+     * Creates a {@link StatsCounter} with tagged metrics for tracking caches hits, misses, evictions, and load times.
+     * <p>Example usage:</p>
+     * {@code
+     * Caffeine.newBuilder()
+     *     .recordStats(() -> CaffeineCacheStats.record(taggedMetricRegistry, "name")
```
carterkozak, line 105: I wonder if we should return a supplier for simplicity?

carterkozak (commented):
Looks great! I think this is a much safe approach to record metrics without inadvertently replacing existing instrumentation.
I added a couple non-blocking comments.

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
schlosna: yeah that would make call sites cleaner. will need to `@SuppressWarnings("NoFunctionalReturnType")` though I think that's ok for this specific use

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
schlosna: might want to do the metric-schema migration for all the Guava & Caffeine caches in a separate PR (possibly before this)

schlosna:
there are still a few gaps between this approach and `registerCache`, e.g. estimated size and request count

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
carterkozak: We could define an interface which extends `Supplier<StatsCounter>` as well, though that might be less obvious to library consumers.

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
carterkozak: Either way I think we'll end up with different names for metrics produced by the StatsCounter vs polling snapshots, since metric names are suffixed based on the metric type.

I think most of these counters should be meters, dropwizard counters are recorded as gauges in several backends because the values can both increment and decrement, where meters can only increase.

schlosna:
closing out for now

PR #1083, opened 2021-05-11
Tritium provides an annotation processor alternative to dynamic proxies

carterkozak:
This makes the implementation more obvious, and moves work to build time
rather than runtime bytecode generation.

Risks:
Unlike the runtime implementations, this solution does not support
adding arbitrary interfaces from the implementation to the instrumented
object. This may make implementations more obvious, but means that
there are cases (e.g. dependency injection) in which the annotation
processor does not produce behavior identical to the proxy.

==COMMIT_MSG==
Tritium provides an annotation processor alternative to dynamic proxies
==COMMIT_MSG==


.gitattributes
```diff
@@ -0,0 +1 @@
+**/src/test/resources/**/*.java.generated linguist-generated=true
```
carterkozak, line 1: https://docs.github.com/en/github/administering-a-repository/customizing-how-changed-files-appear-on-github

This collapses generated code in diffs by default.

tritium-core/src/main/java/com/palantir/tritium/event/Handlers.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /**
+     * Identical to {@link #pre(InvocationEventHandler, InstrumentationFilter, Object, Method, Object[])}
+     * except that {@link InvocationEventHandler#isEnabled()} is checked along with
+     * {@link InstrumentationFilter#shouldInstrument(Object, Method, Object[])}. This should be used when
+     * argument array allocation has already occurred and cannot be avoided.
+     */
+    @Nullable
+    public static InvocationContext preWithEnabledCheck(
```
carterkozak, line 60: Method is very similar to `pre`, they should share a common implementation with additional branches.

I considered always checking `handler.isEnabled()`, but figured since we're doing codegen, we might as well avoid unnecessary overhead of checking the handler twice.

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentationAdvice.java
```diff
@@ -52,53 +49,20 @@ static InvocationContext enter(
...
-        } catch (RuntimeException | Error t) {
-            if (logger.isWarnEnabled()) {
-                logger.warn(
-                        "Failure occurred handling 'preInvocation' invocation on: {}",
-                        UnsafeArg.of("instance", Objects.toString(proxy)),
-                        t);
-            }
-            return null;
-        }
+        return Handlers.preWithEnabledCheck(eventHandler, filter, proxy, methods[index], arguments);
```
carterkozak, line 53: This may have a substantial (positive) impact on byte-buddy bytecode generation performance. Advice methods are inlined into each instrumented method, this change shares additional code for better reuse and less resulting bytecode.

tritium-processor/src/test/resources/com/palantir/tritium/examples/InstrumentedParameterized.java.generated
```diff
@@ -0,0 +1,127 @@
...
+    private static final Method METHOD_PARAMETER;
+
+    private static final Method CONSUME_LIST_OF_PARAMETER;
+
+    static {
+        try {
+            CONSUME_PARAMETER = Parameterized.class.getMethod("consumeParameter", Object.class);
+            PRODUCE_PARAMETER = Parameterized.class.getMethod("produceParameter");
+            METHOD_PARAMETER = Parameterized.class.getMethod("methodParameter", Object.class, Object.class);
+            CONSUME_LIST_OF_PARAMETER = Parameterized.class.getMethod("consumeListOfParameter", List.class);
```
carterkozak, line 28: I opted against adding a utility function for these because graalvm native-image automatically supports `NAME.class.getMethod(CONSTANT_STRING, CONSTANT_CLASSES.class)` without additional configuration. I haven't verified that this works in practice yet though.

carterkozak:
In an ideal world we wouldn’t use reflect.method or an Object array of parameters, preferring a more specific method destructor and view over the input parameters without array allocation.
I implemented this to avoid friction with existing handler implementation, as a drop-in replacement in cases that don’t rely on additional interfaces.

tritium-annotations/src/main/java/com/palantir/tritium/annotations/TritiumInstrument.java
```diff
@@ -0,0 +1,35 @@
...
+ * using the same package, and name prefixed with {@code Instrumented}.
+ * <p/>
+ * Externally defined interfaces may be instrumented by defining an interface with the requested interface
+ * as a supertype. The generator will produce an instrumentation class which can wrap the super-interface.
+ * When type parameters are present, the empty interface must define and pass along type parameters which
+ * exactly match the target interface.
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.CLASS)
+public @interface TritiumInstrument {}
```
schlosna: Nit: should this just be `Instrument` given package?

```suggestion
public @interface Instrument {}
```

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentation.java
```diff
@@ -334,59 +323,10 @@ public boolean isAlive() {
         }
     }
 
-    private enum LoggerInitializer implements LoadedTypeInitializer {
```
schlosna, line 337: Why don't we need this any longer? Is this because the injected advice is not using the logger anymore?

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentationAdvice.java
schlosna: 👍 on extracting `Handlers`

tritium-processor/build.gradle
```diff
@@ -0,0 +1,23 @@
+apply plugin: 'com.palantir.external-publish-jar'
+
+dependencies {
+    compile project(':tritium-annotations')
```
schlosna, line 4:
```suggestion
    implementation project(':tritium-annotations')
```

tritium-processor/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+    compile project(':tritium-api')
```
schlosna, line 5:
```suggestion
    implementation project(':tritium-api')
```

tritium-processor/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+    compile project(':tritium-registry')
```
schlosna, line 6:
```suggestion
    implementation project(':tritium-registry')
```

tritium-processor/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+    compile project(':tritium-core')
```
schlosna, line 7:
```suggestion
    implementation project(':tritium-core')
```

tritium-processor/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+
+    compile 'com.squareup:javapoet'
```
schlosna, line 9:
```suggestion
    implementation 'com.squareup:javapoet'
```

tritium-processor/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+    compile 'com.google.guava:guava'
```
schlosna, line 10:
```suggestion
    implementation 'com.google.guava:guava'
```

tritium-processor/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+    compile 'com.google.auto:auto-common'
```
schlosna, line 13:
```suggestion
    implementation 'com.google.auto:auto-common'
```

tritium-processor/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+    compile 'com.google.code.findbugs:jsr305'
```
schlosna, line 14:
```suggestion
    implementation 'com.google.code.findbugs:jsr305'
```

tritium-processor/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+    compile 'com.palantir.safe-logging:preconditions'
```
schlosna, line 15:
```suggestion
    implementation 'com.palantir.safe-logging:preconditions'
```

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessor.java
```diff
@@ -0,0 +1,438 @@
...
+        return currentElements;
+    }
+
+    private boolean getInterfaces(
+            List<? extends TypeMirror> mirrors,
+            List<DeclaredType> allInterfaces,
+            @Nullable List<DeclaredType> minimalInterfaces) {
+        boolean error = false;
+        for (TypeMirror mirror : mirrors) {
+            error |= getInterfaces(mirror, allInterfaces, minimalInterfaces);
```
schlosna: Should we short circuit?
```suggestion
            if (isInvalid(mirror, allInterfaces, minimalInterfaces)) {
                return true;
            }
```

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessor.java
```diff
@@ -0,0 +1,438 @@
...
+        }
+        inconsistentElements.clear();
+        return currentElements;
+    }
+
+    private boolean getInterfaces(
+            List<? extends TypeMirror> mirrors,
+            List<DeclaredType> allInterfaces,
+            @Nullable List<DeclaredType> minimalInterfaces) {
+        boolean error = false;
```
schlosna:
```suggestion
```

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessor.java
```diff
@@ -0,0 +1,438 @@
...
+        for (TypeMirror mirror : mirrors) {
+            error |= getInterfaces(mirror, allInterfaces, minimalInterfaces);
+        }
+        return error;
```
schlosna:
```suggestion
        return false;
```

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessor.java
```diff
@@ -0,0 +1,438 @@
...
+    private Set<Element> getElementsToProcess(RoundEnvironment env) {
+        Set<Element> currentElements = new HashSet<>(env.getElementsAnnotatedWith(TritiumInstrument.class));
+        for (Name name : inconsistentElements) {
+            currentElements.add(elements.getTypeElement(name));
+        }
+        inconsistentElements.clear();
+        return currentElements;
+    }
+
+    private boolean getInterfaces(
```
schlosna: Naming nit 

```suggestion
    private boolean isInvalid(
```

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessor.java
```diff
@@ -0,0 +1,438 @@
...
+            List<DeclaredType> allInterfaces,
+            @Nullable List<DeclaredType> minimalInterfaces) {
+        boolean error = false;
+        for (TypeMirror mirror : mirrors) {
+            error |= getInterfaces(mirror, allInterfaces, minimalInterfaces);
+        }
+        return error;
+    }
+
+    private boolean getInterfaces(
```
schlosna:
```suggestion
    private boolean isInvalid(
```

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessor.java
```diff
@@ -0,0 +1,438 @@
...
+        }
+        List<DeclaredType> minimalInterfaces = minimalInterfacesFinal;
+        if (mirror.getKind() == TypeKind.DECLARED && types.asElement(mirror).getKind() == ElementKind.INTERFACE) {
+            allInterfaces.add((DeclaredType) mirror);
+            if (minimalInterfaces != null) {
+                minimalInterfaces.add((DeclaredType) mirror);
+                minimalInterfaces = null;
+            }
+        }
+        return getInterfaces(((TypeElement) types.asElement(mirror)).getInterfaces(), allInterfaces, minimalInterfaces);
```
schlosna:
```suggestion
        return isInvalid(((TypeElement) types.asElement(mirror)).getInterfaces(), allInterfaces, minimalInterfaces);
```

versions.lock
```diff
@@ -5,37 +5,45 @@ com.fasterxml.jackson.core:jackson-databind:2.11.3 (3 constraints: 4a3a2999)
 com.fasterxml.jackson.datatype:jackson-datatype-jdk8:2.11.3 (1 constraints: cf0c0316)
 com.fasterxml.jackson.module:jackson-module-afterburner:2.11.3 (1 constraints: cf0c0316)
 com.github.ben-manes.caffeine:caffeine:2.9.0 (1 constraints: 0d051036)
+com.google.auto:auto-common:1.0 (2 constraints: 8b1626fd)
```
schlosna, line 8: Do we expect any downstream issues with this dep?

versions.lock
```diff
@@ -5,37 +5,45 @@ com.fasterxml.jackson.core:jackson-databind:2.11.3 (3 constraints: 4a3a2999)
...
-com.google.guava:guava:30.1.1-jre (2 constraints: 8814c38e)
+com.google.guava:guava:30.1.1-jre (5 constraints: cb427e81)
 com.google.guava:listenablefuture:9999.0-empty-to-avoid-conflict-with-guava (1 constraints: bd17c918)
 com.google.j2objc:j2objc-annotations:1.3 (1 constraints: b809eda0)
 com.palantir.jvm.diagnostics:jvm-diagnostics:0.0.2 (1 constraints: 0405ed35)
 com.palantir.safe-logging:preconditions:1.14.0 (2 constraints: da117b05)
 com.palantir.safe-logging:safe-logging:1.14.0 (3 constraints: 0b23aa52)
 com.palantir.tracing:tracing:4.22.0 (1 constraints: 3a05453b)
 com.palantir.tracing:tracing-api:4.22.0 (2 constraints: 0912df17)
+com.squareup:javapoet:1.13.0 (1 constraints: 3705323b)
```
schlosna, line 21: Is this really just a compile time dependency for the annotation processor?

schlosna (approved):
Overall looks good, a few questions/comments

tritium-annotations/src/main/java/com/palantir/tritium/annotations/TritiumInstrument.java
carterkozak: Good call

tritium-lib/src/main/java/com/palantir/tritium/proxy/ByteBuddyInstrumentation.java
carterkozak: The logger is encapsulated within the new `Handlers` utility class. It was used when a pre/onSuccess/onFailure handler interaction threw an exception, but the generated bytecode isn't responsible for that anymore.

tritium-processor/build.gradle
carterkozak: I don't believe we can use java-library dependency types within annotation processor implementations, they're not correctly added to the annotationProcessor class path for dependent projects, but it's hard to catch in same-project tests. I've broken several annotation processors this way!

versions.lock
carterkozak: That's correct, it's considered a 'main' dependency because the annotation processor (main sources) depend on it, but it won't be pulled into main dependencies of projects which use the annotation processor.

tritium-processor/build.gradle
schlosna: interesting, will the `compile` scope work in Gradle 7? I think tritium is still on 6.9 for now

carterkozak: That’s a good point, I’m not sure.

versions.lock
carterkozak: We've used it several places including the dialogue annotation processor, I don't expect issues, but worst case we can hand-write the META-INF/services file

tritium-processor/build.gradle
carterkozak: @CRogers or @fawind Any idea about Gradle 7 and annotation processor dependencies?

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessor.java
carterkozak: 👍

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessor.java
carterkozak: 👍

versions.lock
carterkozak: Added a comment in build.gradle -- this dependency is used to add the correct `@Generated` annotation based on the target java version.

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessor.java
carterkozak: Good call, updated

tritium-processor/build.gradle
fawind: Yes, `compile` and `runtime` configurations got removed with Gradle 7 ([ref](https://docs.gradle.org/current/userguide/upgrading_version_6.html#sec:configuration_removal)). I would have assumed that at least the `api` scope should have similar behavior to `compile` but not very familiar with annotation processor wiring. If this doesn't work, it sounds like something we should potentially flag to Gradle upstream?

carterkozak: Ya, I seem to recall that API didn't work, but I could be mistaken. We can take a deeper in the next couple weeks as we'll need to solve it across all our annotation processors.

carterkozak:
Let's try it out, easy enough to modify the new module as it won't have many users for a while. I'll plan to merge shortly unless folks would prefer that I wait.

We can figure out the annotation processor/gradle7 coordination separately since we'll have to update all our annotation processors in a similar way.

PR #1085, opened 2021-05-13
Enable gradle incremental annotation processing for tritium-processor

carterkozak:
Based on https://github.com/palantir/dialogue/pull/1231

I opted against using annotation processors for simplicity. I
had minor issues with intellij, and generally find META-INF files
easier to reason about that annotation processors which write
one-line files.

Documentation for incremental processing can be found here:
https://docs.gradle.org/current/userguide/java_plugin.html#sec:incremental_annotation_processing

==COMMIT_MSG==
Enable gradle incremental annotation processing for tritium-processor
==COMMIT_MSG==

## Possible downsides?
It's possible that this feature may cause issues if we attempt to
instrument types that are genereated from other annotation processors,
however that is somewhat unlikely, and the docs describe that
as a valid case for `isolating` configuration.

### DIFF
--- a/changelog/@unreleased/pr-1085.v2.yml
+++ b/changelog/@unreleased/pr-1085.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Enable gradle incremental annotation processing for tritium-processor
+  links:
+  - https://github.com/palantir/tritium/pull/1085
--- a/tritium-processor/build.gradle
+++ b/tritium-processor/build.gradle
@@ -14,9 +14,6 @@ dependencies {
     compile 'com.google.code.findbugs:jsr305'
     compile 'com.palantir.safe-logging:preconditions'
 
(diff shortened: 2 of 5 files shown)


pkoenig10 (commented):
The generated file needs to have a single originating element for incremental compilation to work. See https://github.com/palantir/dialogue/pull/1240.

It looks like we don't specify an originating element when building the `TypeSpec`.

Given that we're relying on the implementation of JavaPoet for this, might be a good idea to just add an assertion that the `TypeSpec` in the `JavaFile` has exactly 1 originating element.

pkoenig10:
> I opted against using annotation processors for simplicity. I
had minor issues with intellij, and generally find META-INF files
easier to reason about that annotation processors which write
one-line files.

Agreed, we should just change Dialogue.

https://github.com/palantir/dialogue/pull/1255

carterkozak:
Good catch, updated with an originating element.

pkoenig10 (approved):

PR #1088, opened 2021-05-16
JMH gradle plugin 0.6.4

schlosna:
## Before this PR
Excavator gradle plugin bumps stopped for the JMH gradle plugin due to migration from gradle plugin ID `me.champeau.gradle.jmh` to `me.champeau.jmh` in [version 0.6.0](https://github.com/melix/jmh-gradle-plugin/blob/master/CHANGELOG.txt#L18).

## After this PR

==COMMIT_MSG==
JMH gradle plugin 0.6.4
==COMMIT_MSG==


carterkozak (approved):
lgtm

PR #1104, opened 2021-06-09
[WIP] Implement a dynamic-proxy annotation processor for closed-world compatibility

carterkozak:
This is a quick sketch largely copied from the tritium processor. Looking for early feedback before I clean it up.
Primary goal is to make atlasdb code safe to consume from native applications without maintaining a list of proxyable classes.

## Before this PR
Difficult to support arbitrary dynamic proxies in a graal native world.

## After this PR
==COMMIT_MSG==
Implement a dynamic-proxy annotation processor to support `InvocationHandler` using build-time proxies that support graalvm native images out of the box.
==COMMIT_MSG==

## Possible downsides?
I'm skeptical that `InvocationHandler` drop-in replacement is
the ideal solution. It will certainly unblock us with
minimal code churn, however in a perfect world we would provide
a clean API that can encapsulate the delegate invocation cleanly,
and provide metadata without necessarily exposing reflective types.
Unfortunately it would be time consuming to get such an API right,
and it's unlikely we'd get everything correct on the first attempt.
In fact it would likely involve several APIs tailored for specific use
cases...


tritium-annotations/src/main/java/com/palantir/tritium/annotations/Proxy.java
```diff
@@ -0,0 +1,35 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Interfaces annotated with {@link Proxy} will result in a generated concrete implementation
+ * using the same package, and name prefixed with {@code Proxied}.
+ * <p/>
+ * Externally defined interfaces may be proxied by defining an interface with the requested interface
+ * as a supertype. The generator will produce an instrumentation class which can wrap the super-interface.
+ * When type parameters are present, the empty interface must define and pass along type parameters which
+ * exactly match the target interface.
```
carterkozak: unnecessary, the proxy can implement a wrapper without impacting results.
```suggestion
```

tritium-processor/src/main/java/com/palantir/tritium/processor/ProxyAnnotationProcessor.java
```diff
@@ -0,0 +1,428 @@
...
+        Map<String, List<MethodElements>> methodsByName = new HashMap<>();
+        for (DeclaredType mirror : allInterfaces) {
+            for (Element methodElement : types.asElement(mirror).getEnclosedElements()) {
+                if (!methodElement.getModifiers().contains(Modifier.STATIC)
+                        && !methodElement.getModifiers().contains(Modifier.PRIVATE)) {
+                    methodElement.accept(
+                            new SimpleElementVisitor8<Void, Void>() {
+                                @Override
+                                public Void visitExecutable(ExecutableElement method, Void _param) {
+                                    if (!Methods.isObjectMethod(elements, method)) {
```
carterkozak: These should still be proxied to match dynamic proxy behavior

tritium-processor/src/main/java/com/palantir/tritium/processor/ProxyAnnotationProcessor.java
```diff
@@ -0,0 +1,428 @@
...
+                    typeElement);
+        }
+        IdentityHashMap<MethodElements, String> methodStaticFields =
+                Methods.methodStaticFieldName(instrumentedMethods, specBuilder, annotatedType);
+        for (MethodElements method : instrumentedMethods) {
+            createMethod(method, specBuilder, methodStaticFields);
+        }
+
+        specBuilder
+                .addMethod(MethodSpec.methodBuilder("toString")
```
carterkozak: delete, handlers may choose to implement toString, or to proxy it.

tritium-processor/src/main/java/com/palantir/tritium/processor/ProxyAnnotationProcessor.java
```diff
@@ -0,0 +1,426 @@
...
+                continue;
+            }
+            TypeElement typeElement = (TypeElement) element;
+            List<DeclaredType> allInterfaces = new ArrayList<>();
+            List<DeclaredType> minimalInterfaces = new ArrayList<>();
+            if (isInvalid(element.asType(), allInterfaces, minimalInterfaces)) {
+                invalidElements.add(typeElement.getQualifiedName());
+                continue;
+            }
+            if (allInterfaces.isEmpty()) {
```
schlosna: What populates `allInterfaces` so that it is not always empty?

tritium-processor/src/main/java/com/palantir/tritium/processor/ProxyAnnotationProcessor.java
```diff
@@ -0,0 +1,426 @@
...
+                messager.printMessage(
+                        Kind.ERROR,
+                        "Cannot generate a instrumented implementation. "
```
schlosna:
```suggestion
                        "Cannot generate an instrumented implementation. "
```

tritium-processor/src/main/java/com/palantir/tritium/processor/ProxyAnnotationProcessor.java
schlosna: Never mind, looks like `isValid` adds to `allInterfaces`

tritium-processor/src/main/java/com/palantir/tritium/processor/ProxyAnnotationProcessor.java
```diff
@@ -0,0 +1,426 @@
...
+        boolean isVoidMethod = method.type().getReturnType().getKind() == TypeKind.VOID;
+        String parameterString = method.element().getParameters().stream()
+                .map(param -> param.getSimpleName().toString())
+                .collect(Collectors.joining(", "));
+
+        ImmutableSet<String> parameters = method.element().getParameters().stream()
+                .map(param -> param.getSimpleName().toString())
+                .collect(ImmutableSet.toImmutableSet());
+        String throwableName = Parameters.disambiguate("throwable", parameters);
+        String rethrownName = Parameters.disambiguate("throwable", parameters);
```
schlosna: Should this string be `"rethrown"`?

```suggestion
        String rethrownName = Parameters.disambiguate("rethrown", parameters);
```

schlosna (commented):
Is the intent to proxy external types, so if for example one wanted to proxy `Runnable` you would define a new interface type and extend Runnable like done in https://github.com/palantir/tritium/blob/96a55407ea147a08725c26efbd717216ad496afc/tritium-processor/src/test/java/com/palantir/tritium/examples/DelegateToRunnable.java ?

tritium-processor/src/test/resources/com/palantir/tritium/examples/ProxiedDelegateToRunnable.java.generated
```diff
@@ -0,0 +1,73 @@
...
+            void.class.cast(returned);
+        } catch (Error | RuntimeException throwable) {
+            throw throwable;
+        } catch (Throwable throwable) {
+            throw new UndeclaredThrowableException(throwable);
+        }
+    }
+
+    public static DelegateToRunnable of(InvocationHandler handler) {
+        return new ProxiedDelegateToRunnable(handler);
```
schlosna: Do we want to generate `handler != null` precondition check for the constructor or factory?

tritium-processor/src/test/resources/com/palantir/tritium/examples/ProxiedBindsParameter.java.generated
```diff
@@ -0,0 +1,74 @@
...
+            Object returned = this.handler.invoke(this, TO_STRING, new Object[]{});
+            return (String) returned;
+        } catch (Error | RuntimeException throwable) {
+            throw throwable;
+        } catch (Throwable throwable) {
+            throw new UndeclaredThrowableException(throwable);
+        }
+    }
+
+    public static BindsParameter of(InvocationHandler handler) {
```
jkozlowski: So the flow would be:

1. Annotate your MyService interface with @Proxy.
2. In your code call ProxiedMyService.of(InvocationHandler).
3. Profit!

So this would replace all the usages of java.lang.reflect.Proxy.newProxyInstance. AbstractInvocationHandler usages should JUST work? https://github.com/google/guava/blob/master/guava/src/com/google/common/reflect/AbstractInvocationHandler.java#L44

tritium-processor/src/test/resources/com/palantir/tritium/examples/ProxiedBindsParameter.java.generated
```diff
@@ -0,0 +1,74 @@
...
+
+    private final InvocationHandler handler;
+
+    private ProxiedBindsParameter(InvocationHandler handler) {
+        this.handler = handler;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public boolean equals(Object obj) {
```
jkozlowski: hashCode?

tritium-processor/src/test/resources/com/palantir/tritium/examples/ProxiedBindsParameter.java.generated
```diff
@@ -0,0 +1,74 @@
...
+public final class ProxiedBindsParameter implements BindsParameter {
+    private static final Method EQUALS;
+
+    private static final Method GET;
+
+    private static final Method TO_STRING;
+
+    static {
+        try {
+            EQUALS = BindsParameter.class.getMethod("equals", Object.class);
```
jkozlowski: I'm assuming the way this works, is that graal will runs this at build time, and assign static method handles?

tritium-processor/src/test/java/com/palantir/tritium/examples/DelegateToRunnable.java
```diff
@@ -17,6 +17,8 @@
 package com.palantir.tritium.examples;
 
 import com.palantir.tritium.annotations.Instrument;
+import com.palantir.tritium.annotations.Proxy;
 
+@Proxy
 @Instrument
 public interface DelegateToRunnable extends Runnable {}
```
jkozlowski: Can you add a test for multiple interfaces being extended? *I think* you have a lot of logic to resolve any overrides etc. might be useful to put that through it's paces.

tritium-processor/src/main/java/com/palantir/tritium/processor/ProxyAnnotationProcessor.java
```diff
@@ -0,0 +1,426 @@
...
+                .addField(FieldSpec.builder(InvocationHandler.class, HANDLER_NAME, Modifier.PRIVATE, Modifier.FINAL)
+                        .build())
+                .addMethod(MethodSpec.constructorBuilder()
+                        .addModifiers(Modifier.PRIVATE)
+                        .addParameter(ParameterSpec.builder(InvocationHandler.class, HANDLER_NAME)
+                                .build())
+                        .addStatement("this.$1N = $1N", HANDLER_NAME)
+                        .build());
+
+        Map<String, List<MethodElements>> methodsByName = new HashMap<>();
```
jkozlowski: Without going through the entirety of this logic, would MoreElements#getAllMethods work?

tritium-processor/src/test/resources/com/palantir/tritium/examples/ProxiedBindsParameter.java.generated
```diff
@@ -0,0 +1,74 @@
+package com.palantir.tritium.examples;
+
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.UndeclaredThrowableException;
+import javax.annotation.processing.Generated;
+
+@Generated("com.palantir.tritium.processor.ProxyAnnotationProcessor")
+public final class ProxiedBindsParameter implements BindsParameter {
```
jkozlowski: Bikeshed: how about "BindsParameterProxy"? "BindsParameterProxyFactory" (because you can never have enough factories)?

tritium-processor/src/main/java/com/palantir/tritium/processor/ProxyAnnotationProcessor.java
```diff
@@ -0,0 +1,426 @@
...
+
+        if (specBuilder.originatingElements.size() != 1) {
+            messager.printMessage(
+                    Kind.ERROR,
+                    "The generated type must have exactly one originating element: " + specBuilder.originatingElements,
+                    typeElement);
+        }
+        return JavaFile.builder(packageName, specBuilder.build())
+                .skipJavaLangImports(true)
+                .indent("    ")
```
jkozlowski: what does this do?

tritium-processor/src/main/java/com/palantir/tritium/processor/ProxyAnnotationProcessor.java
carterkozak: I can remove this block, The structure of this class was copied and slightly adapted from the tritium processor, however in this case we handle object methods that always exist.

carterkozak: Sorry, I was thinking about the methods isEmpty check which can be removed since there are always object methods (e.g. `toString`). I think the `element.getKind() != ElementKind.INTERFACE` check above _should_ make this impossible, however there are edge cases around annotation processor inter-dependence which may be able to trigger this.

tritium-processor/src/main/java/com/palantir/tritium/processor/ProxyAnnotationProcessor.java
carterkozak: I probably need to fix this in the tritium processor as well. I need to audit for the word `instrument` and replace with `proxy`.

tritium-processor/src/main/java/com/palantir/tritium/processor/ProxyAnnotationProcessor.java
carterkozak: oops, good catch

tritium-processor/src/test/resources/com/palantir/tritium/examples/ProxiedDelegateToRunnable.java.generated
carterkozak: :+1: I'll use `Objects.requireNonNull`, that way this processor can be used with no runtime dependencies.

I think it will make the most sense to break this out into a separate annotation processor, mostly because the annotation is not needed at runtime, while the tritium processor (overloaded at this point, InvocationEventHandler processor?) requires a dependency on tritium-lib.

carterkozak:
> Is the intent to proxy external types, so if for example one wanted to proxy Runnable you would define a new interface type and extend Runnable

That's exactly right. The generated method will still return the local interface, however we'd immediately down-cast into the external type.

```java
@Proxy
interface MyRunnable extends Runnable {}

class Main {
  public static void f(InvocationHandler handler) {
     Runnable wrapped = ProxiedMyRunnable.of(handler);
     // This would also compile:
     // MyRunnable wrapped = ProxiedMyRunnable.of(handler);
  }
}
```

tritium-processor/src/test/resources/com/palantir/tritium/examples/ProxiedBindsParameter.java.generated
carterkozak: I like `BindsParameterProxy`, good idea.

tritium-processor/src/test/resources/com/palantir/tritium/examples/ProxiedBindsParameter.java.generated
carterkozak: Not quite, we aren't initializing classes at build time, however compile-time constant reflection calls (in which the target class, method name, and argument classes are all compile-time constants) are automatically replaced by the native-image compiler, so this is safe regardless of how the build is configured. Documentation here: https://www.graalvm.org/reference-manual/native-image/Reflection/#automatic-detection

tritium-processor/src/test/resources/com/palantir/tritium/examples/ProxiedBindsParameter.java.generated
carterkozak: :+1:

tritium-processor/src/test/resources/com/palantir/tritium/examples/ProxiedBindsParameter.java.generated
carterkozak: Yep, that's the plan!

tritium-processor/src/main/java/com/palantir/tritium/processor/ProxyAnnotationProcessor.java
carterkozak: skipJavaLangImports avoids importing things like `java.lang.Object`. Default is to include those imports to avoid ambiguity if there's a type named `Object` defined in the current package, but we  generate code to align with the style guide for the most part.

`.indent("    ")` uses our default indent, I didn't include PJF in this processor becuase it is far more expensive than the analysis/codegen part of annotation processing.

tritium-processor/src/test/resources/com/palantir/tritium/examples/ProxiedBindsParameter.java.generated
jkozlowski: Could we, for V1, detect if we're graaling and build this at compile time?

carterkozak: I'm not sure I understand what you're suggesting.

tritium-processor/src/test/java/com/palantir/tritium/examples/DelegateToRunnable.java
carterkozak: :+1:

tritium-processor/src/main/java/com/palantir/tritium/processor/ProxyAnnotationProcessor.java
carterkozak: I've replaced the visitor with `MoreElements.getLocalAndInheritedMethods`, however that method can return duplicates if they're inherited from two separate superinterfaces, so we still need some of the complexity.

tritium-processor/src/test/resources/com/palantir/tritium/examples/ProxiedBindsParameter.java.generated
jkozlowski: Sorry, I indeed was not clear. I was just wondering if you could initialize any of this at build time, so they’re just constants. Not that it matters, but I think normally statics are initialised at runtime?

carterkozak: Got it, yes, that could be an improvement.

I think I've vastly overcomplicated this, the same automatic detection can work based on constant arguments to `Proxy.newProxyInstance`, so I just need to generated that based on the `Proxy` annotation. :facepalm:

tritium-proxy-processor/src/test/resources/com/palantir/tritium/examples/BindsParameterProxy.java.generated
```diff
@@ -0,0 +1,15 @@
...
+import java.lang.reflect.Proxy;
+import javax.annotation.processing.Generated;
+
+@Generated("com.palantir.tritium.proxy.processor.ProxyAnnotationProcessor")
+public final class BindsParameterProxy {
+    private BindsParameterProxy() {
+    }
+
+    public static BindsParameter of(InvocationHandler handler) {
+        return (BindsParameter) Proxy.newProxyInstance(BindsParameterProxy.class.getClassLoader(), new Class<?>[] {BindsParameter.class}, handler);
```
jkozlowski, line 13: Ok so all you need the processor for is to enforce this little snippet?

carterkozak:
Shipped here: https://github.com/palantir/proxy-processor

PR #1105, opened 2021-06-10
Require JDK 11+ for source & bytecode compatibility

schlosna:
## Before this PR
Source compatibility was JDK 8

## After this PR

==COMMIT_MSG==
Source compatibility JDK 11
==COMMIT_MSG==


carterkozak (approved):
sounds good to me, we use 11 everywhere else at this point.

PR #1143, opened 2021-07-14
tritium-processor outputs are formatted with PJF

carterkozak:
See https://github.com/palantir/goethe

## Before this PR
Code wasn't bad, but didn't strictly align with our coding standards, and wasn't always easy to read.

## After this PR
==COMMIT_MSG==
Tritium annotation processor generated sources are formatted using palantir-java-format
==COMMIT_MSG==

## Possible downsides?
Formatting source code can be relatively expensive. Given we use a formatter on most other code-gen, I figured it would make sense here now that we have a simple library to handle the complexity for us.


versions.props
```diff
@@ -26,7 +26,7 @@ org.mockito:* = 3.11.2
 org.mpierce.metrics.reservoir:* = 1.1.3
 org.openjdk.jmh:* = 1.32
 org.slf4j:* = 1.7.31
-
+com.palantir.goethe:* = 0.2.0
```
schlosna: nit: can you `:1,29!sort -u` this list?

carterkozak: Updated

schlosna (approved):

versions.props
schlosna: Thanks!

PR #1169, opened 2021-07-27
Manually run the SafeLogger migration

carterkozak:
I manually removed a few slf4j-api `implementation` dependencies
and sorted the `com.palantir.safe-logging:logger` dependency.

A future PR should consider a new tritium-logging module which
may replace tritium-slf4j with safelogging loggers.

==COMMIT_MSG==
Migrate most slf4j interactions to SafeLogger
==COMMIT_MSG==

## Possible downsides?
This only includes automated changes to java files, so there is still some slf4j-usage due to `tritium-slf4j` which may introduce a "lava layer" anti-pattern if we're not careful.


carterkozak:
oh huh, this time it worked. I'll iterate on the excavator PR instead.

PR #1170, opened 2021-07-27
Enforce 'PreferSafeLogger'

carterkozak:
PreferSafeLogger is defined in gradle-baseline and leveraged by
the excavator automation (see #1168). We plan to begin enforcing
this check in the coming weeks/months, however it can be enabled
earlier in projects that have taken the migration.

==COMMIT_MSG==
Enforce 'PreferSafeLogger'
==COMMIT_MSG==


schlosna (commented):
sweet, thanks @carterkozak !

schlosna (approved):

Issue #1192, opened 2021-08-15
JavaDoc generation fails on some systems


schlosna:
## What happened?

JavaDoc is failing to generate on system using default US-ASCII encoding
```
> Task :tritium-registry:javadoc FAILED
/Users/davids/Documents/dev/tritium/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java:114: error: unmappable character (0xE2) for encoding US-ASCII
        /* "A common feature of the above techniques???indeed, the key technique that
                                                    ^
/Users/davids/Documents/dev/tritium/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java:114: error: unmappable character (0x80) for encoding US-ASCII
        /* "A common feature of the above techniques???indeed, the key technique that
                                                     ^
/Users/davids/Documents/dev/tritium/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java:114: error: unmappable character (0x94) for encoding US-ASCII
        /* "A common feature of the above techniques???indeed, the key technique that
                                                      ^
/Users/davids/Documents/dev/tritium/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java:115: error: unmappable character (0xE2) for encoding US-ASCII
         * allows us to track the decayed weights efficiently???is that they maintain
                                                             ^
/Users/davids/Documents/dev/tritium/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java:115: error: unmappable character (0x80) for encoding US-ASCII
         * allows us to track the decayed weights efficiently???is that they maintain
                                                              ^
/Users/davids/Documents/dev/tritium/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java:115: error: unmappable character (0x94) for encoding US-ASCII
         * allows us to track the decayed weights efficiently???is that they maintain
                                                               ^
/Users/davids/Documents/dev/tritium/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java:116: error: unmappable character (0xE2) for encoding US-ASCII
[...]

PR #1193, opened 2021-08-15
JavaDoc uses UTF-8 encoding

schlosna:
## Before this PR

https://github.com/palantir/tritium/issues/1192 JavaDoc is failing to generate on system using default US-ASCII encoding

## After this PR

==COMMIT_MSG==
JavaDoc uses UTF-8 encoding
==COMMIT_MSG==

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -112,6 +112,7 @@ allprojects {
     }
 
     tasks.withType(Javadoc) {
+        options.addStringOption('encoding', 'UTF-8')
         // suppress Javadoc doclint warnings in Java 8+
         if (!System.getProperty("java.version").startsWith("1.7")) {
             options.addStringOption('Xdoclint:none', '-quiet')


build.gradle
```diff
@@ -112,6 +112,7 @@ allprojects {
     }
 
     tasks.withType(Javadoc) {
+        options.addStringOption('encoding', 'UTF-8')
```
carterkozak, line 115: Let’s add this to our baseline java encoding plugin which does something similar for javac

schlosna: Ahh, didn't know baseline did that. I assume we can add for JavaDoc task in https://github.com/palantir/gradle-baseline/blob/0f716d5ec6063b4dae06b5f761f252178aac3941/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEncoding.java#L28

carterkozak: Yep!

schlosna: Created https://github.com/palantir/gradle-baseline/pull/1879

schlosna:
Closing in favor of fixing in gradle-baseline

PR #1230, opened 2021-09-27
Remove unnecessary executor metrics

carterkozak:
*`executor.completed` is identical to the `count` value of the `executor.duration` timer. Dashboards will need to replace the former with the latter.
* `executor.scheduled.percent-of-period` is unused, our pipeline has filtered out the metric for a while now to reduce bloat.
* Scheduled executors no longer produce `executor.scheduled.once`, `executor.scheduled.repetitively`, or `executor.submitted` because the data isn't meaningful when load on the executor may be based on recurring tasks.
==COMMIT_MSG==
Remove unnecessary executor metrics. 
==COMMIT_MSG==

## Possible downsides?
This is a metrics _break_ which shouldn't be taken lightly.

### DIFF
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsScheduledExecutorService.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsScheduledExecutorService.java
@@ -36,7 +36,6 @@ final class TaggedMetricsScheduledExecutorService extends AbstractExecutorServic
     private final Counter running;
     private final Timer duration;
 
-    private final Meter scheduledRepetitively;
     private final Counter scheduledOverrun;
 
     TaggedMetricsScheduledExecutorService(ScheduledExecutorService delegate, ExecutorMetrics metrics, String name) {
@@ -47,7 +46,6 @@ final class TaggedMetricsScheduledExecutorService extends AbstractExecutorServic
         this.running = metrics.running(name);
         this.duration = metrics.duration(name);
 
-        this.scheduledRepetitively = metrics.scheduledRepetitively(name);
         this.scheduledOverrun = metrics.scheduledOverrun(name);
     }
 
@@ -75,15 +73,26 @@ final class TaggedMetricsScheduledExecutorService extends AbstractExecutorServic
 
     @Override
     public ScheduledFuture<?> scheduleAtFixedRate(Runnable task, long initialDelay, long period, TimeUnit unit) {
-        scheduledRepetitively.mark();
-        return delegate.scheduleAtFixedRate(
+        ScheduledFuture<?> future = delegate.scheduleAtFixedRate(
                 new TaggedMetricsScheduledRunnable(task, period, unit), initialDelay, period, unit);
+        // RejectedExecutionException should prevent 'submitted' from being incremented.
+        // This means a wrapped same-thread executor will produce delayed 'submitted' values,
+        // however the results will work as expected for the more common cases in which
+        // either a queue is full, or the delegate has shut down.
+        submitted.mark();
+        return future;
     }
 
     @Override
     public ScheduledFuture<?> scheduleWithFixedDelay(Runnable task, long initialDelay, long delay, TimeUnit unit) {
-        scheduledRepetitively.mark();
-        return delegate.scheduleWithFixedDelay(new TaggedMetricsRunnable(task), initialDelay, delay, unit);
+        ScheduledFuture<?> future =
+                delegate.scheduleWithFixedDelay(new TaggedMetricsRunnable(task), initialDelay, delay, unit);
+        // RejectedExecutionException should prevent 'submitted' from being incremented.
+        // This means a wrapped same-thread executor will produce delayed 'submitted' values,
+        // however the results will work as expected for the more common cases in which
+        // either a queue is full, or the delegate has shut down.
+        submitted.mark();
+        return future;
     }
 
     @Override
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsScheduledExecutorService.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsScheduledExecutorService.java
@@ -17,7 +17,6 @@
 package com.palantir.tritium.metrics;
 
 import com.codahale.metrics.Counter;
-import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.Timer;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


carterkozak:
Adding do-not-merge until this has been thoroughly reviewed and considered.

tritium-metrics/src/main/metrics/metrics.yml
```diff
@@ -64,10 +64,6 @@ namespaces:
         type: counter
         tags: [executor]
         docs: A gauge of the number of running tasks.
```
carterkozak: not a gauge
```suggestion
        docs: The number of running tasks.
```

tritium-metrics/src/main/metrics/metrics.yml
```diff
@@ -76,10 +72,6 @@ namespaces:
         type: timer
         tags: [executor]
         docs: A timer of the time it took a task to start running after it was submitted.
-      scheduled.once:
-        type: meter
-        tags: [executor]
-        docs: A meter of the number of one-shot scheduled tasks. Applies only to scheduled executors.
       scheduled.repetitively:
         type: meter
         tags: [executor]
```
carterkozak: I'm not sure what this metric tells us, while we're breaking metrics, should this be removed? We could use `submitted` instead, or not record a metric in this case. Thoughts?

pkoenig10: If we're getting rid of `scheduled.once` we should also get rid of this metric. They're effectively the same metric just for one-shot vs repeated tasks.

pkoenig10: I've never really felt the need to distinguish between the two types of submissions (in my experience, one-shot scheduling is pretty rare) so I think just relying on `submitted` is sufficient.

carterkozak: Might as well squash it into the `submitted` meter the same way as the `scheduled.once` metric I suppose?

carterkozak: :+1:

schlosna: This is marked by calls to `scheduleAtFixedRate` & `scheduleWithFixedDelay`.  I'm not sure how folks are making use of this, especially as a `meter` (could probably be a `counter` if we want to keep it). I guess one might want to track recurring task submission rates. `scheduleAtFixedRate` calls are wrapped with a `TaggedMetricsScheduledRunnable` that tracks `scheduledOverrun`, but we don't currently have anything here tracking tasks recurring tasks that have thrown and would not be rescheduled. My sense is the recurring tasks that have thrown and not been rescheduled might be of most value here.

@pkoenig10 may have thoughts from usage and when this was added in #231

schlosna: ahh, actually looks like this was directly ported over for compatibility with Dropwizard metrics https://github.com/dropwizard/metrics/blob/cb3c59e4c71b4df761f5809d56e00dd0d82ef7cb/metrics-core/src/main/java/com/codahale/metrics/InstrumentedScheduledExecutorService.java#L92-L101

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsScheduledExecutorService.java
```diff
@@ -35,52 +34,65 @@
...
     @Override
     public ScheduledFuture<?> schedule(Runnable task, long delay, TimeUnit unit) {
-        scheduledOnce.mark();
-        return delegate.schedule(new TaggedMetricsRunnable(task), delay, unit);
+        ScheduledFuture<?> future = delegate.schedule(new TaggedMetricsRunnable(task), delay, unit);
+        // RejectedExecutionException should prevent 'submitted' from being incremented.
+        // This means a wrapped same-thread executor will produce delayed 'submitted' values,
+        // however the results will work as expected for the more common cases in which
+        // either a queue is full, or the delegate has shut down.
+        submitted.mark();
```
pkoenig10: This definitely isn't right.`submitted` should be updated when we submit the task to the executor - not when we schedule the tasks. Otherwise we'll be double counting submissions.

It's also weird to mark it here because it diverges from the meaning of `submitted` in non-scheduled executors.

pkoenig10 (dismissed):

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsScheduledExecutorService.java
carterkozak: I'm not sure I follow, this matches how we mark `submitted` in non-scheduled executors. If I schedule a runnable to execute in 1 minute, the `submitted` metric works exactly like an instrumented fixed thread pool executor where the task sits in a queue for 1 minute.

> Otherwise we'll be double counting submissions.

This would cause tests to fail unless I've misunderstood your concern

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsScheduledExecutorService.java
```diff
@@ -35,52 +34,65 @@
...
     public ScheduledFuture<?> scheduleAtFixedRate(Runnable task, long initialDelay, long period, TimeUnit unit) {
-        scheduledRepetitively.mark();
-        return delegate.scheduleAtFixedRate(
+        ScheduledFuture<?> future = delegate.scheduleAtFixedRate(
                 new TaggedMetricsScheduledRunnable(task, period, unit), initialDelay, period, unit);
+        // RejectedExecutionException should prevent 'submitted' from being incremented.
+        // This means a wrapped same-thread executor will produce delayed 'submitted' values,
+        // however the results will work as expected for the more common cases in which
+        // either a queue is full, or the delegate has shut down.
+        submitted.mark();
```
pkoenig10: It's really confusing to use `submitted` for repetitive tasks. I typically view `submitted` as "how many tasks were submitted to the executor" and not "how many times was a executor method called".

carterkozak: In this case a task is being submitted to the executor. It's not quite the same as a one-shot task in that it may be executed 0-N times rather than 0-1.
I could see an argument for excluding the metric entirely on recurring tasks, but I don't have a strong preference.

pkoenig10: Yeah I'd probably just exclude it entirely. I think the metric is significantly less useful when it's combined like this. I generally use `submitted` as a proxy for "how much work is being requested" and that doesn't really work when a single submitted task will be executed N times.

carterkozak: Ya that's fair, at that point we might as well remove the submitted metric entirely from the scheduled executor because there's tricky interplay between recurring tasks and `execute`/`submit` which invalidates most of our assumptions anyhow.

carterkozak:
I've updated to remove the `submitted` metric entirely from ScheduledExecutorService instrumentation because the usage patterns are such that the metric isn't very meaningful.

I've also removed the `executor.thread.finalized` meter because it's not terribly useful and can be estimated by subtracting the active counter from the created meter.

pkoenig10 (approved):

tpetracca (approved):

carterkozak:
Updated with another small optimization: The queuedDuration metric is only updated when tasks take longer than 250ms from the time they're submitted to the time they begin executing. Previously the queued duration metric was updated even when tasks were not queued at all, but with a relatively small value.

schlosna (approved):

carterkozak:
Thanks, everyone!

PR #1259, opened 2021-11-05
Test Tritium on Java 17 (only)

carterkozak:
## Before this PR
Previously we tested on both Java 11 and Java 15.

I've got a note on my backlog to improve the baseline-java-versions
plugin such that arbitrarily many library test targets may be
provided, but haven't gotten around to it yet.

## After this PR
==COMMIT_MSG==
Test Tritium on Java 17 
==COMMIT_MSG==

## Possible downsides?
Fewer test targets.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,6 +23,12 @@ apply plugin: 'java'
 apply plugin: 'com.palantir.baseline'
 apply plugin: 'com.palantir.consistent-versions'
 apply plugin: 'com.palantir.git-version'
+apply plugin: 'com.palantir.baseline-java-versions'
+
+javaVersions {
+    libraryTarget = 11
+    runtime = 17
+}
 
 allprojects {
     group 'com.palantir.tritium'
@@ -33,11 +39,6 @@ allprojects {
     apply plugin: 'com.palantir.baseline-class-uniqueness'
     apply plugin: 'com.palantir.baseline-exact-dependencies'
     apply plugin: 'com.palantir.java-format'
-
-    java {
-        sourceCompatibility JavaVersion.VERSION_11
-        targetCompatibility JavaVersion.VERSION_11
-    }
 
     configurations {
         testImplementationClasspath {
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -123,6 +124,10 @@ allprojects {
         // avoid spurious flags on dropwizard metrics
         ignore 'io.dropwizard.metrics', 'metrics-core'
     }
+    checkUnusedDependencies {
+        // avoid spurious flags on junit-jupiter
+        ignore 'org.junit.jupiter', 'junit-jupiter'
```
carterkozak, line 129: I'm not sure what happened here

carterkozak:
@schlosna What do you think about testing exclusively on Java 17?

build.gradle
```diff
@@ -23,6 +23,12 @@ apply plugin: 'java'
 apply plugin: 'com.palantir.baseline'
 apply plugin: 'com.palantir.consistent-versions'
 apply plugin: 'com.palantir.git-version'
+apply plugin: 'com.palantir.baseline-java-versions'
+
+javaVersions {
+    libraryTarget = 11
+    runtime = 17
```
schlosna, line 30: Will this use JDK 17 for compilation & test running, use JDK 11 source level language features, and generate JDK 11 compatible byte code (version 55.0), is that correct? Does it use JDK 11 or 17 ABI?

carterkozak: We're actually using the full toolchain at the requested language level for compilation, so in this case gradle is downloading a JDK11 which is used for compilation (in practice it uses the detected jdk11 on the CI docker image) and downloading a JDK17 to execute tests with the compiled artifact.

in cjr-api I ended up bumping the compile target because we'd been producing Java 8 bytecode (52) while relying on Java 11 dependencies, this solution detects when incompatible bytecode is present in dependencies. https://github.com/palantir/conjure-java-runtime-api/pull/714

schlosna: Ahh nice, best of both worlds

schlosna (approved):
Looks good, let's test this out

PR #1294, opened 2021-12-08
Upgrade dropwizard-metrics to 4.x

carterkozak:
Internal consumers blocked on 3.x have upgraded

## Before this PR
3.2.6

## After this PR
==COMMIT_MSG==
Upgrade dropwizard-metrics to 4.x
==COMMIT_MSG==

## Possible downsides?
Possible others that I'm not aware of still require 3.x?


versions.props
```diff
@@ -30,5 +30,5 @@ org.slf4j:* = 1.7.32
 
 # dependency-upgrader:OFF
 # Explicitly maintaining back-compat with dropwizard metrics
-io.dropwizard.metrics:* = 3.2.6
+io.dropwizard.metrics:* = 4.1.1
```
schlosna, line 33: Can we recommend latest 4.1.x? Can we update the comment if we have to pin minimum version to 4.1.1 due to Spark (seems like https://github.com/palantir/spark/blob/6e3520d7625749ecc20a4892282ac75950fb1cfa/pom.xml#L157 )? Are we assuming non-Spark down streams will pull latest 4.1.x ([currently 4.1.26](https://github.com/dropwizard/metrics/releases)) since we're not taking auto-bumps in tritium?


```suggestion
io.dropwizard.metrics:* = 4.1.26
```

schlosna: On the Spark side, we may want to consider backporting a few commits (and possibly bump to `4.2.4`):

- https://issues.apache.org/jira/browse/SPARK-35487 : https://github.com/apache/spark/commit/6bd6e46aeccf2f0706461dd5a66ecb53b5ec38ef
- https://issues.apache.org/jira/browse/SPARK-36013 : https://github.com/apache/spark/commit/7f703509297fb5b6984a99e6f3bb96b77ca74b10

schlosna (commented):
Happy to get this in. 

Looking at all the relevant Spark consumers they seem to be on 4.1.x now. Had a couple questions if we want to use this hold higher minimum version bar.

versions.props
carterkozak: I'd certainly prefer 4.2, especially if we can use the upstreamed lock-free reservoir (though removing ours would be another break). I'm not entirely sure what the constraints are for 4.1.1, but it's likely the safest option in the 4.x line. That said, we can probably use the latest 4.2, but we may need to be conscious of abi changes between 4.1.1 and 4.2.x while spark versions are upgrading.

versions.props
```diff
@@ -30,5 +30,5 @@ org.slf4j:* = 1.7.32
 
 # dependency-upgrader:OFF
 # Explicitly maintaining back-compat with dropwizard metrics
```
schlosna:
```suggestion
# Explicitly pinning to minimum dropwizard metrics version defined by Apache Spark
```

versions.props
schlosna: Cool, let's leave this as `4.1.1` for now and we can bump to `4.1.26` and `4.2.4` in downstream consumers & test. We can slow roll tritium as I don't expect we'll need any new Dropwizard APIs (except things upstreamed like the lock-free reservoir).

schlosna (approved):

PR #1365, opened 2022-02-25
Optimize MetricName creation

schlosna:
## Before this PR
MetricName creation can be expensive

## After this PR

==COMMIT_MSG==
Optimize MetricName creation

RealMetricName uses pre-hashed code in equals.
ExtraEntrySortedMap avoids allocation in constructor.
Cleanup warnings in ExtraEntrySortedMap.

Add MetricName#withExtraTag(String, String).
We have cases where we want to append an additional tag to a MetricName
and would benefit from not having to pay the overhead of the MetricName
builder copy, so we can optimize this path to use RealMetricName's
access to ExtraEntrySortedMap.
==COMMIT_MSG==

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
@@ -34,6 +34,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.SortedMap;
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 @SuppressWarnings("JdkObsolete")
@@ -59,6 +60,7 @@ final class ExtraEntrySortedMap<K, V> extends AbstractMap<K, V> implements Sorte
         return base.comparator();
     }
 
+    @Nonnull
     @Override
     public SortedMap<K, V> subMap(K fromKey, K toKey) {
         SortedMap<K, V> newBase = base.subMap(fromKey, toKey);
@@ -68,6 +70,7 @@ final class ExtraEntrySortedMap<K, V> extends AbstractMap<K, V> implements Sorte
         return newBase;
     }
 
+    @Nonnull
     @Override
     public SortedMap<K, V> headMap(K toKey) {
         SortedMap<K, V> newBase = base.headMap(toKey);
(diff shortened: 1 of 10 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
```diff
@@ -48,7 +50,7 @@
         this.extraKey = checkNotNull(extraKey, "extraKey");
         this.extraValue = checkNotNull(extraValue, "extraValue");
         this.ordering = Ordering.from(base.comparator());
-        this.extraEntryHashCode = Maps.immutableEntry(extraKey, extraValue).hashCode();
```
schlosna, line 51: this allocation was popping in some JFR samples

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/MetricName.java
```diff
@@ -39,6 +39,17 @@
...
 
+    /**
+     * Creates a metric name with the specified extra tag name and value entry.
+     * <p>All tags and keys must be {@link Safe} to log.
+     * @param extraTagName extra tag key name
+     * @param extraTagValue extra tag value
+     * @return metric name including keys
+     */
+    default MetricName withExtraTag(@Safe String extraTagName, @Safe String extraTagValue) {
+        return RealMetricName.create(this, extraTagName, extraTagValue);
```
schlosna, line 50: The primary consumer of this will likely be [Dialogue's `VersionedTaggedMetricRegistry#augment`](https://github.com/palantir/dialogue/blob/a7cc75d8203de4189ad10a95c59cf8d187f5dc23/dialogue-core/src/main/java/com/palantir/dialogue/core/VersionedTaggedMetricRegistry.java#L51-L56)

A couple JFR snippets covering Dialogue <> Tritium:

Memory:
image


CPU:
image
image

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExtraEntrySortedMap.java
schlosna: handled in #1366

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/MetricName.java
schlosna: Dialogue case handled by https://github.com/palantir/dialogue/pull/1609

schlosna:
Putting this on hold before adding to MetricName API

schlosna:
working into #1368

PR #1367, opened 2022-02-28
TagMap uses ImmutableList

schlosna:
## Before this PR
RealMetricName and TagMap weren't using error-prone `Immutable` checker (though were effectively immutable)

## After this PR

==COMMIT_MSG==
TagMap uses ImmutableList
==COMMIT_MSG==

## Possible downsides?
There's some potential overhead in using `ImmutableList<String>` over raw `String[]`

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
@@ -56,14 +56,14 @@ final class TagMap implements SortedMap<String, String> {
         if (data instanceof TagMap) {
             return (TagMap) data;
         }
-        return new TagMap(toValues(data));
+        return new TagMap(toSortedKeyValues(data));
     }
 
     private TagMap(ImmutableList<String> values) {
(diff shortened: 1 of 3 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
```diff
@@ -41,55 +45,49 @@
...
+@Immutable
 @SuppressWarnings("JdkObsolete")
 final class TagMap implements SortedMap<String, String> {
+    static final TagMap EMPTY = TagMap.of(ImmutableMap.of());
 
-    static final TagMap EMPTY = new TagMap(new String[0]);
-
-    private final String[] values;
-    private final int hash;
+    private final ImmutableList<String> values;
```
carterkozak, line 53: It may be worth taking a look at this change with [JOL](https://openjdk.java.net/projects/code-tools/jol/) to compare the difference for an average size tag map.

schlosna:
closing out as superseded by #1368 & #1373

PR #1371, opened 2022-03-01
Run jqwik tests

carterkozak:
Previously the runtime was missing and jqwik tests were ignored.

==COMMIT_MSG==
Run jqwik tests
==COMMIT_MSG==


carterkozak:
Filed https://github.com/palantir/gradle-baseline/issues/2098 for tracking

schlosna (approved):
Thanks Carter!

PR #1376, opened 2022-03-02
ignoreGradleMetadata

schlosna:
## After this PR

==COMMIT_MSG==
ignoreGradleMetadata
==COMMIT_MSG==


build.gradle
```diff
@@ -1,7 +1,9 @@
 buildscript {
     repositories {
         gradlePluginPortal()
```
carterkozak: I think we may want the same thing here, the gradle plugin portal redirects to other repositories (thus the pain when jcenter goes down)
```suggestion
        gradlePluginPortal() {
            metadataSources { mavenPom(); ignoreGradleMetadataRedirection() }
        }
```

carterkozak: Probably worth excavating this out to our oss projects

carterkozak (approved):

PR #1378, opened 2022-03-04
Optimize AbstractTaggedMetricRegistry#getOrAdd for read path

schlosna:
## Before this PR
Some JFR profiles were showing excessive allocations of `com.palantir.tritium.metrics.registry.AbstractTaggedMetricRegistry$$Lambda$124+0x0000000800da2fa8.472991420` via `java.lang.invoke.Invokers$Holder.linkToTargetMethod(Object, Object)` from `AbstractTaggedMetricRegistry.getOrAdd(MetricName, Class, Supplier)`

image

## After this PR

==COMMIT_MSG==
Optimize AbstractTaggedMetricRegistry#getOrAdd for read path

Add MetricsLifecycleBenchmark.
JMH benchmarks produce JSON & JFR outputs
==COMMIT_MSG==

[JMH comparison](https://jmh.morethan.io/)

### DIFF
--- a/tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/MetricsLifecycleBenchmark.java
+++ b/tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/MetricsLifecycleBenchmark.java
@@ -1,0 +1,96 @@
+/*
+ * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.microbenchmarks;
+
+import com.palantir.tritium.metrics.registry.DefaultTaggedMetricRegistry;
+import com.palantir.tritium.metrics.registry.MetricName;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.profile.GCProfiler;
+import org.openjdk.jmh.profile.JavaFlightRecorderProfiler;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+/**
+ * Benchmark to measure the typical lifecycle of utilizing a metric in the registry.
+ * We expect that the effort should be dominated by the metrics underlying calculation,
+ * and minimal overhead for creating and adding metrics to the registry.
+ */
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@Warmup(iterations = 2, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 3, time = 1, timeUnit = TimeUnit.SECONDS)
+@Fork(1)
+@State(Scope.Benchmark)
+@SuppressWarnings({"designforextension", "NullAway", "PreferJavaTimeOverload"})
+public class MetricsLifecycleBenchmark {
+    private static final MetricName baseMetricName =
+            MetricName.builder().safeName("name").putSafeTags("tag", "value").build();
+
+    private final AtomicLong metricCounter = new AtomicLong();
+    private final TaggedMetricRegistry metrics = new DefaultTaggedMetricRegistry();
+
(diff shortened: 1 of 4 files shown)


tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/MetricsLifecycleBenchmark.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Benchmark
+    public void meter() {
+        metrics.meter(meterName).mark();
+    }
+
+    @Benchmark
+    public void timer() {
+        metrics.timer(counterName).update(1_000_000L, TimeUnit.NANOSECONDS);
```
carterkozak:
```suggestion
        metrics.timer(timerName).update(1_000_000L, TimeUnit.NANOSECONDS);
```

schlosna: :sadmac: that would explain why timer benchmarks were out of sync

tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/MetricsLifecycleBenchmark.java
```diff
@@ -0,0 +1,110 @@
...
+    private TaggedMetricRegistry metrics;
+
+    @Setup
+    public void before() {
+        metrics = constructBaseRegistry();
+    }
+
+    @Benchmark
+    public void counter() {
+        metrics.counter(counterName).inc();
```
carterkozak: I think we're better off avoiding interactions with the metric object since that's not directly related to lookup performance
```suggestion
    public Counter counter() {
        return metrics.counter(counterName);
```

tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/MetricsLifecycleBenchmark.java
```diff
@@ -0,0 +1,110 @@
...
+    @Benchmark
+    public void timer() {
+        metrics.timer(counterName).update(1_000_000L, TimeUnit.NANOSECONDS);
+    }
+
+    private static TaggedMetricRegistry constructBaseRegistry() {
+        TaggedMetricRegistry registry = new DefaultTaggedMetricRegistry();
+        for (int i = 0; i < 50; i++) {
+            registry.addMetrics("registry id", Integer.toString(i), constructSubRegistry());
+        }
```
carterkozak: I don't think any of the subregistries are checked when we call `AbstractTaggedMetricRegistry.getOrAdd, only the main concurrenthashmap -- we may want to add metrics directly to the top-level registry

tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/MetricsLifecycleBenchmark.java
schlosna: I was intending to have the benchmark capture more of the typical lookup & metric update so that we could compare optimizations from that perspective as opposed to just micro-optimizing the metric lookup/creation. If that makes sense, I will add comment to that effect in Benchmark Javadoc.

carterkozak: 👍

tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/MetricsLifecycleBenchmark.java
schlosna: moving to single registry

tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/MetricsLifecycleBenchmark.java
schlosna: going to simplify this a bit with probabilistic read/write (95/5?) ratio

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AbstractTaggedMetricRegistry.java
```diff
@@ -214,7 +214,11 @@ public final boolean removeMetrics(String safeTagName, String safeTagValue, Tagg
 
     protected final <T extends Metric> T getOrAdd(
             MetricName metricName, Class<T> metricClass, Supplier<T> metricSupplier) {
-        Metric metric = registry.computeIfAbsent(metricName, _name -> metricSupplier.get());
+        Metric metric = registry.get(metricName);
+        if (metric == null) {
+            // assume optimistic read fast path to avoid lamdba capture and fallback to slow write path
+            metric = registry.computeIfAbsent(metricName, _name -> metricSupplier.get());
```
carterkozak, line 220: In an ideal world, metrics would all be accessed exactly once at application startup. Double-traversal will make that (slightly) more expensive. What if we restructured the DefaultTaggedMetricRegistry to pre-compute several singleton `Function<Name, T extends Metric>` instances which can be reused without fear of allocation? I think that should improve both the hot-access and cold startup.

carterkozak: (Likely overkill, scratching the optimization itch. Defer to your judgement whether or not it's worthwhile)

schlosna:
After running benchmarks across JDK 11, 15 and 17, I don't think this is actually worth modifying `getOrAdd`, so I'm going to close this out.

```
JDK 17.0.1, OpenJDK 64-Bit Server VM, 17.0.1+12-LTS

Before
Benchmark                            Mode  Cnt     Score      Error  Units
MetricsLifecycleBenchmark.counter    avgt    3   689.290 ± 8172.421  ns/op
MetricsLifecycleBenchmark.histogram  avgt    3   828.058 ± 3576.649  ns/op
MetricsLifecycleBenchmark.meter      avgt    3   813.318 ± 3006.805  ns/op
MetricsLifecycleBenchmark.timer      avgt    3  1375.768 ± 3100.411  ns/op

After
Benchmark                            Mode  Cnt     Score      Error  Units
MetricsLifecycleBenchmark.counter    avgt    3   734.545 ± 9357.108  ns/op
MetricsLifecycleBenchmark.histogram  avgt    3   858.349 ± 4133.059  ns/op
MetricsLifecycleBenchmark.meter      avgt    3   851.431 ± 2746.278  ns/op
MetricsLifecycleBenchmark.timer      avgt    3  1523.786 ± 9908.532  ns/op
```

PR #1391, opened 2022-03-17
Port memory metrics to metric-schema for docs and validation

carterkozak:
These metrics were referenced in an internal ticket, but our generated metric documentation did not list them.

Consumers are not impacted. The metric-schema generated metrics are additionally tagged with the tritium library-name and library version.

==COMMIT_MSG==
Port memory metrics (`jvm.memory`) to metric-schema for better documentation and validation
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
@@ -16,16 +16,18 @@
 
 package com.palantir.tritium.metrics.jvm;
 
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.RatioGauge;
 import com.codahale.metrics.jvm.BufferPoolMetricSet;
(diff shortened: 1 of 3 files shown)


tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -126,6 +123,50 @@ private static void registerThreads(InternalJvmMetrics metrics) {
         return threadsByState;
     }
 
+    private static void registerJvmMemory(TaggedMetricRegistry registry) {
+        JvmMemoryMetrics metrics = JvmMemoryMetrics.of(registry);
+        MemoryMXBean memoryBean = ManagementFactory.getMemoryMXBean();
```
schlosna, line 130: Should we need check for null here (though in practice I don't think this will be null) to log & return without later failing due to NPE?

carterkozak: Good idea, updated

schlosna (approved):

PR #1392, opened 2022-03-17
Memory gauges report null when data is unavailable

carterkozak:
Previously the default `-1` values were used, resulting in
confusing dashboards that didn't make it obvious the data was
unavailable.

==COMMIT_MSG==
Memory gauges report null when data is unavailable
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
@@ -20,6 +20,7 @@ import com.codahale.metrics.Gauge;
 import com.codahale.metrics.RatioGauge;
 import com.codahale.metrics.jvm.BufferPoolMetricSet;
 import com.codahale.metrics.jvm.ThreadDeadlockDetector;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Suppliers;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.logger.SafeLogger;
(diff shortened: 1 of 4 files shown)


tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -132,48 +134,76 @@ private static void registerJvmMemory(TaggedMetricRegistry registry) {
...
         // jvm.memory.total
-        metrics.totalInit((Gauge<Long>) () -> memoryBean.getHeapMemoryUsage().getInit()
-                + memoryBean.getNonHeapMemoryUsage().getInit());
-        metrics.totalUsed((Gauge<Long>) () -> memoryBean.getHeapMemoryUsage().getUsed()
-                + memoryBean.getNonHeapMemoryUsage().getUsed());
-        metrics.totalMax((Gauge<Long>) () -> memoryBean.getHeapMemoryUsage().getMax()
-                + memoryBean.getNonHeapMemoryUsage().getMax());
+        metrics.totalInit(
+                (NonNegativeGauge) () -> memoryBean.getHeapMemoryUsage().getInit()
+                        + memoryBean.getNonHeapMemoryUsage().getInit());
```
schlosna: are these computations problematic if one or both are not available?

carterkozak: Probably worth checking both, in practice they’re all or nothing in my experience, but that could change.

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -132,48 +135,75 @@ private static void registerJvmMemory(TaggedMetricRegistry registry) {
...
-                + memoryBean.getNonHeapMemoryUsage().getUsed());
-        metrics.totalMax((Gauge<Long>) () -> memoryBean.getHeapMemoryUsage().getMax()
-                + memoryBean.getNonHeapMemoryUsage().getMax());
-        metrics.totalCommitted(
-                (Gauge<Long>) () -> memoryBean.getHeapMemoryUsage().getCommitted()
-                        + memoryBean.getNonHeapMemoryUsage().getCommitted());
+        metrics.totalInit((Gauge<Long>) () -> {
+            long heap = memoryBean.getHeapMemoryUsage().getInit();
+            long nonHeap = memoryBean.getNonHeapMemoryUsage().getInit();
+            return (heap < 0 || nonHeap < 0) ? null : heap + nonHeap;
```
schlosna: I assume the intent here to avoid producing any value if one don't have just one component (e.g. if we have heap init but do not have non-heap init)?

schlosna: I don't have super strong feelings, but curious for thoughts on https://github.com/palantir/tritium/pull/1393

Looking at spec, it seems like [`getInit()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.management/java/lang/management/MemoryUsage.html#getInit()) & [`getMax()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.management/java/lang/management/MemoryUsage.html#getMax()) may return -1 when undefined, but used & committed should be non-negative though I guess we should handle similarly in case

schlosna (approved):

carterkozak:
👍

PR #1393, opened 2022-03-18
Clarify total returns null when either heap or non-heap unavailable

schlosna:
## After this PR

==COMMIT_MSG==
Clarify total returns null when either heap or non-heap unavailable
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
@@ -38,8 +38,9 @@ import java.util.EnumMap;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiFunction;
-import java.util.function.LongSupplier;
+import java.util.function.Function;
 import java.util.function.Supplier;
+import javax.annotation.Nullable;
 
 /** {@link JvmMetrics} provides a standard set of metrics for debugging java services. */
 public final class JvmMetrics {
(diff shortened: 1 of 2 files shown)


tritium-metrics-jvm/src/test/java/com/palantir/tritium/metrics/jvm/JvmMetricsTest.java
```diff
@@ -181,6 +181,7 @@ void testUptimeHasExtraTags() {
     }
 
     @Test
+    @SuppressWarnings({"unchecked", "rawtypes"})
```
carterkozak: Fixed by https://github.com/palantir/gradle-baseline/pull/2119

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -190,6 +183,16 @@ protected Ratio getRatio() {
...
 
+    private static long totalHeapPlusNonHeap(MemoryMXBean memoryBean, ToLongFunction<MemoryUsage> longFunction) {
+        MemoryUsage heap = memoryBean.getHeapMemoryUsage();
+        MemoryUsage nonHeap = memoryBean.getNonHeapMemoryUsage();
+        if (heap == null && nonHeap == null) {
+            return -1;
+        }
+        return (heap == null ? 0 : longFunction.applyAsLong(heap))
+                + (nonHeap == null ? 0 : longFunction.applyAsLong(nonHeap));
+    }
```
carterkozak, line 190: I updated the target branch earlier this morning, no strong preference on which approach we merge though.
new commits:
https://github.com/palantir/tritium/pull/1392/commits/ad9ab1a684dd9d01e0535c4d0c9989a9738ae88d
https://github.com/palantir/tritium/pull/1392/commits/613e4e61a3aa888e2e94799b45ebbf8847e49dc9

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -190,6 +183,16 @@ protected Ratio getRatio() {
         });
     }
 
+    private static long totalHeapPlusNonHeap(MemoryMXBean memoryBean, ToLongFunction<MemoryUsage> longFunction) {
+        MemoryUsage heap = memoryBean.getHeapMemoryUsage();
+        MemoryUsage nonHeap = memoryBean.getNonHeapMemoryUsage();
```
carterkozak: I don't think these are allowed to be null by the api contract (though it's possible for an implementation to return null)

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -190,6 +183,16 @@ protected Ratio getRatio() {
...
+        if (heap == null && nonHeap == null) {
+            return -1;
+        }
+        return (heap == null ? 0 : longFunction.applyAsLong(heap))
+                + (nonHeap == null ? 0 : longFunction.applyAsLong(nonHeap));
```
carterkozak: In the case that heap or non-heap is unknown, I'm not sure that we can safely report a total value assuming heap/non-heap doesn't exist. I'd prefer not to report data rather than reporting misleading data.

carterkozak (approved):

PR #1522, opened 2022-07-11
Optimize Proxies.interfaces

schlosna:
## Before this PR
Proxies.interfaces might do extra work for common cases

## After this PR

==COMMIT_MSG==
Optimize Proxies.interfaces

Presize LinkedHashSet when computing unique set of interfaces.
==COMMIT_MSG==

### DIFF
--- a/tritium-proxy/src/main/java/com/palantir/tritium/proxy/Proxies.java
+++ b/tritium-proxy/src/main/java/com/palantir/tritium/proxy/Proxies.java
@@ -18,13 +18,14 @@ package com.palantir.tritium.proxy;
 
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Proxy;
-import java.util.Arrays;
 import java.util.LinkedHashSet;
 import java.util.Objects;
 import java.util.Set;
 
 @SuppressWarnings("PreferSafeLoggingPreconditions") // this module depends only on JDK
 public final class Proxies {
+
+    private static final Class<?>[] EMPTY_CLASS_ARRAY = new Class<?>[0];
 
     private Proxies() {}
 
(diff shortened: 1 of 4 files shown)


tritium-proxy/src/main/java/com/palantir/tritium/proxy/Proxies.java
```diff
@@ -59,27 +60,27 @@ static Class<?>[] interfaces(Class<?> iface, Class<?> delegateClass) {
...
+        for (Class<?> possibleInterface : delegateInterfaces) {
+            interfaces.add(checkIsInterface(possibleInterface));
+        }
+        return interfaces.toArray(EMPTY_CLASS_ARRAY);
     }
 
-    static void checkIsInterface(Class<?> iface) {
+    static Class<?> checkIsInterface(Class<?> iface) {
         if (!iface.isInterface()) {
             throw new IllegalArgumentException(iface + " is not an interface");
```
carterkozak, line 82: Could use safe-logging exceptions here, but not much difference either (risk exceeding inline threshold if we're worried about that)

tritium-proxy/src/main/java/com/palantir/tritium/proxy/Proxies.java
```diff
@@ -59,27 +60,27 @@ static Class<?>[] interfaces(Class<?> iface, Class<?> delegateClass) {
...
         }
 
-        checkAreAllInterfaces(interfaces);
-        return interfaces.toArray(new Class<?>[0]);
+        Class<?>[] delegateInterfaces = delegateClass.getInterfaces();
+        int expectedSize = delegateInterfaces.length + 1;
+        Set<Class<?>> interfaces = new LinkedHashSet<>((int) ((float) expectedSize / 0.75F + 1.0F));
+        interfaces.add(iface);
+        for (Class<?> possibleInterface : delegateInterfaces) {
+            interfaces.add(checkIsInterface(possibleInterface));
```
carterkozak, line 75: `possibleInterface` may be `iface` (or a super/sub-type of an interface we've already added). Allowed by the contract, depending what we're optimizing for it may or may not be worth changing.

carterkozak (approved):

tritium-proxy/src/main/java/com/palantir/tritium/proxy/Proxies.java
schlosna: Are you suggesting we should coalesce to the minimum set
Of interfaces? Since this is a set, we won't get duplicates, but we effectively union all of the interface types & super types.

tritium-proxy/src/main/java/com/palantir/tritium/proxy/Proxies.java
schlosna: I think I may convert these to logsafe in another PR to remove the suppressions and migrate over.

tritium-proxy/src/main/java/com/palantir/tritium/proxy/Proxies.java
carterkozak: Just a stream of consciousness — in the bytebuddy implementation I think we try to find the minimal set (after the primary interface, anyhow), but that’s probably not necessary here

PR #1525, opened 2022-07-17
Enable NullAway Optional emptiness checks

schlosna:
## Before this PR
NullAway does not [check Optional emptiness by default](https://github.com/uber/NullAway/wiki/Configuration#optional-emptiness-check)

## After this PR
==COMMIT_MSG==
Enable NullAway Optional emptiness checks
==COMMIT_MSG==


iamdanfox (approved):
neat!

carterkozak (approved):
slick!

PR #1537, opened 2022-08-01
MetricName TagMap supports replacement of existing entries

carterkozak:
## Why?
I've been thinking about adding additional data to our metrics describing the jvm runtime.
Some potential approaches would introduce risk that existing tags are repeated, although I suspect we'd use an approach including validation within metric-schema. In the event that there's some risk of key overriding, we'd want tritium to support it gracefully long before anything else relied on the functionality.

## Before this PR
```
SafeIllegalArgumentException: Base must not contain the extra key that is to be added
```

## After this PR

==COMMIT_MSG==
MetricName TagMap supports replacement of existing entries
==COMMIT_MSG==

## Possible downsides?
Slightly more code in `withEntry` could be harder for the jvm to optimize

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
@@ -18,7 +18,6 @@ package com.palantir.tritium.metrics.registry;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Ordering;
-import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.lang.reflect.Array;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -122,7 +121,12 @@ final class TagMap implements SortedMap<String, String> {
             String current = local[newPosition];
             int comparisonResult = current.compareTo(key);
             if (comparisonResult == 0) {
-                throw new SafeIllegalArgumentException("Base must not contain the extra key that is to be added");
+                if (Objects.equals(local[newPosition + 1], value)) {
+                    return this;
+                }
+                String[] newArray = local.clone();
+                newArray[newPosition + 1] = value;
+                return new TagMap(newArray);
             }
             if (comparisonResult > 0) {
                 break;
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
@@ -113,7 +113,7 @@ final class TagMap implements SortedMap<String, String> {
         return values;
     }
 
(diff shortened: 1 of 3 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
```diff
@@ -114,15 +113,20 @@ private static String[] toArray(Map<String, String> data) {
         return values;
     }
 
-    /** Returns a new {@link TagMap} with on additional entry. */
+    /** Returns a new {@link TagMap} with on additional or updated entry. */
```
schlosna: nit:
```suggestion
    /** Returns a new {@link TagMap} with one additional or updated entry. */
```

tritium-registry/src/test/java/com/palantir/tritium/metrics/registry/TagMapTest.java
```diff
@@ -53,12 +53,22 @@ void testSingleton() {
...
     @Test
     void testExtraEntryKeyAlreadyExists() {
-        TagMap map = TagMap.of(Collections.singletonMap("foo", "bar"));
-        assertThatThrownBy(() -> map.withEntry("foo", "baz")).isInstanceOf(IllegalArgumentException.class);
+        TagMap original = TagMap.of(Collections.singletonMap("foo", "bar"));
+        TagMap updated = original.withEntry("foo", "baz");
+        assertThat(updated)
+                .isEqualTo(ImmutableMap.of("foo", "baz"))
+                .as("Original must not be mutated")
+                .isNotSameAs(original);
```
schlosna: Is it worth testing state of original is not mutated? Thinking something like the following:

```suggestion
        assertThat(updated)
                .hasSize(1)
                .containsEntry("foo", "baz")
                .isEqualTo(ImmutableMap.of("foo", "baz"))
                .as("Original must not be mutated")
                .isNotSameAs(original)
                .isNotEqualTo(original);
        assertThat(original)
                .hasSize(1)
                .as("Original must not be mutated")
                .containsEntry("foo", "bar")
                .isNotEqualTo(updated);
```

schlosna (approved):
Overall looks good. Have a couple doc & test nits

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
carterkozak: Thanks! I think I meant "an" here
```suggestion
    /** Returns a new {@link TagMap} with an additional or updated entry. */
```

tritium-registry/src/test/java/com/palantir/tritium/metrics/registry/TagMapTest.java
carterkozak: Good idea, thanks!

PR #1548, opened 2022-08-08
[non urgent] uptime metric tracks whether --enable-preview is set

iamdanfox:
## Before this PR

I've been slowly trying to enable folks to use pattern-matching switch expressions instead of visitors for their conjure unions (see `#hackweek-2022-sealed-unions`), and one of the requirements for this is to add the `--enable-preview` flag at runtime (https://github.com/palantir/sls-packaging/pull/1365).

It occurred to me that if folks ever actually started _using this_ in production, we'd probably want a convenient way to be able to eyeball precisely which installations are using it... especially if we found some undesirable behaviour that only manifested in the --enable-preview case and we wanted to find usages and warn/remediate them.

## After this PR
==COMMIT_MSG==
The `uptime` metric now has an additional tag `enablePreview` which signifies if the `--enable-preview` flag is passed to java at runtime or not.
==COMMIT_MSG==

## Possible downsides?
- I'm not using a typed accessor to detect if this is enabled, so if java maintainers changed their mind about how the flag could be enabled (e.g. using a config file) then my attribute could incorrectly report false.

### DIFF
--- a/changelog/@unreleased/pr-1548.v2.yml
+++ b/changelog/@unreleased/pr-1548.v2.yml
@@ -4,3 +4,4 @@ feature:
     signifies if the `--enable-preview` flag is passed to java at runtime or not.
   links:
   - https://github.com/palantir/tritium/pull/1548
+  - https://openjdk.org/jeps/12
--- a/changelog/@unreleased/pr-1548.v2.yml
+++ b/changelog/@unreleased/pr-1548.v2.yml
@@ -1,0 +1,6 @@
+type: feature
+feature:
+  description: The `uptime` metric now has an additional tag `enablePreview` which
+    signifies if the `--enable-preview` flag is passed to java at runtime or not.
+  links:
+  - https://github.com/palantir/tritium/pull/1548
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
@@ -26,6 +26,7 @@ import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
 import com.palantir.tritium.metrics.MetricRegistries;
+import com.palantir.tritium.metrics.jvm.InternalJvmMetrics.AttributeUptime_EnablePreview;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.lang.management.ClassLoadingMXBean;
 import java.lang.management.ManagementFactory;
@@ -80,6 +81,10 @@ public final class JvmMetrics {
                 .javaRuntimeVersion(System.getProperty("java.runtime.version", "unknown"))
                 .javaVendorVersion(System.getProperty("java.vendor.version", "unknown"))
                 .javaVmVendor(System.getProperty("java.vm.vendor", "unknown"))
+                .enablePreview(
+                        runtimeMxBean.getInputArguments().contains("--enable-preview")
+                                ? AttributeUptime_EnablePreview.TRUE
+                                : AttributeUptime_EnablePreview.FALSE)
                 .build(runtimeMxBean::getUptime);
     }
 
--- a/tritium-metrics-jvm/src/main/metrics/metrics.yml
+++ b/tritium-metrics-jvm/src/main/metrics/metrics.yml
@@ -46,7 +46,7 @@ namespaces:
         - name: javaVmVendor
           docs: "Example: Azul Systems, Inc."
         - name: enablePreview
-          docs: "Whether the JVM is running with --enable-preview (see JEP 12)"
+          docs: "Whether the JVM is running with --enable-preview (see https://openjdk.org/jeps/12 )"
           values: [true, false]
       classloader.loaded:
         type: gauge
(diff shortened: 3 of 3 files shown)


tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -80,6 +81,10 @@ private static void registerAttributes(InternalJvmMetrics metrics) {
                 .javaRuntimeVersion(System.getProperty("java.runtime.version", "unknown"))
                 .javaVendorVersion(System.getProperty("java.vendor.version", "unknown"))
                 .javaVmVendor(System.getProperty("java.vm.vendor", "unknown"))
+                .enablePreview(
+                        runtimeMxBean.getInputArguments().contains("--enable-preview")
```
iamdanfox, line 85: Seems like we don't have junit tests to prove this method gives us sensible values, but I did validate this:

```
$ jshell --enable-preview
|  Welcome to JShell -- Version 15.0.7
|  For an introduction type: /help intro

jshell> java.lang.management.ManagementFactory.getRuntimeMXBean().getInputArguments()
$1 ==> [-agentlib:jdwp=transport=dt_socket,address=localhost:61023, --enable-preview]
```

changelog/@unreleased/pr-1548.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: feature
+feature:
+  description: The `uptime` metric now has an additional tag `enablePreview` which
+    signifies if the `--enable-preview` flag is passed to java at runtime or not.
+  links:
+  - https://github.com/palantir/tritium/pull/1548
```
schlosna, line 6:
```suggestion
  - https://github.com/palantir/tritium/pull/1548
  - https://openjdk.org/jeps/12
```

tritium-metrics-jvm/src/main/metrics/metrics.yml
```diff
@@ -31,16 +31,23 @@ namespaces:
...
+          docs: "Example: 2019-04-16"
+        - name: javaRuntimeVersion
+          docs: "Example: 11.0.3+7-LTS"
+        - name: javaVendorVersion
+          docs: "Example: Zulu11.31+11-CA"
+        - name: javaVmVendor
+          docs: "Example: Azul Systems, Inc."
+        - name: enablePreview
+          docs: "Whether the JVM is running with --enable-preview (see JEP 12)"
+          values: [true, false]
```
schlosna: https://github.com/palantir/metric-schema/pull/832 recently added the runtime JDK version info as tags to all metrics. Including  `enablePreview` tag on every metric produced is likely extraneous noise, so putting this on uptime seems reasonable and shouldn't change metric cardinality.

tritium-metrics-jvm/src/main/metrics/metrics.yml
```diff
@@ -31,16 +31,23 @@ namespaces:
...
+        - name: javaVersionDate
+          docs: "Example: 2019-04-16"
+        - name: javaRuntimeVersion
+          docs: "Example: 11.0.3+7-LTS"
+        - name: javaVendorVersion
+          docs: "Example: Zulu11.31+11-CA"
+        - name: javaVmVendor
+          docs: "Example: Azul Systems, Inc."
+        - name: enablePreview
+          docs: "Whether the JVM is running with --enable-preview (see JEP 12)"
```
schlosna:
```suggestion
          docs: "Whether the JVM is running with --enable-preview (see https://openjdk.org/jeps/12 )"
```

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
schlosna: If we wanted to unit test this, I think we'd need to bump source and library target to at least 12 and enable preview at compilation, effectively forcing consumers to JDK 15. I don't think we are quite ready  to force that JDK dependency bump for tritium quite yet (though possibly soon).

schlosna (commented):
Overall seems reasonable to track.

tritium-metrics-jvm/src/main/metrics/metrics.yml
iamdanfox: :+1: Interesting that we've added it in metric-schema - I guess that means it's slightly easier to split up a graph in datadog by javaversion _only_!

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
iamdanfox: I don't think we _ever_ want to --enable-preview for the compilation of published jars, because they can _only_ be run by the same java version that was used to compile them (i.e. if we did this with java15 for `:tritium-metrics-jvm` then anyone running java17 in production would fail to start)... I think this would quickly lead to unsolveable dependency graphs and much angst in #dev-backend-infra.

We'd need to either invent a gradle API for setting --enable-preview on the test sourceset only (I didn't add support for this initially https://github.com/palantir/gradle-baseline/pull/2322), or make up an entire separate gradle project that we would enable preview on.

iamdanfox:
:+1: to approve dschlosnagle's suggested commits - ~now just down to your +1~

EDIT nvm I think my +1 is meaningless as I'm no longer an admin on this repo sorry

image

schlosna (approved):

PR #1549, opened 2022-08-10
[draft] Add TaggedMetricRegistry#addMetrics(Map<String, String>, TaggedMetricSet)

schlosna:
## Before this PR
`TaggedMetricRegistry` only allowed adding `TaggedMetricSet` for single tag/value entry

## After this PR

==COMMIT_MSG==
Add TaggedMetricRegistry#addMetrics(Map<String, String>, TaggedMetricSet)
==COMMIT_MSG==

### DIFF
--- a/.palantir/revapi.yml
+++ b/.palantir/revapi.yml
@@ -5,3 +5,9 @@ acceptedBreaks:
       old: "class com.palantir.tritium.metrics.registry.MetricName.Builder"
       new: "class com.palantir.tritium.metrics.registry.MetricName.Builder"
       justification: "MetricName.Builder is final"
+  "0.51.0":
+    com.palantir.tritium:tritium-registry:
+    - code: "java.method.addedToInterface"
+      new: "method void com.palantir.tritium.metrics.registry.TaggedMetricRegistry::addMetrics(java.util.Map<java.lang.String,\
+        \ java.lang.String>, com.palantir.tritium.metrics.registry.TaggedMetricSet)"
+      justification: "Provide default implementation via AbstractTaggedMetricRegistry"
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/RealMetricName.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/RealMetricName.java
@@ -90,6 +90,10 @@ final class RealMetricName implements MetricName {
         return new RealMetricName(other.safeName(), withEntry(other.safeTags(), extraTagName, extraTagValue));
     }
 
+    public static RealMetricName create(MetricName metricName, TagMap tag) {
+        return new RealMetricName(metricName.safeName(), tag.withEntries(metricName.safeTags()));
+    }
+
     private static TagMap withEntry(SortedMap<String, String> tags, String extraTagName, String extraTagValue) {
         if (tags instanceof TagMap) {
             return ((TagMap) tags).withEntry(extraTagName, extraTagValue);
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
@@ -18,6 +18,8 @@ package com.palantir.tritium.metrics.registry;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Ordering;
+import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.Safe;
 import java.lang.reflect.Array;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -39,12 +41,12 @@ import javax.annotation.Nullable;
 /**
  * {@link TagMap} is a {@link SortedMap} implementation optimized for creation performance and memory overhead.
  * Primarily optimized to retain as little memory as possible, and create small short-lived intermediate objects.
- *
+ * <p>
  * Note that we expect fairly small tag maps which are iterated over, not used for lookups. Most {@link Map}
  * methods are implemented, but use a naive linear search rather than a binary search.
  */
 @SuppressWarnings("JdkObsolete")
-final class TagMap implements SortedMap<String, String> {
+final class TagMap implements SortedMap<@Safe String, @Safe String> {
 
     static final TagMap EMPTY = new TagMap(new String[0]);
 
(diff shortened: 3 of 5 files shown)


.palantir/revapi.yml
```diff
@@ -5,3 +5,9 @@ acceptedBreaks:
       old: "class com.palantir.tritium.metrics.registry.MetricName.Builder"
       new: "class com.palantir.tritium.metrics.registry.MetricName.Builder"
       justification: "MetricName.Builder is final"
+  "0.51.0":
+    com.palantir.tritium:tritium-registry:
+    - code: "java.method.addedToInterface"
+      new: "method void com.palantir.tritium.metrics.registry.TaggedMetricRegistry::addMetrics(java.util.Map<java.lang.String,\
+        \ java.lang.String>, com.palantir.tritium.metrics.registry.TaggedMetricSet)"
+      justification: "Provide default implementation via AbstractTaggedMetricRegistry"
```
schlosna, line 13: need to consider the implications of this break

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/RealMetricName.java
```diff
@@ -90,6 +90,10 @@ static MetricName create(MetricName other, String extraTagName, String extraTagV
         return new RealMetricName(other.safeName(), withEntry(other.safeTags(), extraTagName, extraTagValue));
     }
 
+    public static RealMetricName create(MetricName metricName, TagMap tag) {
+        return new RealMetricName(metricName.safeName(), tag.withEntries(metricName.safeTags()));
```
schlosna, line 94: need to think through how we want to handle overlap for tags as this may be the inverse of what we'd want

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
```diff
@@ -141,6 +148,14 @@ TagMap withEntry(String key, String value) {
         return new TagMap(newArray);
     }
 
+    TagMap withEntries(SortedMap<@Safe String, @Safe String> stringStringSortedMap) {
+        TagMap tagMap = this;
+        for (Map.Entry<String, String> entry : stringStringSortedMap.entrySet()) {
+            tagMap = tagMap.withEntry(entry.getKey(), entry.getValue());
```
schlosna, line 154: this is not optimized and will likely do `O(n)` expansions based on size of entries, so we'd want to optimize this

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TaggedMetricRegistry.java
```diff
@@ -144,6 +145,8 @@ default void registerWithReplacement(MetricName metricName, Gauge<?> gauge) {
      */
     void addMetrics(@Safe String safeTagName, @Safe String safeTagValue, TaggedMetricSet metrics);
 
+    void addMetrics(Map<@Safe String, @Safe String> tags, TaggedMetricSet metrics);
```
jkozlowski, line 148: Would this mean I'd be able to specify 0 tags? Cause that combined with metric-schema being able to group metrics by tags, would be my preferred solution. And actually something like:

```
interface TaggedMetricSetContext extends AutoCloseable {
  TaggedMetricRegistry registry();
}

TaggedMetricSetContext withScopedMetrics(Map<String, String> tags);
```

then you could write code like (using namespace tags feature in https://github.com/palantir/metric-schema/pull/840):

```
try (TaggedMetricSetContext ctx = taggedMetricRegistry.withScopedMetrics(Map.of())) {
    MyMetrics myMetrics = MyMetrics.builder()
        .registry(ctx.registry())
        .someTagValue("value")
   runJob(myMetrics);
}
```

is probably my favourite of all. Haven't thought through implications yet, but I've certainly written a bunch of tests in the past to make sure I pass the absolutely the same tags to add and remove methods.

schlosna: Can you describe cases where you'd want an empty tag map? I don't think we've touched this API in forever, so need to look at how consumers are using it/want to use it a bit more.

I'm not sure I quite follow the `Closer` proposal. Can you describe the use case/more pseudocode example? Are you envisioning use in try-with-resources to perform some cleanup callback?

jkozlowski: Sorry, should've fleshed it out.

> Can you describe cases where you'd want an empty tag map?
See the internal PR that sparked this conversation. Effectively adding a scoped TaggedMetricRegistry, if metric-schema supports adding per-namespace tags, means that ALL tags that apply to a metric set ARE documented using standard tools (i.e. by adding them in the "tags" section of the yaml), as opposed to in the "documentation" field which is free-form. Right now, if you have a look at internal skiing database or the PR I mentioned, when using the child TaggedMetricRegistries, you end up documenting it in a free-form way.

> I'm not sure I quite follow the Closer proposal. Can you describe the use case/more pseudocode example? Are you envisioning use in try-with-resources to perform some cleanup callback?

Yes, exactly. Being able to use try-with-resources would be useful in the context of the PR which sparked this PR. But it effectively removes the foot-gun of accidentally calling add followed by remove with the tags mixed up somehow.

carterkozak: the closer is nice, however I think it's a bit of a footgun where anything that takes less time than our metric reporting interval is likely not to be reported. The last ~reporting-interval of the scoped component is likely not to be reported, which is the most likely piece to include important and actionable data.

jkozlowski: @carterkozak well that's another thing which I think we should ideally solve: the other perverse case is where we actually report things for a while, and then it gets removed and we don't report the final set of values, which is also not ideal.

In most cases this situation is not that bad, but in the internal PR it means we're loosing some terminal set of values which are only set at the end of a run. I feel like the infra should handle this, not the consumers.

jkozlowski: I updated the example code somewhat.

jkozlowski: I think I have another reason why we want the ability to not specify any tags on a child metric registry: our internal 'titan's child' product is not able to validate tags not defined in metric-schema. Meaning if anyone ever needs to write a check using the 'titan's child' product they will not be able to use the tags passed to #addMetrics method.

jkozlowski: Hmm maybe I'm wrong and there's a special feature for this...

jkozlowski: Anyway, @carterkozak suggested just pushing a useless tag that doesn't mean anything, I think that's good enough

jkozlowski: At this point the Closer idea is probably a nice to have.

schlosna:
closing out for now

PR #1562, opened 2022-09-06
Error-prone 2.15.0

schlosna:
## Before this PR
Outdated error-prone. Simpler alternative to https://github.com/palantir/tritium/pull/1543

## After this PR
==COMMIT_MSG==
Error-prone 2.15.0
==COMMIT_MSG==


build.gradle
```diff
@@ -5,9 +5,9 @@ buildscript {
     }
 
     dependencies {
+        classpath 'com.google.errorprone:error_prone_core:2.15.0'
```
carterkozak, line 8: Thoughts on removing this in order to pick up error_prone 2.15.0 from gradle-baseline?

https://github.com/palantir/gradle-baseline/blob/4.161.0/versions.lock#L17-L23

build.gradle
```diff
@@ -84,6 +84,7 @@ allprojects {
                         ).collect { it.canonicalName() } as String[])
 
                 disable 'AndroidJdkLibsChecker', // ignore Android
+                        'CanIgnoreReturnValueSuggester', // high noise on builders
```
carterkozak, line 87: Latest baseline should take care of this out of the box: https://github.com/palantir/gradle-baseline/blob/4d9c322805ac8fbc1967baa529d5d1a609513fbe/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java#L173

schlosna: https://github.com/palantir/tritium/pull/1567 to switch over to gradle-baseline provided error-prone

PR #1570, opened 2022-09-15
Enforce NullAway for Guava packages

schlosna:
## Before this PR
`NullAway` was not checking Guava `com.google.common` annotated packages. When enabling as a test for https://github.com/palantir/gradle-baseline/pull/2382 it flagged one likely false positive that we can check.

```
> Task :tritium-processor:compileJava
/home/circleci/project/tritium-processor/src/main/java/com/palantir/tritium/processor/Methods.java:78: error: [NullAway] unboxing of a @Nullable value
                methodFieldName = originalMethodFieldName + i;
                                  ^
    (see http://t.uber.com/nullaway )
1 error
```

## After this PR
==COMMIT_MSG==
Enforce NullAway for Guava packages
==COMMIT_MSG==

### DIFF
--- a/tritium-processor/src/main/java/com/palantir/tritium/processor/Methods.java
+++ b/tritium-processor/src/main/java/com/palantir/tritium/processor/Methods.java
@@ -17,7 +17,7 @@
 package com.palantir.tritium.processor;
 
 import com.google.common.base.CaseFormat;
-import com.google.common.base.Strings;
+import com.palantir.logsafe.Preconditions;
 import com.squareup.javapoet.CodeBlock;
 import com.squareup.javapoet.FieldSpec;
 import com.squareup.javapoet.TypeName;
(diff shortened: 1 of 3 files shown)


tritium-processor/src/main/java/com/palantir/tritium/processor/Methods.java
```diff
@@ -66,9 +67,9 @@ static IdentityHashMap<MethodElements, String> methodStaticFieldName(
         IdentityHashMap<MethodElements, String> result = new IdentityHashMap<>();
         for (MethodElements method : instrumentedMethods) {
             String methodName = method.element().getSimpleName().toString();
-            String originalMethodFieldName = CaseFormat.LOWER_CAMEL
+            String originalMethodFieldName = Strings.nullToEmpty(CaseFormat.LOWER_CAMEL
                     .converterTo(CaseFormat.UPPER_UNDERSCORE)
-                    .convert(methodName);
+                    .convert(methodName));
```
schlosna: `methodName` shouldn't ever be null, so I'm not sure exactly why this flags as possibly null. Going to switch this to a `checkNotNull` to enforce that

carterkozak: I believe the intent is that convert input is nullable, and the outputs nullability matches the input, but the annotations don't have a clear way to express that

carterkozak (approved):

tritium-processor/src/main/java/com/palantir/tritium/processor/Methods.java
schlosna: Yeah, I was just a little surprised compiler couldn't verify this was always non-null as it flowed through. There's some interesting bits in [`Converter`](ParametricNullness) and [`com.google.common.base.NullnessCasts#uncheckedCastNullableTToT`](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/NullnessCasts.java#L55)

PR #1576, opened 2022-09-20
Use baseline-null-away

schlosna:
## Before this PR
Using hand rolled NullAway gradle config

## After this PR
==COMMIT_MSG==
Use baseline-null-away
==COMMIT_MSG==


build.gradle
```diff
@@ -53,10 +54,6 @@ allprojects {
     }
 
     plugins.withId('com.palantir.baseline-error-prone', {
-        dependencies {
-            errorprone 'com.uber.nullaway:nullaway'
-        }
-
```
schlosna, line 59: I am going to leave the `NullAway:AnnotatedPackages` & `NullAway:CheckOptionalEmptiness` overrides in place below for now until we can re-enable those in baseline

carterkozak: 👍

carterkozak (approved):

PR #1577, opened 2022-09-20
Normalize versions.props locators

schlosna:
## Before this PR
Several versions.props patterns were not broad enough to ensure we capture the necessary dependencies.

## After this PR
==COMMIT_MSG==
Normalize versions.props locators
==COMMIT_MSG==


versions.props
```diff
@@ -1,24 +1,24 @@
...
 com.palantir.goethe:* = 0.8.0
 com.palantir.jvm.diagnostics:* = 0.2.0
 com.palantir.safe-logging:* = 3.0.0
 com.palantir.tracing:* = 6.15.0
-com.squareup:javapoet = 1.13.0
-com.squareup.okhttp3:okhttp = 4.0.0
-com.uber.nullaway:nullaway = 0.10.1
-io.undertow:undertow-core = 2.2.19.Final
-net.bytebuddy:byte-buddy = 1.12.16
+com.squareup:* = 1.13.0
```
carterkozak: This may be too broad, no?

schlosna: Yeah, this could pick up/conflict with a few other modules if we added dependencies on the them. Right now javapoet is the only `com.squareup` runtime dep, `okhttp3` and `okio` are test only deps. Reverted this line

```
# runtime
com.squareup:javapoet:1.13.0 (2 constraints: 2b113eee)

[Test dependencies]
com.squareup.okhttp3:okhttp:4.0.0 (1 constraints: 0605ff35)
com.squareup.okio:okio:2.2.2 (1 constraints: 500c29fd)
```

carterkozak (approved):
thanks!

PR #1589, opened 2022-09-29
Tritium registry 'getMetrics()' doesn't throw on duplicate names

carterkozak:
It's better to provide what information we can rather than failing when this method is called, especially as it's not used in most production code in favor of the simpler `forEachMetric` method.

## After this PR

==COMMIT_MSG==
Tritium registry 'getMetrics()' doesn't throw on duplicate names
==COMMIT_MSG==

I haven't run into this recently, just something I thought to do when I used `buildKeepingLast` elsewhere.

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AbstractTaggedMetricRegistry.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AbstractTaggedMetricRegistry.java
@@ -182,7 +182,7 @@ public abstract class AbstractTaggedMetricRegistry implements TaggedMetricRegist
                 .forEach((metricName, metric) ->
                         result.put(RealMetricName.create(metricName, tag.getKey(), tag.getValue()), metric)));
 
-        return result.build();
+        return result.buildKeepingLast();
     }
 
     @Override
--- a/changelog/@unreleased/pr-1589.v2.yml
+++ b/changelog/@unreleased/pr-1589.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Tritium registry 'getMetrics()' doesn't throw on duplicate names
+  links:
+  - https://github.com/palantir/tritium/pull/1589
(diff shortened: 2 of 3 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AbstractTaggedMetricRegistry.java
```diff
@@ -182,7 +182,7 @@ public final Map<MetricName, Metric> getMetrics() {
                 .forEach((metricName, metric) ->
                         result.put(RealMetricName.create(metricName, tag.getKey(), tag.getValue()), metric)));
 
-        return result.build();
+        return result.buildKeepingLast();
```
schlosna, line 185: @carterkozak any concerns with this method only existing in latest [Guava 31.1+ from February 28, 2022](https://github.com/google/guava/releases/tag/v31.1)? I think most of the places using tritium will already have picked up a compatible Guava, just wondering about places like Gradle or other `provided` type class paths that might have older Guava and lead to `NoSuchMethodError`.

https://github.com/google/guava/blob/e4522099b392197a7d2870af634b74fc07d780e7/guava/src/com/google/common/collect/ImmutableMap.java#L619-L622

schlosna: Separately I'm wondering if an error-prone check might be useful in some places to have folks use explicit `buildOrThrow()` or `buildKeepingLast()` as I've seen a few places that made incorrect assumptions.

carterkozak: I don’t think we should bend over backwards for consumers that disregard the minimum requirements, seven months is long enough that even broken dependency issues should be improbable.

> Separately I'm wondering if an error-prone check might be useful

That’s a great idea, I’ll write it on my flight in a couple hours

schlosna (approved):
Sounds good, I'm happy holding Guava 31.1 as a minimum version for tritium

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AbstractTaggedMetricRegistry.java
carterkozak: PR is up :-) https://github.com/palantir/gradle-baseline/pull/2416

PR #1600, opened 2022-10-11
Add UtcTimestamp as more efficient way to get OffsetDateTime.now(UTC)

schlosna:
## Before this PR
While profiling some services, noticed some places in sls-logging that popped up around `OffsetDateTime.now(Clock)` where it is looking up the zone rules on each invocation. Since we typically log everything in UTC we can avoid this overhead and also provide some utilities as we often see uses of `OffsetDateTime.now(ZoneOffset.UTC)` and the like.

```
java.time.zone.ZoneRules.<init>(ZoneOffset)
at java.time.zone.ZoneRules.of(ZoneOffset)
at java.time.ZoneOffset.getRules()
at java.time.OffsetDateTime.ofInstant(Instant, ZoneId)
at java.time.OffsetDateTime.now(Clock)
at com.palantir.sls.logging.metric.MetricLogs.safeV1gauge(String, Gauge, Map)
```

## After this PR

==COMMIT_MSG==
Add UtcTimestamp as more efficient way to get OffsetDateTime.now(UTC)
==COMMIT_MSG==


schlosna:
Some of the JMH microbenchmarks for comparison on 16 core Intel(R) Xeon(R) Platinum 8175M CPU @ 2.50GHz. The speedup on JDK 11 is much better (~48%) vs. JDK 17 (~20%), but still worth it for places that are heavily generating timestamps.

```
# VM version: JDK 11.0.16.1, OpenJDK 64-Bit Server VM, 11.0.16.1+9-LTS
Benchmark                                  Mode  Cnt    Score    Error  Units
TimestampBenchmark.offsetDateTimeNowClock  avgt    3  112.421 ±  3.180  ns/op
TimestampBenchmark.offsetDateTimeNowZone   avgt    3  113.639 ± 11.465  ns/op
TimestampBenchmark.utcTimestampNow         avgt    3   68.684 ± 18.600  ns/op
```

```
# VM version: JDK 17.0.4.1, OpenJDK 64-Bit Server VM, 17.0.4.1+9-LTS
Benchmark                                  Mode  Cnt   Score    Error  Units
TimestampBenchmark.offsetDateTimeNowClock  avgt    3  88.114 ± 14.558  ns/op
TimestampBenchmark.offsetDateTimeNowZone   avgt    3  80.861 ±  6.930  ns/op
TimestampBenchmark.utcTimestampNow         avgt    3  64.392 ±  6.818  ns/op
```

tritium-time/src/test/java/com/palantir/tritium/time/UtcTimestampTest.java
```diff
@@ -0,0 +1,72 @@
...
+import java.time.Instant;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.time.temporal.ChronoUnit;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class UtcTimestampTest {
```
carterkozak, line 32: Great tests!

carterkozak (approved):
lgtm, thanks!

PR #1619, opened 2022-12-07
Add metrics for cpu shares and detected processor count

carterkozak:
## Before this PR
Limited insight into jdk changes impacting processor count and cpu shares in k8s (aka millicores in some environments)

## After this PR
==COMMIT_MSG==
Add metrics for cpu shares and detected processor count
==COMMIT_MSG==


tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -113,6 +119,16 @@ private static void registerThreads(InternalJvmMetrics metrics) {
...
     }
 
+    @VisibleForTesting
+    static void registerCpuShares(TaggedMetricRegistry registry, Optional<CpuSharesAccessor> maybeCpuSharesAccessor) {
+        maybeCpuSharesAccessor.ifPresentOrElse(
+                cpuSharesAccessor -> ContainerMetrics.of(registry).cpuShares((Gauge<Long>) () -> {
+                    OptionalLong cpuShares = cpuSharesAccessor.getCpuShares();
+                    return cpuShares.isEmpty() ? null : cpuShares.getAsLong();
+                }),
+                () -> log.info("CPU Shares information is not supported, metrics will not be reported"));
```
schlosna: Is this `null` going to cause problems (e.g. when reporting metrics) or does it get filtered? Should we report it as `-1`?

```suggestion
        maybeCpuSharesAccessor.ifPresentOrElse(
                cpuSharesAccessor -> ContainerMetrics.of(registry).cpuShares((Gauge<Long>)
                        () -> cpuSharesAccessor.getCpuShares().orElse(-1)),
                () -> log.info("CPU Shares information is not supported, metrics will not be reported"));
```

carterkozak: It's recorded as `null`, which isn't picked up by our pipeline. I think you're right that we're better off reporting `-1` as a sentinel that cpu shares aren't being used, but the metric is otherwise functional.

schlosna (approved):

PR #1679, opened 2023-03-17
@Instrument generates static trace method

schlosna:
## Before this PR
Noticed in some JFR that `@Instrument` generated classes were `grow()` the
`InstrumentationBuilder.handlers` list when there were only 1 or 2 handlers being used per instance. Many of these were only adding tracing instrumentation, but eating the overhead assumed by InstrumentationBuilder where multiple `InvocationEventHandler` would be registered. This becomes expensive when instrumenting high volumes of dynamic instances (e.g. per transaction).

## After this PR

==COMMIT_MSG==
@Instrument generates static trace method

Provide a `trace` method for generated Instrumented* classes to simplify and optimize the cases where one only wants tracing instrumentation on dynamic instances (e.g. when tracing individual transactions).

Remove a few excess allocations when using InstrumentationBuilder
==COMMIT_MSG==

### DIFF
--- a/tritium-processor/src/test/resources/com/palantir/tritium/examples/InstrumentedBindsParameter.java.generated
+++ b/tritium-processor/src/test/resources/com/palantir/tritium/examples/InstrumentedBindsParameter.java.generated
@@ -61,6 +61,10 @@ public final class InstrumentedBindsParameter implements BindsParameter {
                 BindsParameter.class, delegate, InstrumentedBindsParameter::new);
     }
 
+    public static BindsParameter trace(BindsParameter delegate) {
+        return InstrumentationBuilder.trace(BindsParameter.class, delegate, InstrumentedBindsParameter::new);
+    }
+
     public static BindsParameter instrument(BindsParameter delegate, TaggedMetricRegistry registry) {
         return builder(delegate).withTaggedMetrics(registry).withTracing().build();
     }
--- a/changelog/@unreleased/pr-1679.v2.yml
+++ b/changelog/@unreleased/pr-1679.v2.yml
@@ -1,0 +1,10 @@
+type: improvement
+improvement:
+  description: |-
+    @Instrument generates static trace method
+
+    Provide a `trace` method for generated Instrumented* classes to simplify and optimize the cases where one only wants tracing instrumentation on dynamic instances (e.g. when tracing individual transactions).
+
+    Remove a few excess allocations when using InstrumentationBuilder
+  links:
+  - https://github.com/palantir/tritium/pull/1679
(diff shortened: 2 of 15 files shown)


iamdanfox (approved):
Ah nice, thanks @schlosna! I've always felt that people might be using `Tritium.instrument` in scenarios when they don't actually care about the metrics (and so the consequent $ spend is unnecessary)... having this new `.trace` static method will mean I can get rid of a load of `.withTracing().build()` lines!

carterkozak:
> Noticed in some JFR that @Instrument generated classes were grow() the
InstrumentationBuilder.handlers list when there were only 1 or 2 handlers being used per instance.

I wouldn't generally expect instrumentation wrappers to be created on a per-operation basis, usually they're created at application startup. The annotation processor certainly has much lower overhead than the reflective proxy or byte-buddy implementations, but I want to make sure we're not doing something unnecessary in product code.

tritium-processor/src/test/resources/com/palantir/tritium/examples/InstrumentedBindsParameter.java.generated
```diff
@@ -61,6 +61,10 @@ public final class InstrumentedBindsParameter implements BindsParameter {
                 BindsParameter.class, delegate, InstrumentedBindsParameter::new);
     }
 
+    public static BindsParameter trace(BindsParameter delegate) {
+        return InstrumentationBuilder.trace(BindsParameter.class, delegate, InstrumentedBindsParameter::new);
+    }
```
carterkozak, line 66: I'd gone back and forth on adding something like this initially -- it probably doesn't make sense to write a tracing-java-specific processor, however that would avoid the InvocationContext overhead incurred by tritium (that said, it would likely be best in the long term to evolve the tritium handler api to avoid that requirement, and centralize this sort of thing in one project)

carterkozak (approved):

tritium-processor/src/test/resources/com/palantir/tritium/examples/InstrumentedBindsParameter.java.generated
schlosna: > > Noticed in some JFR that @Instrument generated classes were grow() the
> > InstrumentationBuilder.handlers list when there were only 1 or 2 handlers being used per instance.
> 
> I wouldn't generally expect instrumentation wrappers to be created on a per-operation basis, usually they're created at application startup. The annotation processor certainly has much lower overhead than the reflective proxy or byte-buddy implementations, but I want to make sure we're not doing something unnecessary in product code.

The use case here is folks instrumenting per transaction stores where we don't need metrics at the per transaction level (cardinality too high and we have higher level metrics for that), but wanting tracing for analyzing at the request level

schlosna: > I'd gone back and forth on adding something like this initially -- it probably doesn't make sense to write a tracing-java-specific processor, however that would avoid the InvocationContext overhead incurred by tritium (that said, it would likely be best in the long term to evolve the tritium handler api to avoid that requirement, and centralize this sort of thing in one project)

Yeah, I debated adding something in tracing-java, but this felt lightweight enough to support `tracing` in addition to `instrument` (that adds metrics + tracing).

PR #1686, opened 2023-03-30
Reduce executor allocation overhead from Timer.Context

carterkozak:
Prompted by https://github.com/palantir/atlasdb/pull/6490 (regardless of whether or not we keep the atlas metrics, we might as well optimize the instrumentation)

==COMMIT_MSG==
Reduce executor allocation overhead from Timer.Context
==COMMIT_MSG==


schlosna (approved):
Thanks @carterkozak !

PR #1696, opened 2023-04-07
Gradle spring cleaning

schlosna:
## Before this PR
Some gradle config was duplicative of what is provided by gradle-baseline or can be applied to all projects.

Dependency sorting was inconsistent, applied `sort -f -u` to all `build.gradle`

## After this PR

==COMMIT_MSG==
Gradle spring cleaning
==COMMIT_MSG==

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,6 +44,7 @@ allprojects {
     apply plugin: 'com.palantir.baseline-null-away'
     apply plugin: 'com.palantir.jakarta-package-alignment'
     apply plugin: 'com.palantir.java-format'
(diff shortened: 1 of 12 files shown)


build.gradle
```diff
@@ -111,22 +112,10 @@ allprojects {
 
     tasks.withType(Javadoc).configureEach {
         // suppress Javadoc doclint warnings in Java 8+
-        if (!System.getProperty("java.version").startsWith("1.7")) {
```
schlosna, line 114: everything is JDK 11+ these days 😄

carterkozak (approved):

PR #1700, opened 2023-04-14
Explicitly pin guava versions to avoid slow dependency resolution

schlosna:
## Before this PR
Slow dependency resolution due to version mismatches for guava peer modules

## After this PR

==COMMIT_MSG==
Explicitly pin guava versions to avoid slow dependency resolution
==COMMIT_MSG==


versions.props
```diff
@@ -4,7 +4,9 @@ com.github.ben-manes.caffeine:* = 3.1.6
 com.google.auto:* = 1.2.1
 com.google.auto.service:* = 1.0.1
 com.google.code.findbugs:* = 3.0.2
-com.google.guava:* = 31.1-jre
+com.google.guava:failureaccess = 1.0.1
+com.google.guava:guava = 31.1-jre
+com.google.guava:listenablefuture = 9999.0-empty-to-avoid-conflict-with-guava
```
pkoenig10: We don't need these other pins - we don't depend on them explicitly anywhere
```suggestion
com.google.guava:guava = 31.1-jre
```

schlosna: Good call

pkoenig10 (approved):

PR #1701, opened 2023-04-17
Bump dropwizard metrics to 4.2.18 in sync with Spark

schlosna:
## Before this PR
Pinned minimum release of `io.dropwizard.metrics:* = 4.1.1` due to older Spark 3.1 minimum dependency. 

## After this PR

==COMMIT_MSG==
Bump dropwizard metrics to 4.2.18 in sync with Spark 3.2.x+
==COMMIT_MSG==


versions.props
```diff
@@ -12,6 +12,8 @@ com.palantir.safe-logging:* = 3.2.0
 com.palantir.tracing:* = 6.17.0
 com.squareup:javapoet = 1.13.0
 com.squareup.okhttp3:* = 4.0.0
+# Keep major.minor metrics version in sync with Apache Spark
+io.dropwizard.metrics:* = 4.2.18
```
carterkozak, line 16: Can we use the upstreamed lock-free expo decay reservoir?

carterkozak (approved):

versions.props
schlosna: Yeah, I want to let this version propagate and we should be able to switch to the upstream version of https://github.com/dropwizard/metrics/blob/release/4.2.x/metrics-core/src/main/java/com/codahale/metrics/LockFreeExponentiallyDecayingReservoir.java with your changes. I need to do a quick diff when I'm on a desktop to make sure they're in sync.

schlosna: Filed https://github.com/palantir/tritium/issues/1704 to FLUP tomorrow

PR #1746, opened 2023-07-02
Simplify TritiumAnnotationProcessor

pkoenig10:
The indirection of `minimalInterfaces` appears to be unnecessary. There is always exactly one element and it is the annotated interface.


pkoenig10:
Closing in favor of https://github.com/palantir/tritium/pull/1747

PR #1747, opened 2023-07-02
Use MoreElements to collect methods

pkoenig10:
We use this utility in most of our other annotation processors that generate delegating implementations, including the conjure undertow annotation processor and a number of our internal annotation processors.

This simplifies the code and will make it easier to support selective per-method instrumentation in a future PR.

The one caveat is that this annotation processor no longer handles that case when methods with the same signature are inherited from difference ancestors. This seems acceptable since 1) this case is likely rare and 2) this isn't supported by any of our other annotation processors that use `MoreElements`. If this becomes an issue we add some logic to handle this case.

```java
interface A {
    void foo();
}

interface B {
    void foo();
}

@Instrument
interface MyInterface extends A, B {}
```

### DIFF
--- a/tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessor.java
+++ b/tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessor.java
@@ -16,6 +16,7 @@
 
 package com.palantir.tritium.processor;
 
+import com.google.auto.common.MoreElements;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
@@ -39,15 +40,10 @@ import com.squareup.javapoet.ParameterizedTypeName;
 import com.squareup.javapoet.TypeName;
 import com.squareup.javapoet.TypeSpec;
 import com.squareup.javapoet.TypeVariableName;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
-import javax.annotation.Nullable;
 import javax.annotation.processing.AbstractProcessor;
 import javax.annotation.processing.Filer;
 import javax.annotation.processing.Generated;
@@ -59,15 +55,12 @@ import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
-import javax.lang.model.element.Name;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.ExecutableType;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
-import javax.lang.model.type.TypeVariable;
 import javax.lang.model.util.Elements;
-import javax.lang.model.util.SimpleElementVisitor8;
 import javax.lang.model.util.Types;
 import javax.tools.Diagnostic.Kind;
 
@@ -78,7 +71,6 @@ public final class TritiumAnnotationProcessor extends AbstractProcessor {
     private static final String HANDLER_NAME = "handler";
     private static final ImmutableSet<String> ANNOTATIONS = ImmutableSet.of(Instrument.class.getName());
 
-    private final Set<Name> invalidElements = new HashSet<>();
     private Messager messager;
     private Filer filer;
     private Elements elements;
@@ -106,12 +98,10 @@ public final class TritiumAnnotationProcessor extends AbstractProcessor {
     @Override
     public boolean process(Set<? extends TypeElement> _annotations, RoundEnvironment roundEnv) {
         if (roundEnv.processingOver()) {
-            if (!invalidElements.isEmpty()) {
-                messager.printMessage(Kind.ERROR, "Processing completed with unresolved elements: " + invalidElements);
-            }
             return false;
         }
-        for (Element element : getElementsToProcess(roundEnv)) {
+
+        for (Element element : roundEnv.getElementsAnnotatedWith(Instrument.class)) {
             if (element.getKind() != ElementKind.INTERFACE) {
                 messager.printMessage(
                         Kind.ERROR,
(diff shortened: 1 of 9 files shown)


tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessor.java
```diff
@@ -150,66 +128,17 @@ public boolean process(Set<? extends TypeElement> _annotations, RoundEnvironment
...
-                return true;
-            }
-        }
-        return false;
-    }
-
-    /**
-     * Returns true if the given {@link TypeMirror} has sufficient type information, otherwise it may need to be
-     * deferred for another annotation processor to complete.
-     */
```
pkoenig10, line 177: Is there a particular reason why we need to handle this case here but not in our other delegate-generating annotation processors that use `MoreElements`?

tritium-processor/src/test/java/com/palantir/tritium/processor/TritiumProcessorTest.java
```diff
@@ -113,8 +113,7 @@ public void testDeprecatedInterface() {
 
     @Test
     public void testEmptyInterface() {
-        Compilation compilation = compileTestClass(TEST_CLASSES_BASE_DIR, Empty.class);
-        assertThat(compilation).hadErrorContaining("The annotated interface has no methods");
+        assertTestFileCompileAndMatches(TEST_CLASSES_BASE_DIR, Empty.class);
```
pkoenig10, line 116: While unusual, there's no reason why we can't support empty interfaces. It was a trivial fix, so I've fixed it here.

schlosna (approved):
Seems reasonable.

What is the behavior now if method is inherited from two super types? Will we fail at compile or runtime?

pkoenig10:
It will fail at compile time, because we will generate the method multiple times.

```
Compilation produced the following diagnostics:
/SOURCE_OUTPUT/com/palantir/tritium/examples/InstrumentedMyInterface.java:57: error: method foo() is already defined in class com.palantir.tritium.examples.InstrumentedMyInterface
    public void foo() {
                ^
Generated Source Files
======================

/SOURCE_OUTPUT/com/palantir/tritium/examples/InstrumentedMyInterface.java:

import com.palantir.tritium.annotations.internal.InstrumentationBuilder;
import com.palantir.tritium.api.event.InstrumentationFilter;
import com.palantir.tritium.event.Handlers;
import com.palantir.tritium.event.InvocationContext;
import com.palantir.tritium.event.InvocationEventHandler;
import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
import java.lang.reflect.Method;
import javax.annotation.processing.Generated;

@Generated("com.palantir.tritium.processor.TritiumAnnotationProcessor")
public final class InstrumentedMyInterface implements MyInterface {
    private static final Method FOO;

    private static final Method FOO1;

    static {
        try {
            FOO = MyInterface.class.getMethod("foo");
            FOO1 = MyInterface.class.getMethod("foo");
        } catch (NoSuchMethodException e) {
            throw new RuntimeException(e);
        }
    }

    private final MyInterface delegate;

    private final InvocationEventHandler<InvocationContext> handler;

    private final InstrumentationFilter filter;

    private InstrumentedMyInterface(
            MyInterface delegate, InvocationEventHandler<InvocationContext> handler, InstrumentationFilter filter) {
        this.delegate = delegate;
        this.handler = handler;
        this.filter = filter;
    }

    @Override
    public void foo() {
        if (this.handler.isEnabled()) {
            InvocationContext invocationContext = Handlers.pre(this.handler, this.filter, this, FOO, new Object[] {});
            try {
                this.delegate.foo();
                Handlers.onSuccess(this.handler, invocationContext);
            } catch (Throwable throwable) {
                Handlers.onFailure(this.handler, invocationContext, throwable);
                throw throwable;
            }
        } else {
            this.delegate.foo();
        }
    }

    @Override
    public void foo() {
        if (this.handler.isEnabled()) {
            InvocationContext invocationContext = Handlers.pre(this.handler, this.filter, this, FOO1, new Object[] {});
            try {
                this.delegate.foo();
                Handlers.onSuccess(this.handler, invocationContext);
            } catch (Throwable throwable) {
                Handlers.onFailure(this.handler, invocationContext, throwable);
                throw throwable;
            }
        } else {
            this.delegate.foo();
        }
    }

    @Override
    public String toString() {
        return "InstrumentedMyInterface{" + delegate + '}';
    }

    public static InstrumentationBuilder<MyInterface, MyInterface> builder(MyInterface delegate) {
        return new InstrumentationBuilder<MyInterface, MyInterface>(
                MyInterface.class, delegate, InstrumentedMyInterface::new);
    }

    public static MyInterface trace(MyInterface delegate) {
        return InstrumentationBuilder.trace(MyInterface.class, delegate, InstrumentedMyInterface::new);
    }

    public static MyInterface instrument(MyInterface delegate, TaggedMetricRegistry registry) {
    }
}
```

schlosna:
> It will fail at compile time, because we will generate the method multiple times

If this fails fast at compile time, that seems fine to release and roll out. We'll have to keep an eye on any repos that fail to bump tritium or server framework (the latter being more problematic where we would need to add support for these or revert).

schlosna:
@pkoenig10 defer to you when you want to merge & release.

pkoenig10:
I'd like to get a +1 from @carterkozak before merging this, particularly with regards to https://github.com/palantir/tritium/pull/1747#discussion_r1249745539.

pkoenig10:
For more context, my plan for per-method instrumentation is to only instrument methods that are declared directly in the wrapped interface.

This nicely solves the problem of multiple inherited methods with the same signature. It also requires users to be explicit about which methods they want to instrument, which feels like a desirable property.

I'd probably go so far as to say that this should be the recommended way for users to instrument their interfaces and that we should soft-deprecate the class level `@Instrument` annotation.

```java
interface A {
    void foo();
}

interface B {
    void foo();
    void bar();
}

// Only foo will be instrumented
interface MyInterface extends A, B {
    @Instrument
    @Override
    void foo();
}
```

carterkozak:
@pkoenig10 and I chatted offline -- current plan is to try moving the tritium processor to our delegate-processors library. I'll open-source delegate-processors if that goes smoothly. This way we get all the delegate implementation factory pieces in one place, and don't have to worry about breaking edge cases where an implementation inherits from multiple interfaces.

pkoenig10:
Closing in favor of https://github.com/palantir/tritium/pull/1766

PR #1754, opened 2023-07-18
Remove log placeholders

schlosna:
## Before this PR
A few log messages still included `{}` placeholders

## After this PR

==COMMIT_MSG==
Remove log placeholders
==COMMIT_MSG==

### DIFF
--- a/tritium-core/src/main/java/com/palantir/tritium/event/AbstractInvocationEventHandler.java
+++ b/tritium-core/src/main/java/com/palantir/tritium/event/AbstractInvocationEventHandler.java
@@ -80,9 +80,9 @@ public abstract class AbstractInvocationEventHandler<C extends InvocationContext
      * @param context invocation context
      */
     protected final void debugIfNullContext(@Nullable InvocationContext context) {
-        if (context == null) {
+        if (context == null && log.isDebugEnabled()) {
             log.debug(
-                    "{} encountered null metric context, likely due to exception in preInvocation",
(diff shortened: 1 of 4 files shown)


tritium-core/src/main/java/com/palantir/tritium/event/AbstractInvocationEventHandler.java
```diff
@@ -80,9 +80,9 @@ public final boolean isEnabled() {
      * @param context invocation context
      */
     protected final void debugIfNullContext(@Nullable InvocationContext context) {
-        if (context == null) {
+        if (context == null && log.isDebugEnabled()) {
             log.debug(
-                    "{} encountered null metric context, likely due to exception in preInvocation",
+                    "Encountered null metric context, likely due to exception in preInvocation",
```
carterkozak, line 85: We have a few places in dialogue where we've intentionally left slf4j placeholders for compatibility with non-sls logging formats. I don't have a strong preference, happy to go with whichever style you prefer in Tritium, but I thought I'd share that context in case it's helpful.

schlosna: Everywhere else we're not using placeholders for args, so just wanted to clean these ones up for consistency.

carterkozak: 👍 sounds good to me

carterkozak:
👍

PR #1766, opened 2023-08-01
Use delegate-processors

pkoenig10:
Change the `TritiumAnnotationProcessor` implementation to use the `delegate-processors` library. This has a couple benefits:
- Simplifies the implementation

    `delegate-processors` takes care of all the method resolution logic.

- Adds support for per-method instrumentation

    The `@Instrument` annotation can now be applied to individual methods instead of an entire interface.


versions.props
```diff
@@ -6,6 +6,7 @@ com.google.auto.service:* = 1.1.1
 com.google.code.findbugs:* = 3.0.2
 com.google.guava:guava = 32.1.2-jre
 com.google.testing.compile:* = 0.19
+com.palantir.delegate.processors:* = 0.79.0
```
carterkozak: 1.1.0 is up on maven central

pkoenig10:
Still need to update the README

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessorStrategy.java
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+        List<? extends TypeMirror> interfaces = type.getInterfaces();
+        if (interfaces.size() != 1) {
+            return typeName;
+        }
+
+        if (type.getEnclosedElements().stream().anyMatch(element -> !isOverride(element))) {
+            return typeName;
+        }
```
pkoenig10, line 94: This logic exists to support cases where you want to partially instrument an external interface. See the added `DelegateToRunnableMethod` and `DelegateToCallableMethod` tests for an example. If this was just `!type.getEnclosedElements().isEmpty()` then we wouldn't use the correct delegate type.

This feels a little loose, but I think will be fine in practice. The Java compiler disallows adding the `@Override` annotation to a method that does not override a superclass method. Worst case, your code fails to compile and you file a bug against Tritium.

This will not work if your delegate interface contains additional elements (ex. static fields), but the current implementation doesn't support that either, which feels reasonable.

tritium-core/src/main/java/com/palantir/tritium/event/Handlers.java
```diff
@@ -29,6 +29,13 @@ public final class Handlers {
 
     private static final SafeLogger log = SafeLoggerFactory.get(Handlers.class);
 
+    /**
+     * Returns a disabled {@link InvocationContext}.
+     */
+    public static InvocationContext disabled() {
+        return DisabledHandlerSentinel.INSTANCE;
+    }
```
carterkozak, line 38: This is reasonable to expose, but we should document that it should only be used to pass directly to other methods in `Handlers`.

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessorStrategy.java
```diff
@@ -0,0 +1,282 @@
...
+        if (type.getKind() != ElementKind.INTERFACE) {
+            arguments
+                    .context()
+                    .messager()
+                    .printMessage(
+                            Kind.ERROR,
+                            "Only interfaces may be instrumented using @" + Instrument.class.getSimpleName(),
+                            arguments.type().type());
+            return typeName;
+        }
```
carterkozak, line 85: I suppose we could relax this constraint in the future, allowing `@Instrument` on concrete implementations to generate a wrapper based on its interfaces -- for a separate PR if you think it would be useful.

carterkozak (approved):
looks great, thank you!

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessorStrategy.java
pkoenig10: Yeah I played around with this a bit, but it ended up being tricky. Particularly because we need a return type for the static factory methods. Currently [we're using the annotated type](https://github.com/palantir/tritium/blob/c9d266d82b7cadf86e6520060dd9ff70afaead7a/tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessorStrategy.java#L251), but this is incorrect if the annotated type is a class - you'll get something like:

```java
public final class InstrumentedAnnotatedClass implements Supplier<String> {
    ...

    public static AnnotatedClass instrument(AnnotatedClass delegate, TaggedMetricRegistry registry) {
    }
}
```

You might be able to make this work when there is a single interface, but it's much more complicated if the class implements multiple interfaces. And the returned value will be pretty hard to use, given that it will be some type parameter. This is partly why I'm not totally convinced that delegate processors should even support this case. This would also change the behavior when instrumenting an external interface - the `InstrumentedDelegateToCallable` factory methods would return `Callable` instead of `DelegateToCallable`.

I think there's value in limiting instrumentation to only interfaces, even if it's technically possible to support classes.

PR #1781, opened 2023-08-25
[test] Update versions.props

asvoboda:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


schlosna:
Looks like nit 🤖 is doing its job, we should be able to close this out, thanks!

PR #1794, opened 2023-09-17
Add ThreadFactory tags to differentiate platform and virtual threads

carterkozak:
Note that the total thread count metrics reported by ThreadMXBean only report on platform threads, so they needn't be updated.

==COMMIT_MSG==
Add ThreadFactory tags to differentiate platform and virtual threads
==COMMIT_MSG==


tritium-metrics/src/test/java/com/palantir/tritium/metrics/TaggedMetricsThreadFactoryTest.java
```diff
@@ -44,8 +47,55 @@ void testInstrumentation() {
...
+        Awaitility.waitAtMost(Duration.ofSeconds(3)).untilAsserted(() -> {
+            assertThat(created.getCount()).isOne();
+            assertThat(running.getCount()).isZero();
+        });
+        Awaitility.waitAtMost(Duration.ofSeconds(1))
+                .untilAsserted(() -> assertThat(thread.isAlive()).isFalse());
+    }
+
+    @Test
+    void testVirtualThreadInstrumentation() {
```
carterkozak, line 84: These tests share a fair bit of code. I bias away from sharing functionality between tests because it often results in refactors down the line that reduce the efficacy of some tests. In this case the tests are validating very similar things with different inputs, so there may be a stronger argument for sharing test code

tritium-metrics/src/test/java/com/palantir/tritium/metrics/TaggedMetricsThreadFactoryTest.java
```diff
@@ -44,8 +47,55 @@ void testInstrumentation() {
...
+        String name = "name";
+        TaggedMetricRegistry registry = new DefaultTaggedMetricRegistry();
+        ThreadFactory delegate = VirtualThreads.get()
+                .orElseThrow(() -> new AssertionError("Expected jdk21+"))
```
schlosna: Should we `assumeThat(VirtualThreads.get).isPresent()` or similar to allow skipping  test on JDK < 21?

schlosna (commented):
Overall looks good to me. One question on skipping virtual thread test on JDK < 21

schlosna (approved):

tritium-metrics/src/test/java/com/palantir/tritium/metrics/TaggedMetricsThreadFactoryTest.java
carterkozak: 👍

PR #1829, opened 2023-11-13
Tritium cpu shares metric reports -2 when unsupported

carterkozak:
This makes the data easier to search. The -2 result matches the result of `jdk.internal.platform.Metrics.getCpuShares` when shares are not supported.

==COMMIT_MSG==
Tritium cpu shares metric reports -2 when unsupported
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
@@ -120,10 +120,14 @@ public final class JvmMetrics {
 
     @VisibleForTesting
     static void registerCpuShares(TaggedMetricRegistry registry, Optional<CpuSharesAccessor> maybeCpuSharesAccessor) {
-        maybeCpuSharesAccessor.ifPresentOrElse(
-                cpuSharesAccessor -> ContainerMetrics.of(registry).cpuShares((Gauge<Long>)
-                        () -> cpuSharesAccessor.getCpuShares().orElse(-1L)),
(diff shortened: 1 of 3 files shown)


tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -120,10 +120,14 @@ private static void registerThreads(InternalJvmMetrics metrics) {
...
-                cpuSharesAccessor -> ContainerMetrics.of(registry).cpuShares((Gauge<Long>)
-                        () -> cpuSharesAccessor.getCpuShares().orElse(-1L)),
-                () -> log.info("CPU Shares information is not supported, cpu share metrics will not be reported"));
+        ContainerMetrics.of(registry)
+                .cpuShares(maybeCpuSharesAccessor
+                        .<Gauge<Long>>map(cpuSharesAccessor ->
+                                () -> cpuSharesAccessor.getCpuShares().orElse(-1L))
+                        .orElseGet(() -> {
+                            log.info("CPU Shares information is not supported, cpu share metrics will not be reported");
+                            return () -> -2L;
```
schlosna, line 129: do we want to document this somewhere?

schlosna (approved):

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
carterkozak: Sorry, I didn't see this before the change merged. I've created a follow-up with updated docs here: https://github.com/palantir/tritium/pull/1830

Issue #1895, opened 2024-03-06
README entry for how to instrument a cache 


ash211:
The README currently talks about how tritium is good for instrumenting a cache, but not how to do it.

I think the main points to hit are:
- Call `CaffeineCacheStats.registerCache()` [here](https://github.com/palantir/tritium/blob/d787bbbe676dcad82edeee1db59e66d868b2aca1/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java#L70-L80)
- if you have a common `TaggedMetricRegistry` like one from witchcraft, pass that as the parameter

schlosna:
Thanks @ash211 , I definitely agree the README could be spruced up with some solid examples. Would you be interested in submitting a PR 😄 ?

ash211:
Also I think you need to call `.recordStats()` on the CaffeineCache.Builder

And then I'm not sure the format that the metrics are emitted in either

ash211:
Do we have any recommendations for the metric name parameter to `registerCache()`?  I see a mix of `someString` and `MetricRegistry.getName(getClass(), "someString")`

ash211:
Also for the README, something like:
> For Palantirians, search for the internal "Caffeine Cache" dashboard to view the results.

schlosna:
Should be resolved by https://github.com/palantir/tritium/pull/1897

PR #1897, opened 2024-03-07
Initial Caffeine CacheStats recorder

schlosna:
## Before this PR

Begin to address https://github.com/palantir/tritium/issues/1895

Supersedes https://github.com/palantir/tritium/pull/1075

## After this PR

==COMMIT_MSG==
Initial Caffeine CacheStats recorder

Example usage:
```
TaggedMetricRegistry taggedMetricRegistry = ...
Cache<Integer, String> cache = Caffeine.newBuilder()
        .recordStats(CacheStats.of(taggedMetricRegistry, "unique-cache-name"))
        .build();

LoadingCache<String, Integer> loadingCache = Caffeine.newBuilder()
        .recordStats(CacheStats.of(taggedMetricRegistry, "unique-loading-cache-name"))
        .build(key::length);
```

==COMMIT_MSG==

### DIFF
--- a/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
+++ b/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
@@ -31,7 +31,7 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 import org.checkerframework.checker.index.qual.NonNegative;
 
-public final class CacheStats implements StatsCounter {
+public final class CacheStats implements StatsCounter, Supplier<StatsCounter> {
     private final CacheMetrics metrics;
     private final String name;
     private final Counter hitCounter;
@@ -112,8 +112,9 @@ public final class CacheStats implements StatsCounter {
         return (LoadingCache<K, V>) register((Cache<K, V>) cache);
     }
 
-    public Supplier<? extends StatsCounter> recorder() {
-        return () -> this;
+    @Override
+    public StatsCounter get() {
+        return this;
     }
 
     @Override
--- a/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
+++ b/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
@@ -42,12 +42,12 @@ public final class CacheStats implements StatsCounter, Supplier<StatsCounter> {
     /**
      * Creates a {@link CacheStats} instance that registers metrics for Caffeine cache statistics.
      * <p>
-     * Example method
+     * Example usage for a {@link com.github.benmanes.caffeine.cache.Cache} or
+     * {@link com.github.benmanes.caffeine.cache.LoadingCache}:
      * <pre>
-     *         CacheStats cacheStats = CacheStats.of(taggedMetricRegistry, "your-cache-name");
-     *         LoadingCache&lt;Integer, String> cache = cacheStats.register(Caffeine.newBuilder()
-     *                 .recordStats(cacheStats.record())
-     *                 .build(key -> computeSomethingExpensive(key));
+     *     LoadingCache&lt;Integer, String> cache = Caffeine.newBuilder()
+     *             .recordStats(CacheStats.of(taggedMetricRegistry, "your-cache-name"))
+     *             .build(key -> computeSomethingExpensive(key));
      * </pre>
      * @param taggedMetricRegistry tagged metric registry to add cache metrics
      * @param name cache name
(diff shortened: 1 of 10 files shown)


tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
```diff
@@ -0,0 +1,131 @@
...
+     * @return Caffeine stats instance to register via
+     * {@link com.github.benmanes.caffeine.cache.Caffeine#recordStats(Supplier)}.
+     */
+    public static CacheStats of(TaggedMetricRegistry taggedMetricRegistry, String name) {
+        return new CacheStats(CacheMetrics.of(taggedMetricRegistry), name);
+    }
+
+    public static CacheStats of(CacheMetrics metrics, String name) {
+        return new CacheStats(metrics, name);
+    }
```
carterkozak: I'd prefer if we made `CacheMetrics` private -- metric-schema doesn't make any guarantees around backward compatibility of generated code because it's expected to be consumed by the project which declares the metrics.

I suppose we may want to reuse the same metrics for guava and caffeine caches. If that's the case, I might recommend a new jar module with `internal` in the name, like `tritium-cache-internal`, where metrics may be defined and generated as public classes. The `tritium-cache-internal` module can be used as an `implementation` dep into the tritium modules which instrument specific cache implementations, so that the classes aren't directly visible to consumers.

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
```diff
@@ -0,0 +1,147 @@
...
+     *     LoadingCache&lt;Integer, String> cache = Caffeine.newBuilder()
+     *             .recordStats(CacheStats.of(taggedMetricRegistry, "your-cache-name"))
+     *             .build(key -> computeSomethingExpensive(key));
+     * </pre>
+     * @param taggedMetricRegistry tagged metric registry to add cache metrics
+     * @param name cache name
+     * @return Caffeine stats instance to register via
+     * {@link com.github.benmanes.caffeine.cache.Caffeine#recordStats(Supplier)}.
+     */
+    public static CacheStats of(TaggedMetricRegistry taggedMetricRegistry, String name) {
```
carterkozak: Might as well sprinkle safety annotations:
```suggestion
    public static CacheStats of(TaggedMetricRegistry taggedMetricRegistry, @Safe String name) {
```

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
schlosna: Makes sense, I can move the metric-schema definition and make package private. 

Generally I'd like to deprecate all uses of Guava caches and migrate any remaining to Caffeine, so not sure its worth adding similar for Guava caches. We already have [`BanGuavaCaches` in baseline](https://github.com/palantir/gradle-baseline/blob/a40f0782f0a1913748408f33eeceec0cdba99d92/.baseline/checkstyle/checkstyle.xml#L177-L181)

carterkozak: Perfect, thanks!

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
schlosna: updated

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
schlosna: updated

tritium-caffeine/src/main/metrics/cache-metrics.yml
```diff
@@ -0,0 +1,38 @@
...
+  javaPackage: com.palantir.tritium.metrics.caffeine
+  javaVisibility: packagePrivate
+namespaces:
+  cache:
+    docs: Cache statistic metrics
+    metrics:
+      hit.count:
+        type: counter
+        tags: [cache]
+        docs: Count of cache hits
```
carterkozak: Sorry I missed this in the previous round of review, I think all the counters should be updated to `type: meter`. Counter is interpreted as a gauge in most backends because it doesn't guarantee monotonic increasing values, where our Meter, confusingly, uses the [Prometheus Counter type](https://prometheus.io/docs/concepts/metric_types/#counter) for slightly easier indexing.

We may want to remove the `.count` name suffixes as well (otherwise we'll end up with queries for `cache_hit_count_count`), since those are described by both the `counter` and `meter` metric types.

tritium-caffeine/src/main/metrics/cache-metrics.yml
```diff
@@ -0,0 +1,38 @@
...
+        tags: [cache]
+        docs: Count of cache misses
+      load.success.count:
+        type: counter
+        tags: [cache]
+        docs: Count of successful cache loads
+      load.failure.count:
+        type: counter
+        tags: [cache]
+        docs: Count of failed cache loads
```
carterkozak: Thinking out loud here, entirely up to you either way:

We could declare a single metric along these lines, using a tag to distinguish success and failure:
```yaml
load:
  type: meter
  tags:
    - name: cache
    - name: result
       values: [success, failure]
```

Sometimes this makes it easier to track total interactions

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
```diff
@@ -0,0 +1,130 @@
...
+    @Override
+    public void recordLoadSuccess(@NonNegative long loadTime) {
+        loadSuccessMeter.mark();
+        totalLoadNanos.add(loadTime);
+    }
+
+    @Override
+    public void recordLoadFailure(@NonNegative long loadTime) {
+        loadFailureMeter.mark();
+        totalLoadNanos.add(loadTime);
```
carterkozak, line 104: I suppose we could make the meter into a timer, which also provides a “count” value in addition to percentiles, more timeseries though

tritium-caffeine/src/main/metrics/cache-metrics.yml
schlosna: yeah, I'm trying to keep metric name compatibility with the existing metrics produced by `com.palantir.tritium.metrics.caffeine.CaffeineCacheStats#registerCache(com.palantir.tritium.metrics.registry.TaggedMetricRegistry, com.github.benmanes.caffeine.cache.Cache<?,?>, java.lang.String)` for now as I'm not sure how complex the migration will be across the fleet.

carterkozak (dismissed):

carterkozak (approved):

PR #1919, opened 2024-03-27
Add cache size metrics to CacheStats

pkoenig10:
Follow up from https://github.com/palantir/tritium/pull/1897.

This PR provides a mechanism to register metrics that record the following information:
- Estimated size
- Weighted size
- Maximum size

These measurements are equivalent to those recorded by `CaffeineCacheStats.registerCache`.

I intentionally did not add `request`, `hit.ratio`, and `miss.ratio` metrics. Those values can be derived from the existing meters. This is better than creating gauges because:
- It reduces the number of metrics we create
- It works when there are multiple caches with the same name

### DIFF
--- a/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
+++ b/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
@@ -19,6 +19,8 @@ package com.palantir.tritium.metrics.caffeine;
 import com.codahale.metrics.Counting;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.Timer;
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Policy;
 import com.github.benmanes.caffeine.cache.RemovalCause;
 import com.github.benmanes.caffeine.cache.stats.StatsCounter;
 import com.google.common.collect.ImmutableMap;
@@ -33,14 +35,16 @@ import java.util.function.Supplier;
 import org.checkerframework.checker.index.qual.NonNegative;
 
 public final class CacheStats implements StatsCounter, Supplier<StatsCounter> {
+
+    private final CacheMetrics metrics;
     private final String name;
     private final Meter hitMeter;
     private final Meter missMeter;
     private final Timer loadSuccessTimer;
     private final Timer loadFailureTimer;
-    private final Meter evictionsTotalMeter;
     private final ImmutableMap<RemovalCause, Meter> evictionMeters;
-    private final LongAdder totalLoadNanos = new LongAdder();
+    private final ImmutableMap<RemovalCause, Meter> evictionWeightMeters;
+    private final LongAdder totalLoadTime = new LongAdder();
 
     /**
      * Creates a {@link CacheStats} instance that registers metrics for Caffeine cache statistics.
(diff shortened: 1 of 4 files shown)


tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
```diff
@@ -95,22 +125,25 @@ public void recordMisses(@NonNegative int count) {
...
-            counter.mark(weight);
+        Meter evictionMeter = evictionMeters.get(cause);
+        if (evictionMeter != null) {
+            evictionMeter.mark();
+        }
+        Meter evictionWeightMeter = evictionWeightMeters.get(cause);
+        if (evictionWeightMeter != null) {
+            evictionWeightMeter.mark(weight);
         }
-        evictionsTotalMeter.mark(weight);
```
pkoenig10, line 113: This was incorrect. The `evictionCount` provided to `CacheStats.of` should be the total number of evictions, not the total eviction weight. See the default `StatsCounter` for an example:

https://github.com/ben-manes/caffeine/blob/5b8e837b33949d1527eb9f76653725e2a53454ac/caffeine/src/main/java/com/github/benmanes/caffeine/cache/stats/ConcurrentStatsCounter.java#L74-L79

I've update this metric to record both the eviction count and eviction weight per-cause.

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
```diff
@@ -33,24 +35,28 @@
...
     private final String name;
     private final Meter hitMeter;
     private final Meter missMeter;
     private final Timer loadSuccessTimer;
     private final Timer loadFailureTimer;
-    private final Meter evictionsTotalMeter;
     private final ImmutableMap<RemovalCause, Meter> evictionMeters;
-    private final LongAdder totalLoadNanos = new LongAdder();
+    private final ImmutableMap<RemovalCause, Meter> evictionWeightMeters;
+    private final LongAdder totalLoadTime = new LongAdder();
```
pkoenig10, line 47: nit: Renamed this to match the `CacheStats.of` parameter name in order to improve comprehension.

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
```diff
@@ -61,17 +67,41 @@ public static CacheStats of(TaggedMetricRegistry taggedMetricRegistry, @Safe Str
...
+    public <K, V, C extends Cache<K, V>> C register(C cache) {
+        metrics.estimatedSize().cache(name).build(cache::estimatedSize);
+        metrics.weightedSize().cache(name).build(() -> cache.policy()
+                .eviction()
+                .flatMap(e -> e.weightedSize().stream().boxed().findFirst())
+                .orElse(null));
+        metrics.maximumSize().cache(name).build(() -> cache.policy()
+                .eviction()
+                .map(Policy.Eviction::getMaximum)
+                .orElse(null));
```
pkoenig10, line 85: Unless the `CaffeineCacheStats` implementation which returns 0 and -1 respectively for these guages when they are not present, this implementation just returns `null` so the metric is omitted. This feels objectively more correct and avoids confusion when the weighted size appears to be 0.

schlosna: @carterkozak do you recall why we used -1? I think there was some reason I didn't return null previously (should have left a comment). I need to take a look at the metric logging output side, but I think null should be fine.

schlosna (commented):
Looks good, I just want to confirm the null return

tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
carterkozak: I think we might have some logging for non-numeric values, however if that flags null, we should update the logging to silently ignore nulls (as opposed to strings/booleans/etc)

schlosna: -1 was a hack, null is more correct

https://github.com/palantir/tritium/pull/268#pullrequestreview-239201095

schlosna (approved):

PR #1920, opened 2024-03-27
Change CacheState.register to a fluent API

pkoenig10:
Follow up from https://github.com/palantir/tritium/pull/1919

This is a slightly less obvious but more ergonomic API for consumers.

### DIFF
--- a/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
+++ b/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
@@ -20,6 +20,7 @@ import com.codahale.metrics.Counting;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.Timer;
 import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.Policy;
 import com.github.benmanes.caffeine.cache.RemovalCause;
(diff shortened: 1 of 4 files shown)


tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
```diff
@@ -68,12 +68,24 @@ public static CacheStats of(TaggedMetricRegistry taggedMetricRegistry, @Safe Str
...
+     *     Cache&lt;Integer, String&gt; cache = CacheStats.of(taggedMetricRegistry, "your-cache-name")
+     *             .register(stats -> Caffeine.newBuilder()
+     *                     .recordStats(stats)
+     *                     .build());
+     * </pre>
+     * @param cacheFactory method which will be invoked to construct the cache
+     * @return the constructed cache instance
      */
-    public <K, V, C extends Cache<K, V>> C register(C cache) {
+    public <K, V, C extends Cache<K, V>> C register(Function<CacheStats, C> cacheFactory) {
```
carterkozak, line 86: nice

carterkozak (approved):

pkoenig10:
One nice thing about this API is that it makes it difficult to forget to call `recordStats`, since the provided `CacheStats` variable will be unused otherwise.

carterkozak:
Before this rolls out broadly, could we add a check to throw if stats aren't being recorded (which implies the cacheStats value hasn't been provided to the builder)?

pkoenig10:
https://github.com/palantir/tritium/pull/1922

PR #1921, opened 2024-03-28
Combine hit and miss cache metrics into single tagged metric

pkoenig10:
We already broke the `cache.hit` and `cache.miss` metrics when we changed them to meters in https://github.com/palantir/tritium/pull/1897. We can take this opportunity to improve these metrics by combining them into a single `cache.request` metric tagged by `result`.

This makes it easier for consumers to make queries for things like the total number of requests or the hit rate, without having to know/assume that "requests = hits + misses".

### DIFF
--- a/changelog/@unreleased/pr-1921.v2.yml
+++ b/changelog/@unreleased/pr-1921.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: The `cache.hit` and `cache.miss` metrics have been combined into a
+    single `cache.request` metric tagged by `result`.
+  links:
+  - https://github.com/palantir/tritium/pull/1921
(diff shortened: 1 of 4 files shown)


tritium-caffeine/src/main/metrics/cache-metrics.yml
```diff
@@ -28,13 +27,6 @@ namespaces:
         type: meter
         tags: [cache, cause]
         docs: Count of evicted weights
-      stats.disabled:
-        type: meter
-        tags: [cache]
-        docs: |
-          Registered cache does not have stats recording enabled, stats will always be zero.
-          To enable cache metrics, stats recording must be enabled when constructing the cache:
-          Caffeine.newBuilder().recordStats()
```
pkoenig10, line 37: This is unused

carterkozak (dismissed):

schlosna:
Agree that hit/miss tag makes metric easier to consume, will just need to update some dashboards anyway

schlosna (dismissed):

pkoenig10:
@carterkozak @schlosna can I get another +1? Looks like the changelog commit invalidated the approval.

schlosna (approved):

schlosna:
> @carterkozak @schlosna can I get another +1? Looks like the changelog commit invalidated the approval.

👍 

We should figure out if policybot should be invalidating on changelog

pkoenig10:
I suspect we do this for our open source repos because commits may be authored by non-Palantir contributors.

PR #1922, opened 2024-03-28
Check that stats recording is enabled when registering a cache

pkoenig10:
See https://github.com/palantir/tritium/pull/1920#issuecomment-2025332845


carterkozak (approved):
thanks!

PR #1936, opened 2024-04-30
Implement metrics to report the JVM DNS cache TTL values

carterkozak:
## Before this PR
Not much insight into DNS cache ttl values

==COMMIT_MSG==
Implement metrics to report the JVM DNS cache TTL values
==COMMIT_MSG==


tritium-metrics-jvm/build.gradle
```diff
@@ -19,3 +19,8 @@ dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter'
     testImplementation 'org.junit.jupiter:junit-jupiter-api'
 }
+
+moduleJvmArgs {
+    // Required for access to 'sun.net.InetAddressCachePolicy' for DNS cache metrics
+    opens 'java.base/sun.net'
+}
```
carterkozak: I don't really like throwing these all over the place -- I considered adding accessors  for these parameters within https://github.com/palantir/jvm-diagnostics instead, but the existing accessors are primarily for dynamic values, where these should be static.
Do you have a preference?

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -152,6 +154,38 @@ private static Map<Thread.State, Integer> threadsByState(ThreadMXBean threads) {
...
+            int positiveCacheSeconds = sun.net.InetAddressCachePolicy.get();
+            int negativeCacheSeconds = sun.net.InetAddressCachePolicy.getNegative();
+            int staleCacheSeconds = getStaleDnsCacheTtlSeconds();
+            metrics.dnsCacheTtlSeconds()
+                    .cache(DnsCacheTtlSeconds_Cache.POSITIVE)
+                    .build(() -> positiveCacheSeconds);
+            metrics.dnsCacheTtlSeconds()
+                    .cache(DnsCacheTtlSeconds_Cache.NEGATIVE)
+                    .build(() -> negativeCacheSeconds);
+            metrics.dnsCacheTtlSeconds().cache(DnsCacheTtlSeconds_Cache.STALE).build(() -> staleCacheSeconds);
```
schlosna: while these are currently static values, I'd prefer to eat the overhead of volatile reads to ensure we get the current values.

```suggestion
            metrics.dnsCacheTtlSeconds()
                    .cache(DnsCacheTtlSeconds_Cache.POSITIVE)
                    .build(sun.net.InetAddressCachePolicy::get);
            metrics.dnsCacheTtlSeconds()
                    .cache(DnsCacheTtlSeconds_Cache.NEGATIVE)
                    .build(sun.net.InetAddressCachePolicy::getNegative);
            metrics.dnsCacheTtlSeconds()
                    .cache(DnsCacheTtlSeconds_Cache.STALE)
                    .build(JvmMetrics::getStaleDnsCacheTtlSeconds);
```

carterkozak: I've realized we do something similar reading cpu share values, which is implemented in jvm-diagnostics, so I'm going to put up a quick PR into that repo which we can leverage from tritium. I think that will separate concerns nicely

tritium-metrics-jvm/build.gradle
schlosna: I'm ok with including this here given there's no public API or JMX accessors

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
schlosna: slight alternative might be something like this to avoid reflective lookup each time:
```
    private static void registerDnsCacheMetrics(InternalJvmMetrics metrics) {
        try {
            metrics.dnsCacheTtlSeconds()
                    .cache(DnsCacheTtlSeconds_Cache.POSITIVE)
                    .build(sun.net.InetAddressCachePolicy::get);
            metrics.dnsCacheTtlSeconds()
                    .cache(DnsCacheTtlSeconds_Cache.NEGATIVE)
                    .build(sun.net.InetAddressCachePolicy::getNegative);
            IntSupplier staleDnsCacheTtlSeconds = getStaleDnsCacheTtlSeconds();
            metrics.dnsCacheTtlSeconds().cache(DnsCacheTtlSeconds_Cache.STALE).build(staleDnsCacheTtlSeconds::getAsInt);
        } catch (Throwable t) {
            log.error("Failed to register DNS cache ttl metrics", t);
        }
    }

    private static IntSupplier getStaleDnsCacheTtlSeconds() {
        if (Runtime.version().feature() >= 21) {
            // Introduced in Java 21 by https://bugs.openjdk.org/browse/JDK-8306653
            Method getStale;
            try {
                getStale = InetAddressCachePolicy.class.getMethod("getStale");
                return () -> {
                    try {
                        return (Integer) getStale.invoke(null);
                    } catch (ReflectiveOperationException roe) {
                        log.debug("Failed to load stale InetAddressCachePolicy", roe);
                        return 0;
                    }
                };
            } catch (ReflectiveOperationException roe) {
                log.debug("Failed to load stale InetAddressCachePolicy", roe);
            }
        }
        return () -> 0;
    }
```

carterkozak: Good call! I've applied this in https://github.com/palantir/jvm-diagnostics/pull/309

mpritham (approved):

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
schlosna: thanks!

PR #1962, opened 2024-06-13
Use metric-schema for instrumentation metrics

pkoenig10:
## Before this PR

The tagged metrics produced by instrumented classes are cumbersome to use because each instrumented class uses a unique metric name. These metrics names are the fully qualified class names, which can be quite long making them difficult to use in metric query UIs.

It is also particularly cumbersome given our internal metrics filtering, since metrics for newly instrumented classes won't be available until they are used in a dashboard.

## After this PR

Instrumentation metrics are defined in metric-schema. There is a single instrumentation metric and the interface name, method name, and result are tags.

## Possible downsides?

Dashboards will need to be updated. I don't think users tend to have these instrumentation metrics on dashboards - they are more commonly used for one-off investigations.


tritium-lib/src/main/java/com/palantir/tritium/proxy/Instrumentation.java
```diff
@@ -148,13 +148,13 @@ public Builder<T, U> withMetrics(MetricRegistry metricRegistry) {
...
+         * @param serviceName - Service name to be used
          * @return - InstrumentationBuilder
          */
-        public Builder<T, U> withTaggedMetrics(TaggedMetricRegistry metricRegistry, String prefix) {
+        public Builder<T, U> withTaggedMetrics(TaggedMetricRegistry metricRegistry, String serviceName) {
             checkNotNull(metricRegistry, "metricRegistry");
-            String serviceName = Strings.isNullOrEmpty(prefix) ? interfaceClass.getName() : prefix;
-            this.handlers.add(new TaggedMetricsServiceInvocationEventHandler(metricRegistry, serviceName));
+            this.handlers.add(new TaggedMetricsServiceInvocationEventHandler(
+                    metricRegistry, Strings.isNullOrEmpty(serviceName) ? interfaceClass.getSimpleName() : serviceName));
```
pkoenig10, line 157: I've intentionally changed the default service name from `interfaceClass.getName()` to `interfaceClass.getSimpleName()` to more closely match the behavior of similar instrumentation elsewhere, such as the Conjure endpoint instrumentation.

tritium-lib/src/test/java/com/palantir/tritium/TritiumTest.java
```diff
@@ -1,180 +0,0 @@
...
-import org.junit.jupiter.api.extension.ExtendWith;
-import org.slf4j.impl.SimpleLogger;
-import org.slf4j.impl.TestLogs;
-import uk.org.webcompere.systemstubs.jupiter.SystemStub;
-import uk.org.webcompere.systemstubs.jupiter.SystemStubsExtension;
-import uk.org.webcompere.systemstubs.properties.SystemProperties;
-
-@ExtendWith(SystemStubsExtension.class)
-@SuppressWarnings("NullAway")
-public class TritiumTest {
```
pkoenig10, line 49: These tests seem duplicative, so I've removed them.

schlosna:
Directionally I'm supportive of this. @pkoenig10 do you have a sense of the migration lift for shifting from existing to new metric names & tags for monitors & queries?

pkoenig10:
It will definitely be manual work for dashboards.

I'm fairly confident that no one is using these metrics in monitors. Doing so would be fairly brittle, as the metric names could change after a refactor.

Searching internally I don't see any monitor definitions that appear to be using one of these metrics. It's a bit difficult since there are a number of legacy metrics that use class names, but which aren't these instrumented metrics. I used the following Sourcegraph query (case-sensitive and regex):

```
metric.*com\.palantir\..*\.[A-Z].+\.\w.+ and endpoint and file:.*\.hype
```

schlosna (approved):

PR #1958, opened 2024-06-06
Target JDK 17 bytecode & APIs

schlosna:
## Before this PR

https://github.com/palantir/tritium/pull/1956 is failing due to `delegate-processors:1.2.0` bump targeting JDK 17 bytecode via https://github.com/palantir/delegate-processors/pull/105

## After this PR

==COMMIT_MSG==
Target JDK 17 bytecode & APIs
==COMMIT_MSG==

PR #1976, opened 2024-07-08
Propagate @SuppressWarnings on @Instrument types

schlosna:
## Before this PR

If the interface that one `@Instrument` references a deprecated type the generated code will not compile if deprecation warnings are enabled, even if one adds `@SuppressWarnings("deprecation")` to the `@Instrument`ed type:

```
@Instrument
@SuppressWarnings("deprecation")
public interface DeprecatedType {
    Foo foo();
    @Deprecated
    interface Foo {
        String bar();
    }
}
```

```
TritiumProcessorTest > testDeprecatedType() FAILED
    Compilation produced the following diagnostics:
    /SOURCE_OUTPUT/com/palantir/tritium/examples/InstrumentedDeprecatedType.java:39: warning: [deprecation] com.palantir.tritium.examples.DeprecatedType.Foo in com.palantir.tritium.examples.DeprecatedType has been deprecated
        public DeprecatedType.Foo foo() {
                             ^
    /SOURCE_OUTPUT/com/palantir/tritium/examples/InstrumentedDeprecatedType.java:39: warning: [deprecation] com.palantir.tritium.examples.DeprecatedType.Foo in com.palantir.tritium.examples.DeprecatedType has been deprecated
        public DeprecatedType.Foo foo() {
                             ^
    /SOURCE_OUTPUT/com/palantir/tritium/examples/InstrumentedDeprecatedType.java:39: warning: [deprecation] com.palantir.tritium.examples.DeprecatedType.Foo in com.palantir.tritium.examples.DeprecatedType has been deprecated
        public DeprecatedType.Foo foo() {
                             ^
    /SOURCE_OUTPUT/com/palantir/tritium/examples/InstrumentedDeprecatedType.java:44: warning: [deprecation] com.palantir.tritium.examples.DeprecatedType.Foo in com.palantir.tritium.examples.DeprecatedType has been deprecated
                DeprecatedType.Foo _result = this.delegate.foo();
                              ^
    error: warnings found and -Werror specified
    Generated Source Files
    ======================

    /SOURCE_OUTPUT/com/palantir/tritium/examples/InstrumentedDeprecatedType.java:

    import com.palantir.tritium.annotations.internal.InstrumentationBuilder;
    import com.palantir.tritium.api.event.InstrumentationFilter;
    import com.palantir.tritium.event.Handlers;
    import com.palantir.tritium.event.InvocationContext;
    import com.palantir.tritium.event.InvocationEventHandler;
    import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
    import java.lang.reflect.Method;
    import java.util.Objects;
    import javax.annotation.processing.Generated;

    @Generated("com.palantir.tritium.processor.TritiumAnnotationProcessor")
    public final class InstrumentedDeprecatedType implements DeprecatedType {
        private static final Method FOO_ed3d1cab;

        static {
            try {
                FOO_ed3d1cab = DeprecatedType.class.getMethod("foo");
            } catch (NoSuchMethodException e) {
                throw new RuntimeException(e);
            }
        }

        private final DeprecatedType delegate;

        private final InvocationEventHandler<InvocationContext> handler;

        private final InstrumentationFilter filter;

        private InstrumentedDeprecatedType(
                DeprecatedType delegate, InvocationEventHandler<InvocationContext> handler, InstrumentationFilter filter) {
            this.delegate = Objects.requireNonNull(delegate, "delegate");
            this.handler = Objects.requireNonNull(handler, "handler");
            this.filter = Objects.requireNonNull(filter, "filter");
        }

        @Override
        public DeprecatedType.Foo foo() {
            InvocationContext _invocationContext = this.handler.isEnabled()
                    ? Handlers.pre(this.handler, this.filter, this, FOO_ed3d1cab, new Object[] {})
                    : Handlers.disabled();
            try {
                DeprecatedType.Foo _result = this.delegate.foo();
                Handlers.onSuccess(this.handler, _invocationContext, _result);
                return _result;
            } catch (Throwable _throwable) {
                Handlers.onFailure(this.handler, _invocationContext, _throwable);
                throw _throwable;
            }
        }

        @Override
        public String toString() {
            return "InstrumentedDeprecatedType{" + this.delegate + "}";
        }

        public static InstrumentationBuilder<DeprecatedType, DeprecatedType> builder(DeprecatedType delegate) {
            return new InstrumentationBuilder<DeprecatedType, DeprecatedType>(
                    DeprecatedType.class, delegate, InstrumentedDeprecatedType::new);
        }

        public static DeprecatedType trace(DeprecatedType delegate) {
            return InstrumentationBuilder.trace(DeprecatedType.class, delegate, InstrumentedDeprecatedType::new);
        }

        public static DeprecatedType instrument(DeprecatedType delegate, TaggedMetricRegistry registry) {
        }
    }
        at app//com.palantir.tritium.processor.TritiumProcessorTest.assertTestFileCompileAndMatches(TritiumProcessorTest.java:157)
        at app//com.palantir.tritium.processor.TritiumProcessorTest.testDeprecatedType(TritiumProcessorTest.java:120)

20 tests completed, 1 failed
```

## After this PR

==COMMIT_MSG==
Propagate @SuppressWarnings on @Instrument types
==COMMIT_MSG==

### DIFF
--- a/tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessorStrategy.java
+++ b/tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessorStrategy.java
@@ -27,6 +27,7 @@ import com.palantir.tritium.event.Handlers;
 import com.palantir.tritium.event.InvocationContext;
 import com.palantir.tritium.event.InvocationEventHandler;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+import com.squareup.javapoet.AnnotationSpec;
 import com.squareup.javapoet.ClassName;
 import com.squareup.javapoet.CodeBlock;
 import com.squareup.javapoet.FieldSpec;
@@ -188,11 +189,17 @@ enum TritiumAnnotationProcessorStrategy implements DelegateProcessorStrategy {
 
     @Override
     public void customize(CustomizeArguments arguments, TypeSpec.Builder generatedType) {
-        if (isDeprecated(arguments.type().type())) {
+        TypeElement type = arguments.type().type();
+        if (isDeprecated(type)) {
             generatedType.addAnnotation(Deprecated.class);
         }
 
-        TypeName annotatedTypeName = TypeName.get(arguments.type().type().asType());
+        SuppressWarnings suppressWarnings = type.getAnnotation(SuppressWarnings.class);
+        if (suppressWarnings != null) {
+            generatedType.addAnnotation(AnnotationSpec.get(suppressWarnings));
+        }
+
+        TypeName annotatedTypeName = TypeName.get(type.asType());
         List<AnnotatedTypeMethod> instrumentedMethods = instrumentedMethods(arguments.type());
 
         if (!instrumentedMethods.isEmpty()) {
--- a/tritium-processor/src/test/java/com/palantir/tritium/examples/DeprecatedType.java
+++ b/tritium-processor/src/test/java/com/palantir/tritium/examples/DeprecatedType.java
@@ -1,0 +1,31 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.examples;
+
+import com.palantir.tritium.annotations.Instrument;
+
+@Instrument
+@SuppressWarnings("deprecation")
+public interface DeprecatedType {
+
+    Foo foo();
+
+    @Deprecated
+    interface Foo {
+        String bar();
+    }
+}
--- a/tritium-processor/src/test/resources/com/palantir/tritium/examples/InstrumentedDeprecatedType.java.generated
+++ b/tritium-processor/src/test/resources/com/palantir/tritium/examples/InstrumentedDeprecatedType.java.generated
@@ -1,0 +1,71 @@
+package com.palantir.tritium.examples;
+
+import com.palantir.tritium.annotations.internal.InstrumentationBuilder;
+import com.palantir.tritium.api.event.InstrumentationFilter;
+import com.palantir.tritium.event.Handlers;
+import com.palantir.tritium.event.InvocationContext;
+import com.palantir.tritium.event.InvocationEventHandler;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+import java.lang.reflect.Method;
+import java.util.Objects;
+import javax.annotation.processing.Generated;
+
+@Generated("com.palantir.tritium.processor.TritiumAnnotationProcessor")
+@SuppressWarnings("deprecation")
+public final class InstrumentedDeprecatedType implements DeprecatedType {
+    private static final Method FOO_ed3d1cab;
+
+    static {
+        try {
+            FOO_ed3d1cab = DeprecatedType.class.getMethod("foo");
+        } catch (NoSuchMethodException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private final DeprecatedType delegate;
+
+    private final InvocationEventHandler<InvocationContext> handler;
+
+    private final InstrumentationFilter filter;
+
+    private InstrumentedDeprecatedType(
+            DeprecatedType delegate, InvocationEventHandler<InvocationContext> handler, InstrumentationFilter filter) {
+        this.delegate = Objects.requireNonNull(delegate, "delegate");
+        this.handler = Objects.requireNonNull(handler, "handler");
+        this.filter = Objects.requireNonNull(filter, "filter");
+    }
+
+    @Override
+    public DeprecatedType.Foo foo() {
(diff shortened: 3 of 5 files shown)


tritium-processor/src/test/resources/com/palantir/tritium/examples/InstrumentedDeprecatedType.java.generated
```diff
@@ -0,0 +1,71 @@
...
+import com.palantir.tritium.event.Handlers;
+import com.palantir.tritium.event.InvocationContext;
+import com.palantir.tritium.event.InvocationEventHandler;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+import java.lang.reflect.Method;
+import java.util.Objects;
+import javax.annotation.processing.Generated;
+
+@Generated("com.palantir.tritium.processor.TritiumAnnotationProcessor")
+@SuppressWarnings("deprecation")
```
schlosna, line 14: this is now propagated from `DeprecatedType`

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessorStrategy.java
```diff
@@ -188,11 +189,17 @@ public List<FieldSpec> additionalFields(AdditionalFieldsArguments arguments) {
...
     @Override
     public void customize(CustomizeArguments arguments, TypeSpec.Builder generatedType) {
-        if (isDeprecated(arguments.type().type())) {
+        TypeElement type = arguments.type().type();
+        if (isDeprecated(type)) {
             generatedType.addAnnotation(Deprecated.class);
         }
 
-        TypeName annotatedTypeName = TypeName.get(arguments.type().type().asType());
+        SuppressWarnings suppressWarnings = type.getAnnotation(SuppressWarnings.class);
```
pkoenig10, line 197: This should probably be `isDeprecated`, right? Otherwise it forces users who want to instrument external interfaces with deprecated methods to define their own interface just so they can add `@SuppressWarnings`.

I think we'd also want to check for `@SuppressWarnings` on the method itself, possibly instead of the class level annotation.

tritium-processor/src/test/java/com/palantir/tritium/examples/DeprecatedType.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.examples;
+
+import com.palantir.tritium.annotations.Instrument;
+
+@Instrument
+@SuppressWarnings("deprecation")
+public interface DeprecatedType {
```
pkoenig10, line 23: Do we care about handling the case where the `@Deprecated` annotation is applied to the type?

pkoenig10 (commented):
Should this instead be part of https://github.com/palantir/delegate-processors? Or does that feel too opinionated? If the generated code fails to compile, it's hard for me to imagine a case where you wouldn't want these annotations on a delegate impl.

tritium-processor/src/test/java/com/palantir/tritium/examples/DeprecatedType.java
schlosna: The `@Deprecated` type case should already be handled & tested:


https://github.com/palantir/tritium/blob/e80ad5b98fbf82115b320c93d09e0f5a83edb1b1/tritium-processor/src/test/java/com/palantir/tritium/examples/DeprecatedInterface.java#L21-L26

https://github.com/palantir/tritium/blob/e80ad5b98fbf82115b320c93d09e0f5a83edb1b1/tritium-processor/src/test/resources/com/palantir/tritium/examples/InstrumentedDeprecatedInterface.java.generated#L13-L15

And the deprecated method case:

https://github.com/palantir/tritium/blob/e80ad5b98fbf82115b320c93d09e0f5a83edb1b1/tritium-processor/src/test/java/com/palantir/tritium/examples/DeprecatedMethod.java#L21-L25

https://github.com/palantir/tritium/blob/e80ad5b98fbf82115b320c93d09e0f5a83edb1b1/tritium-processor/src/test/resources/com/palantir/tritium/examples/InstrumentedDeprecatedMethod.java.generated#L40-L43

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessorStrategy.java
schlosna: The case I'm hitting is an external interface that itself is not deprecated, but some methods return types that are deprecated. If you're instrumenting an external type, you already have to declare your own interface, so I'm somewhat in favor of explicitly needing to also `@SuppressWarnings` so folks acknowledge the underlying type has deprecated types (method arguments or return types).

The code above on line 192-194 already propagates `@Deprecated` on an `@Instrument`ed type (see comment below with linked tests & output).

schlosna (commented):
> Should this instead be part of https://github.com/palantir/delegate-processors? Or does that feel too opinionated? If the generated code fails to compile, it's hard for me to imagine a case where you wouldn't want these annotations on a delegate impl.

I'm not opposed to pushing this into delegate-processors itself, though I'd still want to land this in tritium unless we revert delegate-processors to target JDK 11 per https://github.com/palantir/tritium/pull/1958 as there are still some straggler projects that need to complete JDK 17 upgrade.

tritium-processor/src/main/java/com/palantir/tritium/processor/TritiumAnnotationProcessorStrategy.java
pkoenig10: > If you're instrumenting an external type, you already have to declare your own interface

Oh duh, I was being silly.

> so I'm somewhat in favor of explicitly needing to also @SuppressWarnings so folks acknowledge the underlying type has deprecated types

👍

pkoenig10:
Sounds good, I think it probably makes more sense here than in delegate-processors anyways. Can always move the logic over if we end up needing this in more places ala rule of three.

pkoenig10 (approved):

PR #1989, opened 2024-08-12
AbstractTaggedRegistry creates HistogramWithSamples

marcioapaiva:
Implements HistogramWithSamples, a subclass of Histogram that keeps track of samples and their respective traces, provided via a user-given supplier.
Additionally modifies the AbstractTaggedRegistry's histogram supplier to transparently return an instance of HistogramWithSamples. This will allow loggers to persist samples related to the collected histogram metrics.

==COMMIT_MSG==
AbstractTaggedRegistry creates HistogramWithSamples
==COMMIT_MSG==


tritium-samples-tracing/src/main/java/com/palantir/tritium/samples/tracing/SamplesTracer.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Optional;
+
+public class SamplesTracer {
+    public static Optional<String> getTraceIfObservable() {
+        if (!com.palantir.tracing.Tracer.isTraceObservable()) {
+            return Optional.empty();
+        }
+        return Optional.of(com.palantir.tracing.Tracer.getTraceId());
+    }
+}
```
schlosna, line 28: I would suggest avoiding the `Optional` allocation on hot paths here and just use empty string to indicate no trace ID and document the behavior.

```suggestion
import com.palantir.tracing.Tracer;

public final class SamplesTracer {
    /**
     * Returns the trace ID for this thread's trace if it is observable; otherwise a blank string.
     */
    public static String getTraceIfObservable() {
        if (Tracer.isTraceObservable()) {
            return Tracer.getTraceId();
        }
        return "";
    }

    private SamplesTracer() {}
}
```

tritium-samples/src/main/java/com/palantir/tritium/samples/LockFreeSampler.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private void doObserve(long value, long currentMillis) {
+        Sample existingSample = sampleRef.get();
+        if (existingSample == null
+                || currentMillis - existingSample.getTimestamp() >= maxRetentionPeriodMillis
+                || value > existingSample.getValue()) {
+            Optional<String> maybeNewTrace = traceSupplier.get();
+            if (maybeNewTrace.isPresent()) {
+                sampleRef.set(Sample.of(value, currentMillis, maybeNewTrace.get()));
+            }
```
schlosna, line 83: per above, I would avoid `Optional`:
 
```suggestion
            String maybeNewTrace = traceSupplier.get();
            if (!maybeNewTrace.isEmpty()) {
                sampleRef.set(Sample.of(value, currentMillis, maybeNewTrace));
            }
```

tritium-samples/src/main/java/com/palantir/tritium/samples/LockFreeSampler.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    @Override
+    public List<Sample> collect() {
+        Sample sample = sampleRef.get();
+        if (sample == null) {
+            return Collections.emptyList();
```
schlosna, line 91: nit: consistent immutable list implementations and avoid classloading `java.util.Collections`

```suggestion
            return List.of();
```

tritium-samples/src/main/java/com/palantir/tritium/samples/Sample.java
```diff
@@ -0,0 +1,69 @@
...
+            return false;
+        }
+        Sample sample = (Sample) obj;
+        return value == sample.value && timestamp == sample.timestamp && Objects.equals(traceId, sample.traceId);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(value, timestamp, traceId);
+    }
```
schlosna, line 63: it doesn't look like we're currently putting `Sample` in any hash based collections, but might be worth using allocation free `hashCode()` implementation to avoid issues down the road.

```suggestion
    @Override
    public int hashCode() {
        int result = Long.hashCode(getValue());
        result = 31 * result + Long.hashCode(getTimestamp());
        result = 31 * result + getTraceId().hashCode();
        return result;
    }
```

tritium-samples/src/main/java/com/palantir/tritium/samples/Sample.java
```diff
@@ -0,0 +1,69 @@
...
+
+public final class Sample {
+    private final long value;
+    private final long timestamp;
+    private final String traceId;
+
+    private Sample(long value, long timestamp, String traceId) {
+        this.value = value;
+        this.timestamp = timestamp;
+        this.traceId = traceId;
```
schlosna, line 29:
```suggestion
        this.traceId = Preconditions.checkNotNull(traceId, "traceId");
```

tritium-samples/src/main/java/com/palantir/tritium/samples/Sampler.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package com.palantir.tritium.samples;
+
+import java.util.List;
+
+interface Sampler {
+    void observe(long value);
+
+    List<Sample> collect();
```
schlosna, line 24: is the intent here to be able to use something like a ring buffer to keep last-N or specific outlier samples?

tritium-samples/src/main/java/com/palantir/tritium/samples/LockFreeSampler.java
```diff
@@ -0,0 +1,95 @@
...
+public final class LockFreeSampler implements Sampler {
+
+    private final AtomicReference<Sample> sampleRef = new AtomicReference<>(null);
+
+    private final AtomicLong lastSampleTickRef = new AtomicLong(0);
+
+    private final Clock clock;
+
+    private static final long DEFAULT_MAX_RETENTION_PERIOD_MILLIS = 1 * 1_000; // 30s
+    private static final long DEFAULT_SAMPLE_INTERVAL_NANOS = 70 * 1_000_000; // 70ms
```
schlosna, line 39: how did we arrive at these?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/AbstractTaggedMetricRegistry.java
```diff
@@ -74,7 +77,8 @@ protected Supplier<Counter> counterSupplier() {
     @Nonnull
     @SuppressWarnings("NoFunctionalReturnType") // metric factory
     protected Supplier<Histogram> histogramSupplier() {
-        return () -> new Histogram(createReservoir());
+        return () ->
+                new HistogramWithSamples(createReservoir(), new LockFreeSampler(SamplesTracer::getTraceIfObservable));
```
schlosna, line 81: Can you write up some benchmarks and napkin math on the impacts of capturing 1 (to N) samples per histogram? We often have *many* Histograms (including in timers), so want to understand memory footprint as well as concurrency

schlosna (commented):
I have some initial comments on this draft. I'd be interested in seeing some tests and examples how we intend to utilize this, and understand the overheads for latency, throughput, concurrency, and memory.

marcioapaiva:
@schlosna thanks for taking a look! This was just some initial ideation on the wiring - I'll close it now, but reading through your comments I think many of them are addressed by https://github.com/palantir/tritium/pull/1992 which we've been iterating on. If you can take a look at it, that'd be great - especially around memory footprint on https://github.com/palantir/tritium/pull/1992#discussion_r1733297592

PR #1991, opened 2024-08-15
[Draft] LockFreeExponentiallyDecayingReservoir keeps track of samples

marcioapaiva:
In this tentative approach to keeping tabs on samples, the LockFreeReservoir keeps track of a `ConcurrentSkipListMap<Double, WeightedSampleMetadata>` instead of a `ConcurrentSkipListMap<Double, WeightedSample>`.

**The returned Snapshot instance stays the same. The sample values are exposed through a new method on the reservoir.** When returning a regular snapshot, the WeightedSampleMetadatas get mapped to regular WeightedSamples.

Necessary changes:
- We would then need to have a new Histogram subclass that exposed getSampleMetadata() instead of just getSnapshot.
- sls-logging would then identify the subclass of Histogram (say HistogramWithSampleMetadata) and log it appropriately.

==COMMIT_MSG==
LockFreeExponentiallyDecayingReservoir keeps track of samples
==COMMIT_MSG==


marcioapaiva:
Closing in favor of https://github.com/palantir/tritium/pull/1992

PR #1992, opened 2024-08-16
LockFreeExponentiallyDecayingReservoirWithExemplars implementation

marcioapaiva:
Implements new Reservoir (LockFreeExponentiallyDecayingReservoirWithExemplars) that keeps track of exemplar data. This allows us to keep track of exemplar data for logging and grafana ingestion.

- LockFreeExponentiallyDecayingReservoirWithExemplars keeps track of a ConcurrentSkipListMap<Double, WeightedSampleMetadata> instead of a ConcurrentSkipListMap<Double, WeightedSample>.
- The Snapshot returned by the reservoir's getSnapshot() method is a subclass of WeightedSnapshot that also exposes the sample values through a new interface (ExemplarsCapture).

Necessary changes to other libraries to extract value from this feature:

- sls-logging would perform an instanceof check on the reservoir's returned snapshot and execute `getSamples` in case it's an instance of `ExemplarsCapture`.
- witchcraft will (in a feature-flagged manner) instantiate Histograms with this new Reservoir, passing it an ExemplarMetadataProvider that captures the thread's TraceId and current timestamp when invoked.

==COMMIT_MSG==
LockFreeExponentiallyDecayingReservoirWithExemplars implementation allows Exemplar tracking.
==COMMIT_MSG==

## Possible downsides?
See memory footprint discussion below.


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotMetadata.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+public final class WeightedSnapshotMetadata extends WeightedSnapshot {
+
+    /**
+     * A single sample item with value and its weights for {@link WeightedSnapshotMetadata}.
+     */
+    public static class WeightedSampleMetadata {
+        public final long value;
+        public final double weight;
+        public final SampleMetadata metadata;
```
marcioapaiva: The timestamp could probably go in the SampleMetadata itself, letting the SampleMetadataProvider implementation decide if that's actually useful information.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotMetadata.java
```diff
@@ -0,0 +1,59 @@
...
+import com.codahale.metrics.Snapshot;
+import com.codahale.metrics.WeightedSnapshot;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * A snapshot of a collection of {@link WeightedSampleMetadata}s.
+ */
+public final class WeightedSnapshotMetadata extends WeightedSnapshot {
```
carterkozak: I don't think this should be exposed as public api, instead we should have an interface which this implements which can be public API to expose captured data.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/SampleMetadataProvider.java
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.metrics.registry;
+
+import java.util.function.Supplier;
+
+public interface SampleMetadataProvider extends Supplier<SampleMetadata> {}
```
carterkozak: Should we name this something about exemplars given that's the intended use case?

API design nit: Instead of extending Supplier, I prefer to declare a method, in this case that might look something like:

```java
public interface SampleMetadataProvider {
    SampleMetadata collect();
}
```

In practice I think we may want this to return an Optional so we're not forced to return something if, for example, there's no traceId.

I think we'll also want to make the SampleMetadata type generic so that tritium isn't necessarily responsible for defining what's collected

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotMetadata.java
```diff
@@ -0,0 +1,59 @@
...
+    private final List<SampleMetadata> sampleMetadatas;
+
+    /**
+     * Create a new {@link Snapshot} with the given values.
+     *
+     * @param values an unordered set of values in the reservoir
+     */
+    public WeightedSnapshotMetadata(Collection<WeightedSampleMetadata> values) {
+        super(values.stream().map(v -> new WeightedSample(v.value, v.weight)).collect(Collectors.toList()));
+        sampleMetadatas = values.stream().map(v -> v.metadata).collect(Collectors.toList());
```
carterkozak: I don't recall if `values` is immutable at this point, we may want to iterate over it once and collect all samples+metadata in one iteration so that we know our data is consistent.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoirWithSamples.java
```diff
@@ -79,34 +77,43 @@ private static final class State {
         private final int size;
         private final long startTick;
         // Count is updated after samples are successfully added to the map.
-        private final ConcurrentSkipListMap<Double, WeightedSample> values;
+        private final ConcurrentSkipListMap<Double, WeightedSampleMetadata> values;
```
carterkozak: we should duplicate this class into a new one to handle exemplars, with a name that fits better. We can do that later on, the way this is structured makes the diff easier to follow for now :-)

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/SampleMetadataProvider.java
marcioapaiva: Yeah that's fair. I had been trying to avoid the term "Exemplars" as product-specific and instead just using Sample, but it seems like it's defined by OpenTelemetry too, which I hadn't seen (https://opentelemetry.io/docs/specs/otel/metrics/data-model/#exemplars) so having Exemplars in the name makes sense to me.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoirWithSamples.java
marcioapaiva: You mean the reservoir class or the WeightedSampleMetadata? I can def rename WeightedSampleMetadata to something like WeightedSampleWithExemplar.

carterkozak: The reservoir type, sorry I was ambiguous! For later though, since this makes it easier to review for the time being

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/SampleMetadataProvider.java
marcioapaiva: Something like https://github.com/palantir/tritium/pull/1992/commits/5cf5d7febeacc359ce10e0d72f3e9841993875f8 for the generic type?

carterkozak: I think we need to get a bit clever with how we handle the generic type. When we report metrics via `registry.forEachMetric(name, metric -> {})` we won't have the generic type info anymore.

I could see something like leveraging the `ExemplarMetadataProvider<T>` for extraction, that way we end up calling something along these lines:

```
public interface ExemplarsCapture {
    <T> List<T> getExemplars(ExemplarMetadataProvider<T> provider);
}
```

Internally we would be able to check whether or not the exemplar values are sourced from this provider otherwise return an empty list.
This way the reservoir itself needn't be generic, and if someone were to register a different histogram from another metadata provider, it would be ignored without causing issues in the reporting code.

marcioapaiva: But the consumer (eg sls-logging) won't have access to the provider to be able to pass it in. Do you mean to expose it and pass it in to check that the parameterized type (T) maps the exemplar collection we have internally? Or to have the WeightedSnapshot actually have some metadata on which provider was used?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotMetadata.java
marcioapaiva: good point, it doesn't seem to be have any immutability guarantees. Changed to do a single pass - had to switch to extending Snapshot and keeping a local instance of WeightedSnapshot (instead of just extending WeightedSnapshot), but that's probably better anyways despite being a little more verbose.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExemplarMetadata.java
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.metrics.registry;
+
+public interface ExemplarMetadata {}
```
carterkozak: I think this class can be deleted at this point

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExemplarMetadataProvider.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.metrics.registry;
+
+import java.util.Optional;
+
+public interface ExemplarMetadataProvider<T> {
+    Optional<T> collect();
```
carterkozak: We may want to use something like this instead, since it can potentially be a hot path. 

```
@Nullable
T collect();`
```

When the optional is empty, that's relatively inexpensive, but the additional object wrapper allocation could impact us if we're not careful.
We generally try to avoid using nulls, but we can make trade-offs in hot code-paths, and this isn't an API that users will be interacting with, so it may be safe.
I don't have a terribly strong opinion on this one, mostly thinking out loud.

carterkozak: The other piece we should think about is how we plan to expose the measured value. When a histogram is updated with value `123`, we don't currently expose a way for the exemplar to reference that value in any way.

It may be helpful to include that value in `collect(long value);`. With only that, the exemplar samples will need to duplicate the value already held in the reservoir, which might be reasonable, but we could instead consider exposing something different from `ExemplarsCapture` like 

```diff
- <T> List<T> getSamples(ExemplarMetadataProvider<T> provider);
+ <T> List<LongExemplar<T>> getSamples(ExemplarMetadataProvider<T> provider);
```
Where `LongExemplar<T>` is roughly
```java
interface LongExemplar<T> {
    T exemplar();
    long value();
}
```

marcioapaiva: Yeah good point! I'm definitely more comfortable not duplicating the samples at the reservoir level - so made the switch you proposed. I do have the values duplicated in the Snapshot for easier implementation but that should be way less of a concern - let me know if you disagree.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExemplarMetadataProvider.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.metrics.registry;
+
+import javax.annotation.Nullable;
+
+public interface ExemplarMetadataProvider<T> {
+    @Nullable
+    T collect(long value);
```
carterkozak: Since we've added the `LongExemplar<T>` type, I think we can remove the `long value` arg here -- the collected metadata and value will be combined later on (unless you have a case in mind where the value is necessary for metadata collection -- otherwise it becomes easy to accidentally hold the same value on heap in two places unnecessarily).

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LongExemplar.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.metrics.registry;
+
+public interface LongExemplar<T> {
+    T metadata();
```
carterkozak: Let's add log-safety annotations on the `T` generic type, requiring all exemplar data to be logsafe. It's much easier to start with that requirement and loosen it than it is to go in the other direction!

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/ExemplarMetadataProvider.java
marcioapaiva: You're right and I don't have a case in mind for using the value in metadata collection, just a mind blip.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -79,34 +78,44 @@ private static final class State {
...
-        private void addSample(double priority, long value, double itemWeight, boolean bypassIncrement) {
-            if (values.putIfAbsent(priority, new WeightedSample(value, itemWeight)) == null
+        private void addSample(
+                double priority,
+                long value,
+                double itemWeight,
+                boolean bypassIncrement,
+                @Nullable Object exemplarMetadata) {
+            if (values.putIfAbsent(priority, new WeightedSampleWithExemplar(value, itemWeight, exemplarMetadata))
+                            == null
```
carterkozak: I think we discussed memory utilization of holding samples at one point, thinking through this out loud here:

If we assume a no-op exemplar sampler (always returns null), we've replaced the old weighted sample with a new type that's nearly identical, but has an additional Object reference.

Assuming a standard configuration on a reasonably modern JVM (e.g. G1GC + heap smaller than 32g) we should go from

before (32 bytes):
```
com.codahale.metrics.WeightedSnapshot.WeightedSample object internals:
OFF  SZ     TYPE DESCRIPTION               VALUE
  0   8          (object header: mark)     N/A
  8   4          (object header: class)    N/A
 12   4          (alignment/padding gap)   
 16   8     long WeightedSample.value      N/A
 24   8   double WeightedSample.weight     N/A
Instance size: 32 bytes
Space losses: 4 bytes internal + 0 bytes external = 4 bytes total
```

<details>
  <summary>Without compressed oops (also 32 bytes)</summary>

```
com.codahale.metrics.WeightedSnapshot.WeightedSample object internals:
OFF  SZ     TYPE DESCRIPTION               VALUE
  0   8          (object header: mark)     N/A
  8   8          (object header: class)    N/A
 16   8     long WeightedSample.value      N/A
 24   8   double WeightedSample.weight     N/A
Instance size: 32 bytes
Space losses: 0 bytes internal + 0 bytes external = 0 bytes total
```

</details>

after (32 bytes):
```
com.palantir.tritium.metrics.registry.WeightedSnapshotWithExemplars.WeightedSampleWithExemplar object internals:
OFF  SZ     TYPE DESCRIPTION                           VALUE
  0   8          (object header: mark)                 N/A
  8   4          (object header: class)                N/A
 12   4   Object WeightedSampleWithExemplar.metadata   N/A
 16   8     long WeightedSampleWithExemplar.value      N/A
 24   8   double WeightedSampleWithExemplar.weight     N/A
Instance size: 32 bytes
Space losses: 0 bytes internal + 0 bytes external = 0 bytes total
```

However, if we don't have compressed oops (e.g. Z1GC), the size will be impacted (40 bytes):
```
com.palantir.tritium.metrics.registry.WeightedSnapshotWithExemplars.WeightedSampleWithExemplar object internals:
OFF  SZ     TYPE DESCRIPTION                           VALUE
  0   8          (object header: mark)                 N/A
  8   8          (object header: class)                N/A
 16   8     long WeightedSampleWithExemplar.value      N/A
 24   8   double WeightedSampleWithExemplar.weight     N/A
 32   8   Object WeightedSampleWithExemplar.metadata   N/A
Instance size: 40 bytes
Space losses: 0 bytes internal + 0 bytes external = 0 bytes total
```

32->40 bytes isn't terrible in the zgc case, especially considering the map key is a boxed double, which takes an additional ~24 bytes, plus the Map.Entry object, so the 8 byte change shouldn't be terribly dramatic.

We could extract an interface and provide two implementations, one which has the metadata slot, and another which does not, eliminating that memory change for zgc services -- unclear if that's worthwhile. What do you think? @schlosna may have opinions :-)

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotWithExemplars.java
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * A {@link WeightedSnapshot} with support for storing exemplar metadata for each sample.
+ */
+final class WeightedSnapshotWithExemplars extends Snapshot implements ExemplarsCapture {
+
+    /**
+     * A single sample item with value and its weights for {@link WeightedSnapshotWithExemplars}.
+     */
+    @SuppressWarnings("VisibilityModifier")
+    public static class WeightedSampleWithExemplar {
```
carterkozak: package private
```suggestion
   static class WeightedSampleWithExemplar {
```

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotWithExemplars.java
```diff
@@ -0,0 +1,129 @@
...
+
+        public final long value;
+        public final double weight;
+        public final @Nullable Object metadata;
```
carterkozak: let's make these private and use accessor methods

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
marcioapaiva: Thanks for the detailed analysis! My gut reaction here is that when we wire this up to witchcraft, we can just use the upstream Reservoir if we're feeding it a a No-op ExemplarMetadataProvider - so for an actual no-op case we'd have no differences in memory footprint.

**For the cases with an actual provider**
You guys probably know better than me how prevalent the older JVMs are, but I'm tempted to ignore the `32->40` bytes per-sample increase on the WeightedSampleWithExemplar class and accept the alignment "luck" of not having any memory footprint difference with compressed oops.

What I am more concerned about for that situation is the actual footprint of the generic type object itself (the `T exemplarMetadata`) and the timestamp, which (assuming we can store a TraceId with 64 bits and avoid a string) would be 8 bytes for the traceId + 8 bytes for the timestamp + object headers (8+4?), so potentially closer to 32 bytes with alignment. But I think we can discuss that when we have the object definition in the PR.




So for now as far as the no-ops go (for folks not using Exemplars at all) I think using the upstream Reservoir makes more sense and then there's no issue merging this, but curious for @schlosna's opinion!

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -244,9 +260,19 @@ public static final class Builder {
         private double alpha = DEFAULT_ALPHA;
         private Duration rescaleThreshold = DEFAULT_RESCALE_THRESHOLD;
         private Clock clock = Clock.defaultClock();
+        private ExemplarMetadataProvider<?> exemplarMetadataProvider = () -> null;
 
         private Builder() {}
 
+        /**
+         * Maximum number of samples to keep in the reservoir. Once this number is reached older samples are
+         * replaced (based on weight, with some amount of random jitter).
```
schlosna: fix javadoc copy/paste from `Builder#size`

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -51,7 +51,7 @@
  *     rescaling is substantially less common than updating at peak load. Even so, when size is reasonably small
  *     it can be more efficient to rescale than to park and context switch.
  * </ul>
- *
+ * <p>
  * See {@link com.codahale.metrics.ExponentiallyDecayingReservoir} Copyright 2010-2012 Coda Hale and Yammer, Inc.
  * Licensed under the Apache License, Version 2.0. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
```
schlosna: Should we undeprecate this class if we're intentionally using it again?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
schlosna: I assume if we want to use exemplars we'd instantiate a `com.palantir.tritium.metrics.registry.LockFreeExponentiallyDecayingReservoir` and if we didn't want exemplars we'd create a `com.codahale.metrics.LockFreeExponentiallyDecayingReservoir`. I think that is mostly reasonable, but we may need to be aware of possible megamorphic `Reservoir` callsites as a service may use multiple different `Reservoir` implementations beyond these two `LockFreeExponentiallyDecayingReservoir`, see https://shipilev.net/jvm/anatomy-quarks/16-megamorphic-virtual-calls/

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotWithExemplars.java
```diff
@@ -0,0 +1,140 @@
...
+        final List<WeightedSample> weightedSamples = new ArrayList<>();
+        this.exemplarMetadatas = new ArrayList<>();
+
+        values.forEach(v -> {
+            weightedSamples.add(new WeightedSample(v.value, v.weight));
+            if (v.exemplarMetadata != null) {
+                exemplarMetadatas.add(DefaultLongExemplar.of(v.exemplarMetadata, v.value));
+            }
+        });
+
```
schlosna: I'd suggest presizing `weightedSamples`, and only allocate the `ArrayList` when copying exemplar metadata

```suggestion
        final List<WeightedSample> weightedSamples = new ArrayList<>(values.size());
        List<LongExemplar<Object>> metadatas = null;

        for (WeightedSampleWithExemplar v : values) {
            weightedSamples.add(new WeightedSample(v.value, v.weight));
            if (v.exemplarMetadata != null) {
                if (metadatas == null) {
                    metadatas = new ArrayList<>();
                }
                metadatas.add(DefaultLongExemplar.of(v.exemplarMetadata, v.value));
            }
        }

        this.exemplarMetadatas = (metadatas == null) ? List.of() : metadatas;
```

carterkozak: Perhaps a presized ImmutableList.builderWithExpectedSize to avoid mutable state

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
carterkozak: I'd suggested modifying LockFreeExponentiallyDecayingReservoir for the initial round of review, then copying it to a new reservoir implementation and restoring the original deprecated version of this reservoir before the feature ends up merging, that way we're not re-reviewing the reservoir implementation, nor pushing folks back to a deprecated class -- does that pathing sound reasonable to you?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
carterkozak: I think there's some unknown space as we begin to roll out exemplars -- we may choose to sample exemplars only when traces are sampled, so even when enabled, the memory footprint may look very similar to the state of the world today

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotWithExemplars.java
marcioapaiva: Yeah presizing and allocating on demand (esp given we might end up with many cases of no collected exemplars) sounds reasonable to me.
We can use the builder with expected size for weightedSamples - but we just use it to feed the `WeightedSnapshot` class which copies it anyway, and we then don't keep a reference. I think the ImmutableList has more gains for the exemplarMetadatas list since we do keep a reference to that - but we don't have an expected size for that one so we should probably just use the standard ImmutableList.builder(). Changed both in any case for consistency, keeping the optimization for when we have no metadata: https://github.com/palantir/tritium/pull/1992/commits/466a5b2dcd92bd0ba90bb5b961cd7106a8badb4f

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotWithExemplars.java
```diff
@@ -0,0 +1,147 @@
...
+        @Nullable
+        public Object getExemplarMetadata() {
+            return exemplarMetadata;
+        }
+    }
+
+    private final WeightedSnapshot weightedSnapshot;
+    private final ExemplarMetadataProvider<?> exemplarProvider;
+    private final List<LongExemplar<Object>> exemplarMetadatas;
```
carterkozak: pedantic nit: metadata is already plural, no need to add the "s"

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotWithExemplars.java
```diff
@@ -0,0 +1,147 @@
...
+        public long getValue() {
+            return value;
+        }
+
+        public double getWeight() {
+            return weight;
+        }
+
+        @Nullable
+        public Object getExemplarMetadata() {
```
carterkozak: style nit: no need to mark these public since they're not overriding an interface method, package private is sufficient. I'd prefer to match the record style with getters like `value()` and `weight()` as opposed to `getValue()` and `getWeight()`/etc

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotWithExemplars.java
marcioapaiva: 👍  let me rename this one `exemplars` to get rid of the problem and also take in the fact that this is now a LongExemplar (so contains both the metadata and the value itself so a full Exemplar).

carterkozak:
This looks good to me -- @schlosna anything on your end before this merges?

schlosna (dismissed):
Overall seems reasonable to me.

carterkozak (approved):

PR #2003, opened 2024-08-29
Deprecate CaffeineCacheStats.registerCache

ash211:
## Before this PR
We added an improved mechanism for registering stats on a cache in https://github.com/palantir/tritium/pull/1897, with a few FLUPs (linked at the bottom of that PR).  The new method is also documented in the tritium README [here](https://github.com/palantir/tritium?tab=readme-ov-file#instrumenting-a-caffeine-cache).

However, existing code is still using the old mechanism, and callers who forgot to call `.recordStats()` are still being found (we noticed several this week).

## After this PR
==COMMIT_MSG==
Deprecate CaffeineCacheStats.registerCache
==COMMIT_MSG==

By marking this method as deprecated, we nudge developers who see these warnings in their IDE towards the preferred pattern.

## Possible downsides?
Repos that have opted into making deprecations fail builds and are using these now-deprecated methods won't be able to take the upgrade automatically.  However, this something those repos have opted into and should expect as a result of that choice.


tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CaffeineCacheStats.java
```diff
@@ -78,9 +78,9 @@ public static void registerCache(MetricRegistry registry, Cache<?, ?> cache, Str
      * @param cache cache to instrument
      * @param name cache name
      * <p>
-     * Soon to be deprecated, prefer {@link Caffeine#recordStats(Supplier)} and {@link CacheStats#of(TaggedMetricRegistry, String)}
+     * @deprecated Prefer {@link Caffeine#recordStats(Supplier)} and {@link CacheStats#of(TaggedMetricRegistry, String)}
```
schlosna: While we're here, can we add a link to the readme and update the JavaDoc here (maybe embed the recommended examples from readme)?

```suggestion
     * @deprecated Prefer {@link CacheStats#of(TaggedMetricRegistry, String)}.{@link CacheStats#register(Function) register(stats ->) {@link Caffeine#recordStats(Supplier) Caffeine.newBuilder().recordStats(stats)}...  
     * <p/>
     * See <a href="https://github.com/palantir/tritium?tab=readme-ov-file#instrumenting-a-caffeine-cache">Instrumenting a [Caffeine cache](https://github.com/ben-manes/caffeine/)</a>
```

carterkozak:
How widely is this API still used? How many of those repos compile with `-Werror`? We should take deprecation seriously -- unless humans are keeping a close eye on upgrade PRs, this will prevent those projects from taking almost all first party library upgrades until humans intervene!

ash211:
> How widely is this API still used? How many of those repos compile with `-Werror`? We should take deprecation seriously -- unless humans are keeping a close eye on upgrade PRs, this will prevent those projects from taking almost all first party library upgrades until humans intervene!

I discussed this in the PR description:

> Repos that have opted into making deprecations fail builds and are using these now-deprecated methods won't be able to take the upgrade automatically. However, this something those repos have opted into and should expect as a result of that choice.

I count 575 callsites internally: https://pl.ntr/2ok.  This deprecation would only affect repos that have `-Werror` AND `-Xlint:deprecation` enabled.  The first is enabled in the new default for highline-created repos now, as of March: 0682272.  But the second is not, and is less widely deployed (140+ internal repos).

Personally I think `-Xlint:deprecation` causes more problems than it's worth, and it's not enabled on most repos I regularly work with.  There needs to be a way to noticeably signal that something is going away, WITHOUT requiring that it happen immediately.  `@Deprecation` annotation is that, unless `-Xlint:deprecation` is turned on.  A Javadoc `@deprecated` isn't loud enough in IDEs.

bmarcaur (approved):

PR #2005, opened 2024-08-29
The `cache.stats.disabled` metric is listed in metric-schema

carterkozak:
This commit changes the type from counter to meter -- the recorded result should be the same, however meters can be elided when their value is zero because they cannot decrease, slightly reducing load on our indexing system.
In practice we only created the counters as needed, so this shouldn't change things in practice, but it will retain that safety even if refactored to create the metric object eagerly.

==COMMIT_MSG==
The `cache.stats.disabled` metric is listed in metric-schema
==COMMIT_MSG==


schlosna (approved):
Thanks @carterkozak !

PR #2018, opened 2024-09-12
Use metric-schema to report JVM Buffer Pool metrics

mpritham:
## Before this PR
We weren't using [metric-schema](https://github.com/palantir/metric-schema) when registering the JVM Buffer Pool metrics. Using metric-schema gives us some nice tooling, like markdown file generation allowing devs to find all published metrics easily.

## After this PR
==COMMIT_MSG==
Use metric-schema to report JVM Buffer Pool metrics
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
@@ -219,11 +219,11 @@ public final class JvmMetrics {
         JvmBuffersMetrics jvmBuffersMetrics = JvmBuffersMetrics.of(registry);
         for (BufferPoolMXBean pool : ManagementFactory.getPlatformMXBeans(BufferPoolMXBean.class)) {
             String poolName = pool.getName();
-            if (poolName.equals("direct")) {
+            if ("direct".equals(poolName)) {
                 jvmBuffersMetrics.directCount(nonNegative(pool::getCount));
                 jvmBuffersMetrics.directUsed(nonNegative(pool::getMemoryUsed));
                 jvmBuffersMetrics.directCapacity(nonNegative(pool::getTotalCapacity));
-            } else if (poolName.equals("mapped")) {
+            } else if ("mapped".equals(poolName)) {
                 jvmBuffersMetrics.mappedCount(nonNegative(pool::getCount));
                 jvmBuffersMetrics.mappedUsed(nonNegative(pool::getMemoryUsed));
                 jvmBuffersMetrics.mappedCapacity(nonNegative(pool::getTotalCapacity));
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
@@ -18,7 +18,6 @@ package com.palantir.tritium.metrics.jvm;
 
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.RatioGauge;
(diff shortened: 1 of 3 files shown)


tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -216,6 +215,16 @@ protected RatioGauge.Ratio getRatio() {
         });
     }
 
+    private static void registerJvmBufferPools(TaggedMetricRegistry registry) {
+        JvmBufferMetrics jvmBufferMetrics = JvmBufferMetrics.of(registry);
+        for (BufferPoolMXBean pool : ManagementFactory.getPlatformMXBeans(BufferPoolMXBean.class)) {
```
mpritham, line 220: This is a change.

The [BufferPoolMetricSet](https://github.com/dropwizard/metrics/blob/release/4.2.x/metrics-jvm/src/main/java/com/codahale/metrics/jvm/BufferPoolMetricSet.java#L26) explicitly gets the `direct` and `mapped` pools. I think we can just report metrics for all the pools exposed via `ManagementFactory.getPlatformMXBeans(BufferPoolMXBean.class)`?

tritium-metrics-jvm/src/main/metrics/metrics.yml
```diff
@@ -147,4 +147,25 @@ namespaces:
         docs: Amount of committed memory in bytes for the JVM non-heap (e.g. direct memory).
       non-heap.usage:
         type: gauge
-        docs: Ratio of `jvm.memory.non-heap.used` to `jvm.memory.non-heap.max`.
\ No newline at end of file
+        docs: Ratio of `jvm.memory.non-heap.used` to `jvm.memory.non-heap.max`.
+  jvm.buffer:
```
mpritham: typo: buffer**s**

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
mpritham: The metric names are currently: `jvm.buffers.direct.<>` and `jvm.buffers.mapped.<>`. I'd like to maintain backward compatibility. Let's just get the metrics for those specific pools.

tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -216,6 +215,22 @@ protected RatioGauge.Ratio getRatio() {
         });
     }
 
+    private static void registerJvmBufferPools(TaggedMetricRegistry registry) {
+        JvmBuffersMetrics jvmBuffersMetrics = JvmBuffersMetrics.of(registry);
+        for (BufferPoolMXBean pool : ManagementFactory.getPlatformMXBeans(BufferPoolMXBean.class)) {
+            String poolName = pool.getName();
+            if (poolName.equals("direct")) {
```
carterkozak: If you want to be particularly defensive against nulls, you could do something like this:
```suggestion
            if ("direct".equals(poolName)) {
```
Or `Objects.equals(poolName, "direct")` which handles both values being null (although that's impossible for a constant string). I wouldn't expect getName to ever return null either, but we've seen crazier things...

Entirely up to you whether to change this :-]

mpritham: Ah good call. I'll change this to the first idea: `"direct".equals(poolName)` for the null protection.

carterkozak (approved):
Thanks!

PR #2037, opened 2024-10-01
Add PerformanceSensitive annotations

schlosna:
## Before this PR

Tritium had no mechanisms for documenting the performance characteristics and considerations for modules, classes, or methods.

## After this PR

==COMMIT_MSG==
Document module, class, and/or method performance considerations.
==COMMIT_MSG==

### DIFF
--- a/tritium-annotations/src/main/java/com/palantir/tritium/annotations/PerformanceSensitive.java
+++ b/tritium-annotations/src/main/java/com/palantir/tritium/annotations/PerformanceSensitive.java
@@ -1,0 +1,56 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Elements annotated with {@link PerformanceSensitive} indicate that there are performance considerations to balance
+ * when modifying this element.
+ */
+@Target({ElementType.TYPE, ElementType.METHOD, ElementType.MODULE})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface PerformanceSensitive {
+
+    /**
+     * This element is sensitive to object allocations.
+     */
+    String Allocations = "Allocations";
+
+    /**
+     * This element is sensitive to cache access patterns.
+     */
+    String Cache = "Cache";
+
+    /**
+     * This element is sensitive to overall latency.
+     */
+    String Latency = "Latency";
+
+    /**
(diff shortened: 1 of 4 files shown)


tritium-annotations/src/main/java/com/palantir/tritium/annotations/PerformanceSensitive.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * This element is sensitive to overall latency.
+     */
+    String Latency = "Latency";
+
+    /**
+     * This element is sensitive to overall throughput.
+     */
+    String Throughput = "Throughput";
```
carterkozak: I wonder if this should be an enum, so we can manage the expected values centrally, and match them more easily with static analysis? No strong opinion, only thinking out loud

tritium-annotations/src/main/java/com/palantir/tritium/annotations/PerformanceSensitive.java
```diff
@@ -0,0 +1,56 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Elements annotated with {@link PerformanceSensitive} indicate that there are performance considerations to balance
+ * when modifying this element.
+ */
+@Target({ElementType.TYPE, ElementType.METHOD, ElementType.MODULE})
+@Retention(RetentionPolicy.RUNTIME)
```
carterkozak: Do we need runtime retention? If we're using this as a target to opt into specific static analysis rules, SOURCE retention should be sufficient (and doesn't require us to include jars at runtime)

schlosna: I was thinking we could do bytecode inspection, but probably not necessary now.

tritium-annotations/src/main/java/com/palantir/tritium/annotations/PerformanceSensitive.java
schlosna: Good call, updated to enum of `Condideration`

tritium-annotations/src/main/java/com/palantir/tritium/annotations/PerformanceSensitive.java
schlosna: We could still write error-prone checks validating certain behaviors (e.g. restricting use of streams within certain scopes).

tritium-annotations/src/main/java/com/palantir/tritium/annotations/PerformanceSensitive.java
```diff
@@ -26,31 +26,34 @@
...
      */
-    String[] value();
+    Consideration[] value();
+
+    enum Consideration {
+
+        /**
+         * This element is sensitive to object allocations.
+         */
+        Allocations,
```
carterkozak: Pedantic naming nits:

Perhaps allocation should be singular matching cache? As in “allocation sensitive”.

We usually use all-caps enum constants, might be worth standardizing

tritium-annotations/src/main/java/com/palantir/tritium/annotations/PerformanceSensitive.java
carterkozak: Source retention should be fine for this case, as long as we’re inspecting the annotated element. I imagine calling into a perf sensitive method shouldn’t be relevant to the caller, so bytecode retention shouldn’t buy us much unless we want to add specific instrumentation 👍

carterkozak:
Lgtm, defer to your judgement on the naming comment 👍

tritium-annotations/src/main/java/com/palantir/tritium/annotations/PerformanceSensitive.java
schlosna: good call, updated

carterkozak:
👍

PR #2176, opened 2025-05-16
Support metrics for AsyncCache

pkoenig10:
## Before this PR

`CacheStats.register` does not support `AsyncCache`.

## After this PR

Add a `CacheStats.registerAsync` method when using `AsyncCache`.

### DIFF
--- a/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
+++ b/tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
@@ -21,6 +21,7 @@ import static com.palantir.logsafe.Preconditions.checkState;
 import com.codahale.metrics.Counting;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.Timer;
+import com.github.benmanes.caffeine.cache.AsyncCache;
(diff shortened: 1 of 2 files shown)


tritium-caffeine/src/main/java/com/palantir/tritium/metrics/caffeine/CacheStats.java
```diff
@@ -70,6 +71,28 @@ public static CacheStats of(TaggedMetricRegistry taggedMetricRegistry, @Safe Str
...
+        metrics.estimatedSize().cache(name).build(cache::estimatedSize);
+        metrics.weightedSize().cache(name).build(() -> cache.policy()
+                .eviction()
+                .flatMap(e -> e.weightedSize().stream().boxed().findFirst())
+                .orElse(null));
+        metrics.maximumSize().cache(name).build(() -> cache.policy()
+                .eviction()
+                .map(Policy.Eviction::getMaximum)
+                .orElse(null));
+    }
```
pkoenig10, line 94: This is basically the API before https://github.com/palantir/tritium/pull/1920.

I decided to make this public so consumers have the option to register caches without the somewhat awkward (but less error-prone) fluent API.

schlosna (approved):
thanks @pkoenig10 !

PR #2177, opened 2025-05-17
Optimize WeightedSnapshotWithExemplars

schlosna:
## Before this PR

`WeightedSnapshotWithExemplars` uses `ImmutableList.Builder` that adds unnecessary extra array copies.

## After this PR

==COMMIT_MSG==
WeightedSnapshotWithExemplars uses ArrayList
==COMMIT_MSG==

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotWithExemplars.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotWithExemplars.java
@@ -18,6 +18,7 @@ package com.palantir.tritium.metrics.registry;
 import com.codahale.metrics.Snapshot;
 import com.codahale.metrics.WeightedSnapshot;
 import com.codahale.metrics.WeightedSnapshot.WeightedSample;
+import com.palantir.logsafe.Preconditions;
 import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -63,6 +64,13 @@ final class WeightedSnapshotWithExemplars extends Snapshot implements ExemplarsC
     private final ExemplarMetadataProvider<?> exemplarProvider;
     private final List<LongExemplar<Object>> exemplars;
 
+    private WeightedSnapshotWithExemplars(
+            ExemplarMetadataProvider<?> provider, List<LongExemplar<Object>> exemplars, WeightedSnapshot snapshot) {
+        this.exemplars = Collections.unmodifiableList(exemplars);
+        this.weightedSnapshot = Preconditions.checkNotNull(snapshot, "snapshot");
+        this.exemplarProvider = Preconditions.checkNotNull(provider, "provider");
+    }
+
     /**
      * Create a new {@link Snapshot} with the given values.
      *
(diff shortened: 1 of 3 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotWithExemplars.java
```diff
@@ -70,22 +71,21 @@ Object exemplarMetadata() {
...
             if (v.exemplarMetadata != null) {
                 if (exemplarsBuilder == null) {
-                    exemplarsBuilder = ImmutableList.builder();
+                    exemplarsBuilder = new ArrayList<>();
                 }
                 exemplarsBuilder.add(DefaultLongExemplar.of(v.exemplarMetadata, v.value));
             }
         }
-        this.exemplars = (exemplarsBuilder == null) ? ImmutableList.of() : exemplarsBuilder.build();
+        this.exemplars = (exemplarsBuilder == null) ? List.of() : Collections.unmodifiableList(exemplarsBuilder);
```
bmarcaur: Since we are always setting `this.exemplars` to _some_ list, can we set `exemplarsBuilder` to an empty collection instead of `null` and then avoid the `if (exemplarsBuilder == null)` check and also simplify this line to just `Collections.unmodifiableList(exemplarsBuilder);` 

What benefit are we getting by having the builder be `null`? At the start of the method?

schlosna: I was trying to avoid an ArrayList allocation, but we can just eat it (empty `ArrayList` doesn't allocate the backing `Object[]` array until an element is added, so its not too bad.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotWithExemplars.java
```diff
@@ -70,22 +71,16 @@ Object exemplarMetadata() {
...
-            weightedSamplesBuilder.add(new WeightedSample(v.value, v.weight));
+        List<WeightedSample> weightedSamples = new ArrayList<>(values.size());
+        List<LongExemplar<Object>> exemplarsBuilder = new ArrayList<>();
+        values.forEach(v -> {
+            weightedSamples.add(new WeightedSample(v.value, v.weight));
             if (v.exemplarMetadata != null) {
-                if (exemplarsBuilder == null) {
-                    exemplarsBuilder = ImmutableList.builder();
-                }
                 exemplarsBuilder.add(DefaultLongExemplar.of(v.exemplarMetadata, v.value));
```
schlosna: image

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/WeightedSnapshotWithExemplars.java
```diff
@@ -62,31 +64,35 @@ Object exemplarMetadata() {
...
-        for (WeightedSampleWithExemplar v : values) {
-            weightedSamplesBuilder.add(new WeightedSample(v.value, v.weight));
+    static Snapshot snapshot(ExemplarMetadataProvider<?> provider, Collection<WeightedSampleWithExemplar> values) {
+        List<WeightedSample> weightedSamples = new ArrayList<>(values.size());
+        List<LongExemplar<Object>> exemplars = new ArrayList<>();
+        values.forEach(v -> {
+            weightedSamples.add(new WeightedSample(v.value, v.weight));
             if (v.exemplarMetadata != null) {
-                if (exemplarsBuilder == null) {
-                    exemplarsBuilder = ImmutableList.builder();
```
schlosna, line 81: image

pkoenig10 (dismissed):

schlosna:
👍

pkoenig10 (approved):

PR #2184, opened 2025-05-29
Remove queued duration minimum threshold

pkoenig10:
## Before this PR

In our internal authentication service, we have a single threaded executor where, at any given time, there is at most one executing task and one queued task. The code looks something like:

```java
private final ExecutorService updateExecutor = Executors.newSingleThreadExecutor();
private final AtomicReference<SettableFuture<Void>> pendingUpdate = new AtomicReference<>();

Future<Void> updateCache() {
    SettableFuture<Void> future = pendingUpdate.get();
    if (future != null) {
        return future;
    }

    future = SettableFuture.create();

    SettableFuture<Void> witness = pendingUpdate.compareAndExchange(null, future);
    if (witness != null) {
        return witness;
    }

    future.setFuture(updateExecutor.submit(this::doUpdateCache, null));

    return future;
}

void doUpdateCache() {
    pendingUpdate.set(null);

    ...
}
```

Metrics seem to indicate that the queued duration p99 is longer than the duration p99. Here are metrics from our internal test environment.

Screenshot 2025-05-29 at 12 38 30 PM

This should be impossible, given how this executor is used.

But it happens because `TaggedMetricsExecutorService` is simply dropping any samples below the threshold. This causes the value of the queued duration metrics to be artificially inflated - especially for executors that typically have short queue duration.

It's confusing for measurements to simply be dropped in this way and causes the resulting metrics to be misleading.

## After this PR

`TaggedMetricsExecutorService` no longer excludes measurements from the queued duration metric. This metric now accurately captures the time between submission and execution for all submitted tasks.

### DIFF
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsExecutorService.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsExecutorService.java
@@ -29,14 +29,6 @@ import org.jspecify.annotations.Nullable;
 
 final class TaggedMetricsExecutorService extends AbstractExecutorService {
 
-    // 250ms minimum threshold is required to update the queued duration timer.
-    // The queued duration is an estimate based on time between a task being submitted
-    // and beginning to execute, there is always a delta between these operations, but
-    // it doesn't necessarily mean there's a queue at all. We assume anything longer than
-    // this threshold, which should be longer than pauses in most cases, is the result
-    // of queueing.
-    private static final long QUEUED_DURATION_MINIMUM_THRESHOLD_NANOS = 250_000_000L;
-
     private final ExecutorService delegate;
     private final String name;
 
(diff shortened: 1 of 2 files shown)


tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsExecutorService.java
```diff
@@ -29,14 +29,6 @@
 
 final class TaggedMetricsExecutorService extends AbstractExecutorService {
 
-    // 250ms minimum threshold is required to update the queued duration timer.
-    // The queued duration is an estimate based on time between a task being submitted
-    // and beginning to execute, there is always a delta between these operations, but
-    // it doesn't necessarily mean there's a queue at all. We assume anything longer than
-    // this threshold, which should be longer than pauses in most cases, is the result
-    // of queueing.
-    private static final long QUEUED_DURATION_MINIMUM_THRESHOLD_NANOS = 250_000_000L;
```
schlosna, line 38: https://github.com/palantir/tritium/pull/1230 was what originally added this threshold. Per discussions with @carterkozak & @pkoenig10 , we explicitly do not add queue metrics for cached executors in tritium clients (support for this was added in https://github.com/palantir/tritium/pull/1012).

schlosna (approved):

PR #2217, opened 2025-07-17
Remove custom Error Prone configuration

pkoenig10:
Baseline upgrades are blocked because Error Prone removed the `Java7ApiChecker` check.

```
Execution failed for task ':tritium-time:compileJava'.
> Java7ApiChecker is not a valid checker name
```

There's no need to disable these checks. It's better to selectively disable them in specific locations using `@SuppressWarnings` so that:
- Things don't break when checks are removed.
- We still get these checks in places we may want them

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -72,24 +72,6 @@ allprojects {
                 option('NullAway:AnnotatedPackages', 'com.palantir,com.google.common')
                 option('NullAway:CheckOptionalEmptiness', 'true')
 
-                // warnings not explicitly provided by error-prone
-                error 'NullAway',
-                        'Slf4jLogsafeArgs',
-                        'PreferCollectionTransform',
-                        'PreferListsPartition',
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -72,24 +72,6 @@ allprojects {
...
                 option('NullAway:CheckOptionalEmptiness', 'true')
 
-                // warnings not explicitly provided by error-prone
-                error 'NullAway',
-                        'Slf4jLogsafeArgs',
-                        'PreferCollectionTransform',
-                        'PreferListsPartition',
-                        'PreferSafeLoggingPreconditions',
-                        'PreferSafeLoggableExceptions',
-                        'PreferSafeLogger'
```
pkoenig10, line 82: This should be unnecessary with `-Werror`.

schlosna (approved):
Thanks @pkoenig10 !

PR #2236, opened 2025-08-11
Remove remoting3 tracing support

pkoenig10:
remoting3 has been dead for a long time. There are no longer any internal uses. The only places where repos are still using `com.palantir.remoting3:tracing` is because they are pulling it in transitively through outdated dependencies.


schlosna (approved):

pkoenig10:
👍

PR #2264, opened 2025-09-15
Downgrade undertow version to 2.3.18.Final

aldexis:
## Before this PR
There is a suspected bug in Undertow 2.3.19.Final, which released last week.

Downgrading this version manually in a few core repos (which would otherwise bump it to a lot of places) until we figure this out.

## After this PR

==COMMIT_MSG==
Downgrade undertow version to 2.3.18.Final
==COMMIT_MSG==


bjlaub:
👍

PR #2270, opened 2025-09-20
Fully enable NullAway:JSpecifyMode checking

schlosna:
## Before this PR

We were not using the full [JSpecify mode of NullAway](https://github.com/uber/NullAway/wiki/JSpecify-Support) and several packages were not fully `@NullMarked`.

## After this PR

==COMMIT_MSG==
All tritium packages are fully `@NullMarked` with jspecify `@Nullable` or `@NonNull` annotations as needed by APIs. Enable JSpecify mode for stricter NullAway nullness checking.

See https://github.com/uber/NullAway/wiki/JSpecify-Support and https://github.com/uber/NullAway/wiki/How-NullAway-Works#jspecify
==COMMIT_MSG==

### DIFF
--- a/tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
+++ b/tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
@@ -53,7 +53,7 @@ public final class CompositeInvocationEventHandler extends AbstractInvocationEve
 
     @Override
     public InvocationContext preInvocation(@NonNull Object instance, @NonNull Method method, @NonNull Object[] args) {
-        InvocationContext[] contexts = new InvocationContext[handlers.length];
+        @Nullable InvocationContext[] contexts = new InvocationContext[handlers.length];
 
         for (int i = 0; i < handlers.length; i++) {
             contexts[i] = Handlers.preWithEnabledCheck(
@@ -71,7 +71,7 @@ public final class CompositeInvocationEventHandler extends AbstractInvocationEve
         }
     }
 
-    private void success(@NonNull InvocationContext[] contexts, @Nullable Object result) {
+    private void success(@Nullable InvocationContext @NonNull [] contexts, @Nullable Object result) {
         for (int i = contexts.length - 1; i > -1; i--) {
             Handlers.onSuccess(handlers[i], contexts[i], result);
         }
@@ -85,7 +85,7 @@ public final class CompositeInvocationEventHandler extends AbstractInvocationEve
         }
     }
 
-    private void failure(InvocationContext[] contexts, @NonNull Throwable cause) {
+    private void failure(@Nullable InvocationContext @NonNull [] contexts, @NonNull Throwable cause) {
         for (int i = contexts.length - 1; i > -1; i--) {
             Handlers.onFailure(handlers[i], contexts[i], cause);
         }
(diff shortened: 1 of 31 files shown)


tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
```diff
@@ -98,15 +98,19 @@ public String toString() {
 
     static class CompositeInvocationContext extends DefaultInvocationContext {
 
-        private final InvocationContext[] contexts;
+        private final @Nullable InvocationContext @NonNull [] contexts;
```
schlosna: with jspecify this is equivalent to saying that the `contexts` array itself is not null; however, the `InvocationContext` elements may be null.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
```diff
@@ -108,7 +109,7 @@ private static String[] toArray(Map<String, String> data) {
             int valuesIndex = 2 * i;
             String key = keys[i];
             values[valuesIndex] = key;
-            values[valuesIndex + 1] = data.get(key);
+            values[valuesIndex + 1] = Strings.nullToEmpty(data.get(key));
```
schlosna, line 112: this is a slight behavior change where previously we may have allowed null values through, we now convert to an empty string tag value `""`, though previously we would likely trigger a `NullPointerException` when copying the tag map to an `ImmutableMap<String, String>`

tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
aldexis: These are pretty confusing to be honest. I can't imagine a better way to annotate these (maybe it would have been nicer to have a `@NullableElements` annotations, but I don't imagine these is one). For what it's worth, I've always assumed the annotation to apply to the field itself, so I would likely naturally read this the exact opposite (`@Nullable` applying to the field, and `@NonNull` applying to the elements in the array)

It's clearer when you read through https://jspecify.dev/docs/user-guide/#type-use-annotation-syntax but can certainly get folks to shoot themselves in the foot :(

It's unfortunate that this pattern is in so many places, as I think a quick comment explaining what this means would be great here.

For instance, below, 
```
        @Nullable
        InvocationContext @NonNull [] getContexts() {
```
clearly reads to me as `the return value of `getContexts()` can be null.

tritium-lib/src/main/java/com/palantir/tritium/proxy/InvocationEventProxy.java
```diff
@@ -72,7 +72,7 @@ public String toString() {
     /** Optimized to avoid excessive stack frames for readable stack traces. */
     @Override
     @Nullable
-    public final Object invoke(Object proxy, Method method, @Nullable Object[] nullableArgs) throws Throwable {
+    public final Object invoke(Object proxy, Method method, Object @Nullable [] nullableArgs) throws Throwable {
```
aldexis, line 75: This is an example of what I described in my other comment as being confusing. And makes me wonder if places like https://github.com/palantir/tritium/blob/98661b54d2bdf0c8f2d51a25639f2c5bba61b47c/tritium-metrics/src/main/java/com/palantir/tritium/event/metrics/TaggedMetricsServiceInvocationEventHandler.java#L80-L80 are correct or not

tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
pkoenig10: Does it work if we move the `@NonNull` annotation to the field instead?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
pkoenig10: I don't like silently replacing `null` with the empty string. These are semantically different values and this implicit conversion can be confusing for consumers.

tritium-core/src/main/java/com/palantir/tritium/event/CompositeInvocationEventHandler.java
schlosna: I'm going to switch this back to `draft` as there are some places that were not previously annotated correctly and some places where NullAway does not seem to properly handle propagation with arrays in jspecify mode right now.

tritium-lib/src/main/java/com/palantir/tritium/proxy/InvocationEventProxy.java
schlosna: I think this should actually be:

```suggestion
    public final Object invoke(Object proxy, Method method, @Nullable Object @Nullable [] nullableArgs) throws Throwable {
```

schlosna:
closing in favor of https://github.com/palantir/tritium/pull/2581

PR #2295, opened 2025-10-16
Add opinionated cache API

pkoenig10:
This PR aims to provide a simpler, more tailored, and harder to misuse cache API in an effort to avoid many of the pitfalls we find developers running into when using caches.

This code was adapted from a similar API we designed in an internal service.

## Background

The Caffeine API is powerful but has some gotchas:

1. Synchronous loading caches can block the retrieval while other entires we being loaded.

    This is due to the coarse nature of `ConcurrentHashMap` locking. The `Map.compute*` APIs suffer from a similar problem. But when using an in-memory map, this mapping function is typically pure so the critical section is short. But loading values into a cache typically requires I/O, which holds the lock for a much longer period of time.

2. By default, caches use global, shared resources.

    Most notably, the default executor is the `ForkJoinPool.commonPool()`. `ForkJoinPool.commonPool()` is a work-stealing executor that is intended to be used for cooperative multi-tasking. It only has [as many threads as processors](https://github.com/openjdk/jdk/blob/890adb6410dab4606a4f26a942aed02fb2f55387/src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java#L2767) - the expectation is that you are not performing blocking I/O on this executor.

    Caffeine uses this executor to load values in async caches and refresh values in both sync and async caches. Loading/refreshing values often requires I/O and these tasks can quickly starve other tasks submitted to `ForkJoinPool.commonPool()`. Nested cache loads introduce the potential for deadlock, although `CompletableFuture` has some [tricks](https://github.com/openjdk/jdk/blob/jdk-21%2B35/src/java.base/share/classes/java/util/concurrent/CompletableFuture.java#L1886-L1887) to avoid deadlocks when using `ForkJoinPool`.

    `Scheduler.systemScheduler()` is another example of a global shared resource. Using this scheduled with a something like a direct executor can hog the single global scheduling thread, delaying the execution of other scheduled tasks.

3. `invalidateAll` does not provide the same guarantees as `ConcurrentHashMap.clear()`.

    `invalidateAll` is not always performed while holding the map's locks. [This Caffeine issue](https://github.com/ben-manes/caffeine/issues/1739) does a good job describing the limitations and the constraints that resulted in this tradeoff.

    This can be problematic. It is common to have caches that we want to clear in response to events indicating the the backing values have changed. But without these guarantees, it's possible that `invalidateAll` will not see entries that have already started to be loaded. These loaders then may load stale values (values before the change that precipitated the invalidation event) into the cache.

## Change in this PR

This PR introduces the `tritium-cache` library without own Cache API and implementations that wrap Caffeine caches in order to provider the enforcement and behavior we desire. The cache API is a mostly drop-in replacement for the subset of the Caffeine API we support and should be sufficient for the vast majority of users.

The cache interfaces, cache implementations, and builder API are pretty straight-forward. There are 3 interfaces:
- `Cache` is a non-loading cache
- `LoadingCache` is a loading cache that supports loading singular values at a time
- `BulkLoadingCache` is a loading cache that supports loading multiple values at a time

Caches are constructed using the builder API defined in `CacheBuilder`. This is a staged builder to force decisions to be explicit, such as:
- Whether a cache is sync or async
- What executor this cache uses
- If this cache records metrics

More considerations are detailed in comments below.

### DIFF
--- a/tritium-cache/build.gradle
+++ b/tritium-cache/build.gradle
@@ -1,0 +1,18 @@
+apply plugin: 'com.palantir.external-publish-jar'
+apply plugin: 'org.revapi.revapi-gradle-plugin'
+
+dependencies {
+    api project(':tritium-registry')
+
+    implementation project(':tritium-caffeine')
+    implementation 'com.github.ben-manes.caffeine:caffeine'
+    implementation 'com.google.errorprone:error_prone_annotations'
+    implementation 'com.google.guava:guava'
+    implementation 'com.palantir.safe-logging:preconditions'
+    implementation 'com.palantir.safe-logging:safe-logging'
+    implementation 'org.jspecify:jspecify'
+
+    testImplementation 'org.assertj:assertj-core'
+    testImplementation 'org.junit.jupiter:junit-jupiter'
+    testImplementation 'org.junit.jupiter:junit-jupiter-api'
+}
--- a/tritium-cache/build.gradle
+++ b/tritium-cache/build.gradle
@@ -10,8 +10,10 @@ dependencies {
     implementation 'com.google.guava:guava'
     implementation 'com.palantir.safe-logging:preconditions'
     implementation 'com.palantir.safe-logging:safe-logging'
+    implementation 'com.palantir.tracing:tracing'
     implementation 'org.jspecify:jspecify'
 
+    testImplementation 'com.palantir.tracing:tracing-api'
     testImplementation 'org.assertj:assertj-core'
     testImplementation 'org.junit.jupiter:junit-jupiter'
     testImplementation 'org.junit.jupiter:junit-jupiter-api'
--- a/tritium-cache/src/main/java/com/palantir/tritium/cache/AsyncBulkLoadingCache.java
+++ b/tritium-cache/src/main/java/com/palantir/tritium/cache/AsyncBulkLoadingCache.java
@@ -1,0 +1,27 @@
+/*
+ * (c) Copyright 2025 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.cache;
+
+import java.util.Map;
+
+public interface AsyncBulkLoadingCache<K, V> extends AsyncLoadingCache<K, V> {
+
+    /**
+     * See {@link com.github.benmanes.caffeine.cache.LoadingCache#getAll(Iterable)}.
+     */
+    Map<K, V> getAll(Iterable<? extends K> keys);
+}
--- a/tritium-cache/src/main/java/com/palantir/tritium/cache/AsyncCacheImpl.java
+++ b/tritium-cache/src/main/java/com/palantir/tritium/cache/AsyncCacheImpl.java
@@ -21,6 +21,7 @@ import static com.palantir.logsafe.Preconditions.checkState;
 import com.google.common.collect.Iterators;
 import com.google.errorprone.annotations.MustBeClosed;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
+import com.palantir.tracing.Tracers;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -35,9 +36,11 @@ import org.jspecify.annotations.Nullable;
 
 class AsyncCacheImpl<K, V> implements AsyncCache<K, V> {
 
+    private final String loadOperation;
     private final com.github.benmanes.caffeine.cache.AsyncCache<K, V> cache;
 
-    AsyncCacheImpl(com.github.benmanes.caffeine.cache.AsyncCache<K, V> cache) {
+    AsyncCacheImpl(String name, com.github.benmanes.caffeine.cache.AsyncCache<K, V> cache) {
+        this.loadOperation = name + " cache load";
         this.cache = cache;
     }
 
@@ -51,8 +54,8 @@ class AsyncCacheImpl<K, V> implements AsyncCache<K, V> {
     @Nullable
     public final V get(K key, Function<? super K, ? extends V> mappingFunction) {
         CompletableFuture<V> future;
-        try (ValueFactory<K, V> valueFactory = new ValueFactory<>(mappingFunction)) {
-            future = cache.get(key, valueFactory);
+        try (Loader<K, V> loader = loader(mappingFunction)) {
+            future = cache.get(key, loader);
         }
 
         return await(future);
@@ -68,9 +71,8 @@ class AsyncCacheImpl<K, V> implements AsyncCache<K, V> {
             Iterable<? extends K> keys,
             Function<? super Set<? extends K>, ? extends Map<? extends K, ? extends V>> mappingFunction) {
         CompletableFuture<Map<K, V>> future;
-        try (ValueFactory<Set<? extends K>, Map<? extends K, ? extends V>> valueFactory =
-                new ValueFactory<>(mappingFunction)) {
-            future = cache.getAll(keys, valueFactory);
+        try (Loader<Set<? extends K>, Map<? extends K, ? extends V>> loader = loader(mappingFunction)) {
+            future = cache.getAll(keys, loader);
         }
 
         return await(future);
@@ -126,22 +128,28 @@ class AsyncCacheImpl<K, V> implements AsyncCache<K, V> {
         }
     }
 
+    @MustBeClosed
+    private <I, O> Loader<I, O> loader(Function<? super I, ? extends O> mappingFunction) {
+        return new Loader<>(loadOperation, mappingFunction);
+    }
+
     // This class exists to ensure that we do not start loading values until the entry is inserted into the cache.
     // This ensures that invalidateAll will remove entries that are being loaded.
     //
     // By default, Caffeine creates async cache entries with something like CompletableFuture.supplyAsync. If the load
     // completes and invalidateAll is called before the entry is inserted into the cache, then we may insert a stale
     // entry into the cache.
-    private static final class ValueFactory<I, O>
-            implements BiFunction<I, Executor, CompletableFuture<O>>, AutoCloseable {
+    private static final class Loader<I, O> implements BiFunction<I, Executor, CompletableFuture<O>>, AutoCloseable {
 
+        private final String loadOperation;
         private final Function<? super I, ? extends O> mappingFunction;
 
         @Nullable
         private Runnable runnable;
 
         @MustBeClosed
-        ValueFactory(Function<? super I, ? extends O> mappingFunction) {
+        Loader(String loadOperation, Function<? super I, ? extends O> mappingFunction) {
+            this.loadOperation = loadOperation;
             this.mappingFunction = mappingFunction;
         }
 
@@ -153,13 +161,13 @@ class AsyncCacheImpl<K, V> implements AsyncCache<K, V> {
 
             runnable = () -> {
                 try {
-                    executor.execute(() -> {
+                    executor.execute(Tracers.wrap(loadOperation, () -> {
                         try {
                             future.complete(mappingFunction.apply(key));
                         } catch (Throwable t) {
                             future.completeExceptionally(t);
                         }
-                    });
+                    }));
                 } catch (Throwable t) {
                     future.obtrudeException(t);
                 }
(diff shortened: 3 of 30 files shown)


tritium-cache/build.gradle
```diff
@@ -0,0 +1,18 @@
+apply plugin: 'com.palantir.external-publish-jar'
+apply plugin: 'org.revapi.revapi-gradle-plugin'
+
+dependencies {
+    api project(':tritium-registry')
```
pkoenig10, line 5: I have deliberately avoided using any Caffeine types in the API so Caffeine is entirely an implementation detail. This has a couple of benefits:
- We aren't coupled to the Caffeine API, including any API breaks.
- We can swap out the implementation in the future, if desired.

tritium-cache/src/main/java/com/palantir/tritium/cache/AsyncBulkLoadingCache.java
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.cache;
+
+import java.util.Map;
+
+public interface AsyncBulkLoadingCache<K, V> extends AsyncLoadingCache<K, V> {
```
pkoenig10, line 21: We define a separate interface for this to get compile-time type safety instead of `UnsupportedOperationException` if we attempt to call `getAll` on a cache whose cache loader [does not support bulk loads](https://github.com/ben-manes/caffeine/blob/93d845e58d8e7bf2dfc88a31c5a078bca5bf4dbf/caffeine/src/main/java/com/github/benmanes/caffeine/cache/CacheLoader.java#L90-L92).

tritium-cache/src/main/java/com/palantir/tritium/cache/AsyncCacheImpl.java
```diff
@@ -0,0 +1,178 @@
...
+    @Override
+    @Nullable
+    public final V get(K key, Function<? super K, ? extends V> mappingFunction) {
+        CompletableFuture<V> future;
+        try (ValueFactory<K, V> valueFactory = new ValueFactory<>(mappingFunction)) {
+            future = cache.get(key, valueFactory);
+        }
+
+        return await(future);
+    }
```
pkoenig10, line 62: This is how we get the desired `invalidateAll` behavior for async caches. We don't actually start to load the value until the future has been inserted into the cache.

tritium-cache/src/main/java/com/palantir/tritium/cache/Cache.java
```diff
@@ -0,0 +1,82 @@
...
+        SizeBuilder<K, V> name(@CompileTimeConstant String name);
+    }
+
+    @CheckReturnValue
+    public interface SizeBuilder<K, V> {
+
+        ExpiryBuilder<K, V> maximumSize(long maximumSize);
+
+        ExpiryBuilder<K, V> maximumSize(long maximumSize, Weigher<? super K, ? super V> weigher);
+    }
```
pkoenig10, line 43: I avoided providing methods that take a `Refreshable<Long> maximumSize` for a couple of reasons:

- Avoid adding a `refreshable` dependency.

- Simplify the implementation.

    Updating the maximum size requires creating a refreshable subscriber whose lifetime we have to manage. It's not clear how we would do that unless we make the entire `Cache` `AutoCloseable`, which is sort of weird.

- It's not clear to me that there is a real need for this.

    There are only a handful of places internally where we have live-reloaded maximum sizes, and I suspect that we rarely change the maximum size at runtime. Most of those places would probably be fine with fixed maximum cache sizes.

tritium-cache/src/main/java/com/palantir/tritium/cache/Cache.java
```diff
@@ -0,0 +1,82 @@
...
+        ExecutorBuilder<K, V> metrics(TaggedMetricRegistry taggedMetrics);
+
+        ExecutorBuilder<K, V> noMetrics();
+    }
+
+    @CheckReturnValue
+    public interface ExecutorBuilder<K, V> {
+
+        Builder<K, V> executor(ExecutorFactory executorFactory);
+    }
```
pkoenig10, line 65: We don't have a good way to prevent users from using undesirable executors (ex. the fork-join common pool). But at least this being a staged builder forces users to think about this. And I doubt users will reach for the fork-join common pool here unless they do intentionally want that.

pkoenig10:
I'm putting up this PR because we talked about pulling this out to a library so we could use it in other places. It's not entirely clear to me whether:
- We still think this is desirable
- Tritium is an appropriate place for it (this doesn't really have anything to do with instrumentation)

On the latter point, I'm happy to move this to a new repo, but would like to first confirm that we actually want to create this library and agree on the API.

tritium-cache/src/main/java/com/palantir/tritium/cache/AsyncBulkLoadingCache.java
ben-manes: Note that Caffeine shouldn't throw that in general practice since it uses [reflection](https://github.com/ben-manes/caffeine/blob/93d845e58d8e7bf2dfc88a31c5a078bca5bf4dbf/caffeine/src/main/java/com/github/benmanes/caffeine/cache/LocalLoadingCache.java#L219-L228) at construction to inspect if the method was overridden. That seemed more elegant than a custom exception with `writableStackTrace` disabled, though that would have been nicer for adapters.

The tradeoff seemed to be if more api surface area is preferred to only provide `getAll` if bulk loads or else force an explicit loop of individual `get` loads. Since the cache is an implementation detail, like using a HashMap, then the `getAll` code and the loader definition are often defined in the same source file so the behavior is not that surprising. For an opinionated api then I think more types probably wins that argument as a worthy cost in the complexity budget, whereas for the general library there is so much conceptual weight to juggle that minimizing api surface area does reduces some tension. Similar thinking for preferring [bulk](https://github.com/ben-manes/caffeine/blob/93d845e58d8e7bf2dfc88a31c5a078bca5bf4dbf/caffeine/src/main/java/com/github/benmanes/caffeine/cache/CacheLoader.java#L211-L229) static factory methods in the general case, whereas more type driven might be nicer in yours.

tritium-cache/src/main/java/com/palantir/tritium/cache/AsyncCacheImpl.java
ben-manes: fwiw, the reason Caffeine couldn't do it this way is because the future is supplied by the user in their `AsyncCacheLoader`. Since `CompletableFuture` is identity equality it is important to return the user's future instance back to them. For instance if they want to cancel or remove the instance by the `asMap` view.

tritium-cache/src/main/java/com/palantir/tritium/cache/Cache.java
ben-manes: I was hoping to someday switch to virtual threads by default, but even on JDK 25 I've found it a problematic to use due to an abundance of native code (e.g. brotli compression) and being cooperatively scheduled. It is kind of odd that the current core team argues the common pool default for `CompletableFuture` will always be ideal as they see futures as cpu-driven instead of I/O-driven work, but I think that might be Stockholm syndrome until VTs gain preemption support.

tritium-cache/src/main/java/com/palantir/tritium/cache/AsyncBulkLoadingCache.java
pkoenig10: Right, thanks for the clarification.

Re-reading this, my comment above isn't the reason why we went with a separate interface. The pitfall we've seen is creating a loading cache with a cache loader that does not implement `loadAll` and then using `getAll`, not realizing that the values will be loaded sequentially. Our caches are typically backed by a DB which does support bulk reads, but devs often use a lambda to define the cache loader because it's the immediate/easy thing to do, not realizing that there is this somewhat hidden `loadAll` method that they also should be implementing.

This design flips the `load`/`loadAll` dependency so users are forced to implement `loadAll` and `load` delegates to `loadAll` by default. This helps push users into the "pit of success" by making the easy thing the performant thing.

tritium-cache/src/main/java/com/palantir/tritium/cache/AsyncCacheImpl.java
pkoenig10: Yeah that totally makes sense.

I greatly appreciate all the work you've done on Caffeine and I think most of the design tradeoffs Caffeine makes are reasonable as a general-purpose caching library, where it's probably better to bias towards more flexibility/capability, sometimes at the expense of it becoming a slightly sharper tool.

This API is making a different set of tradeoffs, optimizing for usage by less experienced developers who may not be aware of some of the subtle concerns when using caches.

tritium-cache/src/main/java/com/palantir/tritium/cache/Cache.java
pkoenig10: > they see futures as cpu-driven instead of I/O-driven work

This also feels somewhat odd to me. In my experience, futures are used overwhelmingly for I/O-bound work. Cooperatively scheduled tasks are somewhat rare - if something is CPU-bound we typically just execute it directly.

We've generally eschewed `CompletableFuture` in favor of Guava's `ListenableFuture` for various reasons - most notably because of their [inability interrupt execution](https://stackoverflow.com/a/29014156) and the [inconsistent way that failures manifest to consumers](https://stackoverflow.com/a/29014156).

tritium-cache/src/main/java/com/palantir/tritium/cache/AsyncCacheImpl.java
```diff
@@ -0,0 +1,178 @@
...
+                throw runtimeException;
+            } else if (e.getCause() instanceof Error error) {
+                error.addSuppressed(new SafeRuntimeException("Cache load failed"));
+                throw error;
+            } else {
+                throw new SafeRuntimeException("Cache load failed", e.getCause());
+            }
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
```
schlosna:
```suggestion
            throw new SafeRuntimeException("Interrupted while loading", e);
```

tritium-cache/src/test/java/com/palantir/tritium/cache/CacheTest.java
```diff
@@ -0,0 +1,365 @@
...
+                .noExpiry()
+                .noMetrics()
+                .executor(_name -> executor)
+                .buildSync();
+
+        cache.put("key1", "value1");
+
+        Future<String> future = executor.submit(() -> {
+            return cache.get("key2", _key -> {
+                startLatch.countDown();
```
schlosna, line 306: Do we need the `startLatch` as we're not `await`ing on it?

tritium-cache/src/test/java/com/palantir/tritium/cache/CacheTest.java
```diff
@@ -0,0 +1,365 @@
...
+                .noExpiry()
+                .noMetrics()
+                .executor(_name -> executor)
+                .buildAsync();
+
+        cache.put("key1", "value1");
+
+        Future<String> future = executor.submit(() -> {
+            return cache.get("key2", _key -> {
+                startLatch.countDown();
```
schlosna, line 341: same question about `startLatch`

tritium-cache/src/main/java/com/palantir/tritium/cache/Cache.java
schlosna: Agreed we probably don't want to manage the lifetime & cleanup of Refreshable, and the places currently using live reloading policies can continue just using Caffeine directly.

@Shibi-bala for SA

schlosna (dismissed):
Overall looks good to me, thanks @pkoenig10 .

Curious if you have cut a local release to test with this yet or if we should cut an RC before locking in API? I'm fairly confident in the API, but have a couple projects I'd like to migrate over as a test.

tritium-cache/src/test/java/com/palantir/tritium/cache/CacheTest.java
pkoenig10: Oh this is a bug - we should be awaiting it before we call `entries` which demonstrates that `entires` does no return in-flight loads.

pkoenig10:
I haven't cut an RC, but we can try this out by just publishing to `mavenLocal`.

The API here is very similar to the API in the internal service this was adapted from.

pkoenig10:
I think trying this out in a couple places before merging would be a good idea.

schlosna (commented):
Approving to cut RC and test this out

schlosna (dismissed):

schlosna (dismissed):

tritium-cache/src/main/java/com/palantir/tritium/cache/AsyncCacheImpl.java
```diff
@@ -0,0 +1,178 @@
...
+
+            runnable = () -> {
+                try {
+                    executor.execute(() -> {
+                        try {
+                            future.complete(mappingFunction.apply(key));
+                        } catch (Throwable t) {
+                            future.completeExceptionally(t);
+                        }
+                    });
```
pkoenig10: We made a change to our internal implementation to wrap this in `Tracers.wrap` so that async cache load spans use the requestor's trace ID. This has proved quite valuable.

How do we feel about taking a dependency on `tracing-java` here?

schlosna: that seems reasonable. we'll have a transitive dependency of `tritium-cache` -> `tracing-java` -> `com.palantir.conjure.java.api:errors` -> `tritium-ids` we should be aware, but given GCV should be fine in practice

schlosna (dismissed):

schlosna (approved):

PR #2319, opened 2025-11-06
Update JVM metrics to make thread metrics optional

andybradshaw:
## Before this PR

Previously, a user could either register all tritium JVM metrics or none. This could cause problems for applications where collecting some of the thread metrics is a known expensive operation. This PR enables users to opt out of some metrics when registering JVM metrics.

## After this PR

==COMMIT_MSG==
Update JVM metric registration API to allow opting out of thread metrics
==COMMIT_MSG==

## Possible downsides?

This not provides _some_ configurability to something that used to be consistent for all users of the library. In this case, I believe the complexity is worth it, but we can add additional docs to make it clear that most users will not need to opt out.

### DIFF
--- a/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
+++ b/tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
@@ -18,6 +18,7 @@ package com.palantir.tritium.metrics.jvm;
 
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.RatioGauge;
+import com.codahale.metrics.RatioGauge.Ratio;
 import com.codahale.metrics.jvm.ThreadDeadlockDetector;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Suppliers;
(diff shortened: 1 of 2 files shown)


tritium-metrics-jvm/src/main/java/com/palantir/tritium/metrics/jvm/JvmMetrics.java
```diff
@@ -259,4 +260,52 @@ private static Long negativeToNull(@Nullable Long value) {
...
+            }
+        }
+
+        /**
+         * Registers the set of enabled metrics.
+         *
+         * @param registry metric registry
+         */
+        @SuppressWarnings("checkstyle:CyclomaticComplexity")
+        public void register(TaggedMetricRegistry registry) {
```
schlosna: I'm on the fence about converting this to a builder, I think we might want to just expose `JvmMetrics.registerWithoutThreadMetrics(registry)` for folks that want to opt-out of the thread state gauges, i.e. https://github.com/palantir/tritium/pull/2320

schlosna (approved):

andybradshaw:
👍

tpetracca:
👍🏻

PR #2360, opened 2026-01-08
Revert unintended error-prone change in #2333

aldexis:
## Before this PR
See https://github.com/palantir/tritium/pull/2333/changes#r2670096520

I don't believe this would have any adverse impact, as it would just recursively call `registerWithReplacement` and go through the early return in the next call, but this was unintended

## After this PR

==COMMIT_MSG==
Revert unintended error-prone change
==COMMIT_MSG==


ash211:
This bug is suspected to be blocking an excavator in internal PR foundry/interactive-infra/pull/11698

aldexis:
Ah indeed, I hadn't realized there would be implementations of gauge that wouldn't return exactly the gauge we pass in, which would cause an infinite loop indeed

kunalrkak (approved):

ash211:
Good to merge+release?

PR #2376, opened 2026-01-28
Add `queued-delay` metric for ScheduledExecutorService

fsamuel-bs:
## Before this PR

We would not populate the `executor.queued-duration` metric for scheduled executor services.

## After this PR
We populate this metric.

Additional considerations:
- thought of adding a new metric to `ExecutorMetrics`, but biased for ease-of-use - dashboards already reference this metric, and it still makes sense in this context
- considered adding a `reportQueuedDuration` boolean, similar to how we instrument `ExecutorService`. Decided against that because all internal usages require a fixed thread pool for `ScheduledExecutorService`.

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsScheduledThreadPoolExecutor.java
+++ b/tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsScheduledThreadPoolExecutor.java
@@ -17,12 +17,20 @@
 package com.palantir.tritium.metrics;
 
 import com.codahale.metrics.Timer;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import org.jetbrains.annotations.NotNull;
 
+/**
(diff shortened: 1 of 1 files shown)


pkoenig10:
Can you please update the PR description to describe the motivation for adding this metric?

I don't think this metric makes sense.

`ScheduledThreadPoolExecutor` has no maximum pool size, so any task that can be run immediately should never be queued. The only time I'd expect a non-negligible value for this metric is when the thread itself isn't getting scheduled immediately. And the hiccup meter already measures this.

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsScheduledThreadPoolExecutor.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.concurrent.TimeUnit;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Instruments an {@link ScheduledThreadPoolExecutor} that monitors the delay between when a task was scheduled
+ * to be executed and when it actually got executed.
+ * Use in conjunction with {@link MetricRegistries#instrument(TaggedMetricRegistry, ScheduledExecutorService, String)}
+ * to fully instrument an {@link ScheduledExecutorService}.
+ */
+public class TaggedMetricsScheduledThreadPoolExecutor extends ScheduledThreadPoolExecutor {
```
pkoenig10, line 34: The other `TaggedMetrics*Executor` classes are private implementation details of Tritium that are only exposed through the `MetricRegistries` utility. This is a departure from that design. I don't think we want to be directly exposing utility classes like this.

PR #2381, opened 2026-01-30
Add `queued-delay` metric for ScheduledExecutorService

fsamuel-bs:
Alternative proposal to https://github.com/palantir/tritium/pull/2376.

Internally, all our scheduled thread pool executors are constructed with `ScheduledThreadPoolExecutor(int corePoolSize, ThreadFactory threadFactory)`. This means they have a fixed `corePoolSize` with `Integer.MAX_VALUE` maximumPoolSize, which effectively yields a fixed thread pool size with `corePoolSize` threads.

This makes such executors prone to saturation - if enough clients submit long-running tasks to such an executor service, they might be queued before execution. The metric `executor.queued-duration` should thus contain how long these executors were queued for.

## Before this PR

We would not populate the `executor.queued-duration` metric for scheduled executor services.

## After this PR
We populate this metric.

==COMMIT_MSG==
==COMMIT_MSG==


tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsScheduledExecutorService.java
```diff
@@ -121,24 +132,50 @@ public String toString() {
         return "TaggedMetricsScheduledExecutorService{name=" + name + ", delegate='" + delegate + "'}";
     }
 
+    private static long triggerTime(long initialDelay, long periodInNanos) {
+        return System.nanoTime() + Math.min(initialDelay, MAX_NANOS) + Math.min(periodInNanos, MAX_NANOS);
```
pkoenig10: What is the purpose of `Math.min(value, MAX_NANOS)` here?

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsScheduledExecutorService.java
```diff
@@ -121,24 +132,50 @@ public String toString() {
...
 
         @Override
         @SuppressWarnings("PreferJavaTimeOverload") // performance sensitive
         public void run() {
             running.inc();
             long startNanos = System.nanoTime();
+
+            queuedDuration.update(startNanos - triggerTime, TimeUnit.NANOSECONDS);
+            if (kind == Kind.RATE) {
+                triggerTime = triggerTime(triggerTime, periodInNanos);
```
pkoenig10: This should not call `System.nanoTime()` again - it can just increment `triggerTime`.

I'd prefer to write this in a way that mirrors the implementation in `ScheduledThreadPoolExecutor`, unifying these with a `setNextRunTime` invocation in the `finally` block.

tritium-metrics/src/test/java/com/palantir/tritium/metrics/TaggedMetricsScheduledExecutorServiceTest.java
```diff
@@ -103,4 +103,43 @@ void testScheduledDurationMetrics(TaggedMetricRegistry registry) throws Exceptio
...
+        Future<String> secondTask =
+                executorService.submit(() -> Thread.currentThread().getName());
+        executorService.shutdown();
+
+        startLatch.await();
+
+        assertThat(metrics.running(NAME).getCount()).isOne();
+        assertThat(metrics.duration(NAME).getCount()).isZero();
+
+        Thread.sleep(1L);
```
pkoenig10: We should not sleep in tests

schlosna (commented):
I'm supportive of these changes, should address Patrick's comments

tritium-metrics/src/main/java/com/palantir/tritium/metrics/TaggedMetricsScheduledExecutorService.java
fsamuel-bs: Something that had come from `ScheduledThreadPoolExecutor`'s implementation. I don't think we need it, addressed.

schlosna (approved):

PR #2450, opened 2026-05-14
Replace WeightedSnapshot with primitive-array PrimitiveWeightedSnapshot

schlosna:
## Before this PR

The codahale `WeightedSnapshot` incurred allocation and CPU overhead due to boxing arrays.

## After this PR

==COMMIT_MSG==
Eliminates boxing overhead from codahale WeightedSnapshot which used Arrays.sort(Object[], Comparator.comparingLong) triggering TimSort on boxed WeightedSample objects. The new PrimitiveWeightedSnapshot sorts parallel long[]/double[] arrays directly with a custom quicksort (median-of-three pivot, insertion sort for small partitions, O(log n) stack depth via tail-recursion elimination).

Additional optimizations:
- Eliminate per-snapshot normWeights array (~8KB at 1028 elements), replaced with scalar inverseSumWeight and deferred multiplication
- Eliminate intermediate WeightedSample allocations in reservoir getSnapshot() paths
- Lazy exemplar list allocation in WeightedSnapshotWithExemplars
- Empty snapshot singleton short-circuit

JMH benchmark results (us/op, lower is better):

  getSnapshot (n=100):  WeightedSnapshot 4.67 -> Primitive 3.00 (1.6x)
  getSnapshot (n=1028): WeightedSnapshot 93.2 -> Primitive 53.9 (1.7x)
  getSnapshotAndRead (n=1028): WeightedSnapshot 121.1 -> Primitive 55.7 (2.2x)

  Allocation per snapshot (n=1028): 47,193 B -> 24,752 B (48% reduction)
==COMMIT_MSG==

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
@@ -19,12 +19,12 @@ package com.palantir.tritium.metrics.registry;
 import com.codahale.metrics.Clock;
 import com.codahale.metrics.Reservoir;
 import com.codahale.metrics.Snapshot;
-import com.codahale.metrics.WeightedSnapshot;
 import com.codahale.metrics.WeightedSnapshot.WeightedSample;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.time.Duration;
+import java.util.Collection;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
@@ -224,7 +224,20 @@ public final class LockFreeExponentiallyDecayingReservoir implements Reservoir {
     @Override
     public Snapshot getSnapshot() {
         State stateSnapshot = rescaleIfNeeded(clock.getTick());
-        return new WeightedSnapshot(stateSnapshot.values.values());
+        Collection<WeightedSample> samples = stateSnapshot.values.values();
+        int sampleCount = samples.size();
+        if (sampleCount == 0) {
+            return PrimitiveWeightedSnapshot.EMPTY;
+        }
+        long[] sampleValues = new long[sampleCount];
+        double[] sampleWeights = new double[sampleCount];
+        int idx = 0;
+        for (WeightedSample sample : samples) {
+            sampleValues[idx] = sample.value;
+            sampleWeights[idx] = sample.weight;
+            idx++;
+        }
+        return new PrimitiveWeightedSnapshot(sampleValues, sampleWeights);
     }
 
     public static Builder builder() {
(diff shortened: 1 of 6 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/PrimitiveWeightedSnapshot.java
```diff
@@ -0,0 +1,234 @@
...
+
+    private static final int INSERTION_SORT_THRESHOLD = 47;
+
+    private final long[] values;
+    private final double[] weights;
+    private final double[] quantiles;
+    private final double inverseSumWeight;
+
+    PrimitiveWeightedSnapshot(long[] values, double[] weights) {
+        sortParallelArrays(values, weights, 0, values.length - 1);
```
fsamuel-bs, line 44: I know we already put the comment in the class, but I'd put a comment here highlighting this is the only change in this class.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
```diff
@@ -224,7 +224,20 @@ private State doRescale(long currentTick, State stateSnapshot) {
     @Override
     public Snapshot getSnapshot() {
         State stateSnapshot = rescaleIfNeeded(clock.getTick());
-        return new WeightedSnapshot(stateSnapshot.values.values());
+        Collection<WeightedSample> samples = stateSnapshot.values.values();
```
fsamuel-bs, line 227: We'd also need to re-route internal usages to point to the `com.palantir.tritium.metrics.registry.LockFreeExponentiallyDecayingReservoir.java` rather than the codahale one, right?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/PrimitiveWeightedSnapshot.java
```diff
@@ -0,0 +1,234 @@
...
+import java.io.PrintWriter;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+/**
+ * A {@link Snapshot} implementation that avoids the boxing overhead of
+ * {@link com.codahale.metrics.WeightedSnapshot} by sorting primitive {@code long[]} values
+ * directly rather than using {@code Arrays.sort(Object[], Comparator)}.
+ */
+final class PrimitiveWeightedSnapshot extends Snapshot {
```
schlosna, line 32: we can look at upstreaming to https://github.com/dropwizard/metrics/blob/release/4.2.x/metrics-core/src/main/java/com/codahale/metrics/WeightedSnapshot.java though would like to proceed with package private version for now

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/LockFreeExponentiallyDecayingReservoir.java
schlosna: Yes, folks would need to be using `LockFreeExponentiallyDecayingReservoir` or `LockFreeExponentiallyDecayingReservoirWithExemplars` to pick up this new `PrimitiveWeightedSnapshot` if/until upstreamed. Fortunately, this is the default for `DefaultTaggedMetricRegistry` so likely covers 99%+ of histograms/timers.

https://github.com/palantir/tritium/blob/363a0ec6cce143c393a179469533c876bb4d2078/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java#L26-L27

PR #2451, opened 2026-05-14
More efficient TagMap construction

fsamuel-bs:
## Before this PR

Most of our metrics use https://github.com/palantir/metric-schema to be constructed - this means that tags are code-generated and known at compile time. But they still build metrics with the pattern 
```
                            MetricName.builder()
                                    .safeName("cache.request.count")
                                    .putSafeTags("cache", "test")
                                    .build())
```
which calls `TagMap#withEntries`. `TagMap#withEntries` made this codepath be inefficient both CPU and memory-wise - it copied all entries for every tag inserted.

This PR adds a new builder for metrics, `MetricName#builderWithExpectedTags(int expectedTags)`, which expects keys in lexicographical order, and does not copy keys and values to construct a MetricName.

Benchmarks (with prior version of this code, pre #2453):

```
  ┌───────────┬──────────────────┬──────────────────┬─────────┬───────────────────────┬───────────────────────┐
  │ Benchmark │ Standard (ns/op) │ PreSized (ns/op) │ Speedup │ Alloc Standard (B/op) │ Alloc PreSized (B/op) │
  ├───────────┼──────────────────┼──────────────────┼─────────┼───────────────────────┼───────────────────────┤
  │ 3 tags    │             43.0 │             24.1 │    1.8x │                   192 │                   104 │
  ├───────────┼──────────────────┼──────────────────┼─────────┼───────────────────────┼───────────────────────┤
  │ 7 tags    │            147.9 │             46.0 │    3.2x │                   496 │                   136 │
  ├───────────┼──────────────────┼──────────────────┼─────────┼───────────────────────┼───────────────────────┤
  │ 13 tags   │            385.8 │             83.3 │    4.6x │                  1192 │                   184 │
  └───────────┴──────────────────┴──────────────────┴─────────┴───────────────────────┴───────────────────────┘
```

After https://github.com/palantir/tritium/pull/2453, copying here the metrics in the description of that PR:
```
Benchmark results from AMD EPYC 7R13 (pre-sized builder, sorted keys)

┌──────┬─────────────────┬──────────────────┬─────────┬───────────────┬────────────────┬─────────────────┐
│ Tags │ Regular (ns/op) │ PreSized (ns/op) │ Speedup │ Regular Alloc │ PreSized Alloc │ Alloc Reduction │
├──────┼─────────────────┼──────────────────┼─────────┼───────────────┼────────────────┼─────────────────┤
│ 3    │ 56.37 ± 4.20    │ 12.42 ± 1.70     │ 4.5×    │ 136 B/op      │ 80 B/op        │ −41%            │
├──────┼─────────────────┼──────────────────┼─────────┼───────────────┼────────────────┼─────────────────┤
│ 7    │ 242.60 ± 13.97  │ 44.61 ± 3.04     │ 5.4×    │ 416 B/op      │ 112 B/op       │ −73%            │
├──────┼─────────────────┼──────────────────┼─────────┼───────────────┼────────────────┼─────────────────┤
│ 13   │ 802.12 ± 37.91  │ 93.74 ± 4.00     │ 8.6×    │ 976 B/op      │ 160 B/op       │ −84%            │
└──────┴─────────────────┴──────────────────┴─────────┴───────────────┴────────────────┴─────────────────┘
```

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/MetricName.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/MetricName.java
@@ -46,7 +46,7 @@ public interface MetricName {
      * Returns an optimized builder for callers that know the number of tags upfront.
      * Tags must be added via {@link PreSizedBuilder#putSafeTags(String, String)} in lexicographic key order.
      */
-    static PreSizedBuilder builder(int expectedTags) {
+    static PreSizedBuilder builderWithExpectedTags(int expectedTags) {
         return new PreSizedBuilder(expectedTags);
     }
 
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/MetricName.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/MetricName.java
@@ -42,5 +42,19 @@ public interface MetricName {
         return new Builder();
     }
 
+    /**
+     * Returns an optimized builder for callers that know the number of tags upfront.
+     * Tags must be added via {@link PreSizedBuilder#putSafeTags(String, String)} in lexicographic key order.
+     */
+    static PreSizedBuilder builder(int expectedTags) {
+        return new PreSizedBuilder(expectedTags);
+    }
+
     final class Builder extends ImmutableMetricName.Builder {}
+
+    final class PreSizedBuilder extends ImmutableMetricName.PreSizedBuilder {
+        PreSizedBuilder(int expectedTags) {
+            super(expectedTags);
+        }
+    }
 }
--- a/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
+++ b/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
@@ -18,6 +18,8 @@ package com.palantir.tritium.metrics.registry;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Ordering;
+import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
 import java.lang.reflect.Array;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -111,6 +113,41 @@ final class TagMap implements SortedMap<String, String> {
             values[valuesIndex + 1] = data.get(key);
         }
         return values;
+    }
+
+    /** Builds a new {@link TagMap}. Keys are expected to be inserted in lexicographical order. */
+    static final class Builder {
+        private final String[] values;
+        private final int expectedSize;
+        private int size;
+
+        Builder(int expectedSize) {
+            this.values = new String[expectedSize * 2];
+            this.expectedSize = expectedSize;
+            this.size = 0;
+        }
+
+        Builder put(String key, String value) {
+            int arrayLen = size * 2;
+            values[arrayLen] = key;
+            values[arrayLen + 1] = value;
+            size++;
+            return this;
+        }
+
+        TagMap build() {
+            if (size == 0) {
+                return EMPTY;
+            }
+
+            Preconditions.checkArgument(
+                    size == expectedSize,
+                    "Mismatch between number of keys expected and number of keys inserted in TagMap#builder",
+                    SafeArg.of("actualSize", size),
+                    SafeArg.of("expectedSize", expectedSize));
+
+            return new TagMap(values);
+        }
     }
 
     /** Returns a new {@link TagMap} with an additional or updated entry. */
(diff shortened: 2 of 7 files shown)


tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
```diff
@@ -113,6 +115,47 @@ private static String[] toArray(Map<String, String> data) {
...
+                    size == expectedSize,
+                    "TagMap#builder inserted keys should be equal to the number of expected keys",
+                    SafeArg.of("size", size),
+                    SafeArg.of("expectedSize", expectedSize));
+
+            for (int i = 2; i < values.length; i += 2) {
+                Preconditions.checkArgument(
+                        values[i - 2].compareTo(values[i]) < 0,
+                        "TagMap#builder keys should be inserted in lexicographically ascending order");
+            }
```
fsamuel-bs: It pains me to have to check keys are indeed given in lexicographical order, but 🤷, not sure we can trust clients here.

schlosna: This feels a bit dangerous and likely to lead to a lot of runtime failures. 

I would much prefer just sorting the tag key/value pairs here.

Do we have tests asserting behavior?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
```diff
@@ -113,6 +115,47 @@ private static String[] toArray(Map<String, String> data) {
...
+        TagMap build() {
+            if (size == 0) {
+                return EMPTY;
+            }
+
+            Preconditions.checkArgument(
+                    size == expectedSize,
+                    "TagMap#builder inserted keys should be equal to the number of expected keys",
+                    SafeArg.of("size", size),
+                    SafeArg.of("expectedSize", expectedSize));
```
schlosna: do we actually want to throw rather than shrink array (at cost of allocation)?

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/MetricName.java
```diff
@@ -42,5 +42,19 @@ static Builder builder() {
         return new Builder();
     }
 
+    /**
+     * Returns an optimized builder for callers that know the number of tags upfront.
+     * Tags must be added via {@link PreSizedBuilder#putSafeTags(String, String)} in lexicographic key order.
+     */
+    static PreSizedBuilder builder(int expectedTags) {
```
schlosna: nit: can we clarify name here?

```suggestion
    static PreSizedBuilder builderWithExpectedTags(int expectedTags) {
```

schlosna (commented):
Directionally supportive of this change, thanks @fsamuel-bs ! My main concern is around potential runtime failures if tag key/value puts are not sorted in code and failing if not exactly sized (e.g. imagine one has code with originally sized `5` then removes a line for a `.putSafeTags("foo", "bar")` and now that fails at runtime due to oversized array. The inverse of undersizing the estimate is also a concern, though we need to decide the growth function / multiplier.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
fsamuel-bs: Added a test.

fsamuel-bs:
> My main concern is around potential runtime failures if tag key/value puts are not sorted in code and failing if not exactly sized (e.g. imagine one has code with originally sized 5 then removes a line for a .putSafeTags("foo", "bar") and now that fails at runtime due to oversized array

> This feels a bit dangerous and likely to lead to a lot of runtime failures.

> do we actually want to throw rather than shrink array (at cost of allocation)?

The idea of this PR is to expose a constructor that can be used by `metric-schema`, which knows the number and order of tags at compile-time. I'm happy to call it `unsafeBuilder` if that prevents others from using it outside of metric-schema. I'm also happy to add fail-safes (e.g. shrinkage / growth) if we really need to, but I'd rather keep things simple if we could and fail loudly if someone used this in an unexpected way.

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
schlosna: https://github.com/palantir/tritium/pull/2453 on top of this branch for what I was thinking in terms of expansion, contraction, sorting, and expanded test coverage

fsamuel-bs:
👍

schlosna (approved):

bjlaub (approved):

PR #2513, opened 2026-07-07
CircleCI uses JDK 17 image

schlosna:
## Before this PR

CircleCI was using JDK 11 base image

## After this PR

==COMMIT_MSG==
CircleCI uses JDK 17 base image to unblock gradle 9
==COMMIT_MSG==

## Possible downsides?

We need to excavate this [across many other repos](https://github.com/search?q=org%3Apalantir%20export%20JDK%3D11&type=code)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ aliases:
 jobs:
 
   check:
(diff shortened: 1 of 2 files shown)


FinlayRJW:
👍

PR #2559, opened 2026-08-24
Cache stat meters don't record moving averages

occasional-handbook:
## Before this PR

When recording hits and misses for caffeine caches, we use a meter. This requires, broadly:
- one call to System.nanotime()
- 4 increments (1 for counter, and 3 for moving averages)

We'd like to remove some of this overhead - we never use the moving averages, so we end up paying the cost for no reason. Additionally, they somewhat harm local benchmarks non-obviously - eg on an M5 mac, running the benchmark on 6 threads contends heavily on nanoTime in a way that's not obvious to the user. (note it's also not clear from JFRs, since System.nanoTime is a JIT intrinsic which will not emit a frame - benefits will likely be fairly platform-dependent).

The natural choice would be using a counter, but https://github.com/palantir/tritium/pull/1897#discussion_r1516789394 indicates that we want to keep those as meters - so opting to just zero out the moving averages.

running the benchmark on develop:
t=1
```
Benchmark                           Mode  Cnt   Score   Error  Units

CacheStatsBenchmark.recordEviction  avgt   30  33.113 ± 0.961  ns/op

CacheStatsBenchmark.recordHit       avgt   30  15.856 ± 0.187  ns/op
```
t=6
```
Benchmark                           Mode  Cnt    Score   Error  Units

CacheStatsBenchmark.recordEviction  avgt   30  392.320 ± 3.555  ns/op

CacheStatsBenchmark.recordHit       avgt   30  197.206 ± 3.214  ns/op
```

on this branch
t=1
```
Benchmark                           Mode  Cnt  Score   Error  Units

CacheStatsBenchmark.recordEviction  avgt   30  4.601 ± 0.048  ns/op

CacheStatsBenchmark.recordHit       avgt   30  4.285 ± 0.063  ns/op
```
t=6
```
Benchmark                           Mode  Cnt  Score   Error  Units

CacheStatsBenchmark.recordEviction  avgt   30  6.030 ± 0.157  ns/op

CacheStatsBenchmark.recordHit       avgt   30  4.384 ± 0.037  ns/op
```

Note I don't think the contention will be real on linux (those numbers are from an M5 mac), but this is hard to measure.

## After this PR

==COMMIT_MSG==
Cache stat meters don't record moving averages
==COMMIT_MSG==

## Possible downsides?

it's pretty confusing that the rate() metrics will get zero'd out... it feels like we should just bite the bullet and make those counters instead.

### DIFF
--- a/tritium-caffeine/src/test/java/com/palantir/tritium/metrics/caffeine/DisabledMovingAveragesTest.java
+++ b/tritium-caffeine/src/test/java/com/palantir/tritium/metrics/caffeine/DisabledMovingAveragesTest.java
@@ -1,0 +1,47 @@
+/*
+ * (c) Copyright 2026 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tritium.metrics.caffeine;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.codahale.metrics.Clock;
+import com.codahale.metrics.Meter;
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicLong;
+import org.junit.jupiter.api.Test;
+
+final class DisabledMovingAveragesTest {
+
+    @Test
+    void countsWithoutRates() {
+        AtomicLong nanos = new AtomicLong();
+        Meter meter = new Meter(DisabledMovingAverages.INSTANCE, new Clock() {
+            @Override
+            public long getTick() {
+                return nanos.get();
+            }
+        });
+
+        meter.mark(5);
+        nanos.addAndGet(Duration.ofMinutes(1).toNanos());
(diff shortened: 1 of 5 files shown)


schlosna (approved):
Thanks for the contribution! [`System.nanoTime()` definitely adds overhead](https://shipilev.net/blog/2014/nanotrusting-nanotime/) and we don't need that for these meters.

I ran the benchmark on AMD EPYC 7R13 with JDK 25, 3 forks, 10 warmup and 10 measurement iterations. Both branches had effectively zero allocations.

| Benchmark | develop | Feature | Improvement |
|---|---:|---:|---:|
| recordHit | 35.192 ± 0.228 ns/op | 2.842 ± 0.005 ns/op | 12.38× faster |
| recordEviction | 70.816 ± 0.363 ns/op | 7.643 ± 0.019 ns/op | 9.27× faster |

pkoenig10:
I don't think this library is an appropriate place for this change.

Our internal metric reporting library never reports moving averages for meters, but that doesn't mean that this library should also take that opinion. Instead this should be done by our internal metric reporting library. This has the added benefit of picking up this improvement for all of our meters, not just the Tritium cache metrics.

tritium-caffeine/src/test/java/com/palantir/tritium/metrics/caffeine/DisabledMovingAveragesTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.codahale.metrics.Clock;
+import com.codahale.metrics.Meter;
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicLong;
+import org.junit.jupiter.api.Test;
+
+final class DisabledMovingAveragesTest {
```
pkoenig10, line 27: This feels like a fairly useless test.

PR #2345, opened 2025-12-08
Remove error-prone AllSuggestionsAsWarnings

schlosna:
## Before this PR

[`RequireExplicitNullMarking` NullAway error-prone check](https://github.com/uber/NullAway/blob/master/nullaway/src/main/java/com/uber/nullaway/RequireExplicitNullMarking.java) enabled by `AllSuggestionsAsWarnings` is blocking gradle-baseline upgrade https://github.com/palantir/tritium/pull/2344/ . There is https://github.com/palantir/tritium/pull/2270 to expand full jspecify + NullAway support; however, that is currently on hold.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

PR #2581, opened 2026-09-18
Enable NullAway JSpecify mode

schlosna:
## Before this PR

NullAway used its legacy annotation interpretation, so JSpecify annotations did not consistently describe array and element nullness across Tritium.

## After this PR

==COMMIT_MSG==
Enable NullAway's JSpecify mode, mark previously unmarked packages with `@NullMarked`, and correct the affected nullness declarations. Add compile-time dependencies required by the newly checked sources and Byte Buddy's JSR-305 annotation metadata.
==COMMIT_MSG==

## Possible downsides?

The stricter nullness model may surface additional compilation failures when code in these packages changes.


build.gradle
```diff
@@ -115,6 +116,53 @@ allprojects {
...
+            }
+        }
+
+        if (!missingPackages.isEmpty()) {
+            throw new GradleException(
+                    'Every com.palantir Java source package must declare @NullMarked in package-info.java:\n'
+                            + missingPackages.sort().collect { "  - ${it}" }.join('\n'))
+        }
+    }
+}
```
pkoenig10: We shouldn't write complicated custom Gradle checks like this.

`NullAway` has an Error Prone check that will enforce this for us if we opt-in:

https://github.com/uber/NullAway/wiki/JSpecify-Support#requireexplicitnullmarking-checker

tritium-lib/build.gradle
```diff
@@ -3,6 +3,8 @@ apply plugin: 'com.palantir.metric-schema'
 apply plugin: 'com.palantir.revapi'
 
 dependencies {
+    compileOnly 'com.google.code.findbugs:jsr305'
```
pkoenig10, line 6: I don't think this is necessary?

build.gradle
schlosna: good call, will enable that & `JSpecifyUnrecognizedAnnotationLocation ` as well as bring back `AllSuggestionsAsWarnings` that was removed by https://github.com/palantir/tritium/pull/2345 due to `RequireExplicitNullMarking`

tritium-lib/build.gradle
schlosna: This is needed due to the Byte Buddy dependency as its `@MaybeNull` annotation references JSR-305’s `When.MAYBE`, but JSR-305 isn’t supplied transitively.

I tested removing it but `:tritium-lib:compileJava` fails with class file for `javax.annotation.meta.When` not found because `-Werror` makes the warning fatal.

schlosna: Long term I'd like to remove dependency on jsr305, but that depends on the dependencies

tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java
```diff
@@ -109,7 +109,7 @@ private static String[] toArray(Map<String, String> data) {
             int valuesIndex = 2 * i;
             String key = keys[i];
             values[valuesIndex] = key;
-            values[valuesIndex + 1] = data.get(key);
+            values[valuesIndex + 1] = Preconditions.checkNotNull(data.get(key), "value");
```
pkoenig10, line 112: nit: static import `checkNotNull`

build.gradle
```diff
@@ -71,8 +71,15 @@ allprojects {
             ]
 
             options.errorprone {
+                error('RequireExplicitNullMarking')
+                error('JSpecifyUnrecognizedAnnotationLocation')
```
pkoenig10, line 75: nit: alphabetize and whitespace after


```suggestion
                error('JSpecifyUnrecognizedAnnotationLocation')
                error('RequireExplicitNullMarking')

```

tritium-lib/build.gradle
pkoenig10: I see, this is sort of a deficiency of byte-buddy. It has a `com.google.code.findbugs:jsr305` dependency but it's a `provided` dependency rather than a `compile` dependency because it isn't needed at runtime. But Maven has no equivalent to Gradle's `compileOnlyApi` to expose compile-only dependencies to consumers.

pkoenig10 (approved):

